PR #4, opened 2015-07-04
Added postcss-modules-local-by-default dependency to package.json

jamiefaye:
I was getting webpack issues: Error: Cannot find module 'postcss-modules-local-by-default', so I added that dependency to the root package.json file.


necolas:
It was unused. Fixed now. Thanks

PR #21, opened 2015-09-15
No need to explicitly `npm install` React

AnSavvides:
I might be missing something, but `react` is a dependency for `react-native-web` as per `package.json`, so there is no need to explicitly `npm install` it since it will get installed anyway.


necolas:
React is now a dev dependency: 1c444569ae8f12130197459ed0d34829826008a0

PR #22, opened 2015-09-23
Server side rendering

necolas:
Exploring adding support for server-side inlining of the backing CSS library

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,6 +51,26 @@ const styles = {
 }
 ```
 
+### Server-side rendering
+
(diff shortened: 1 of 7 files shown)


necolas:
tests are failing in firefox because the stylesheet is no longer automatically injected when a component is mounted. probably best to avoid measuring the height of the textarea, especially since it's not consistent across browsers. perhaps fork the textarea dependency and add tests in the longer term.

necolas:
abandoning this approach

PR #43, opened 2015-12-07
Image: 'backgroundColor' default should be 'transparent'

Dremora:
Trying to render an image with transparent pixels results in a nasty background.
This doesn't happen in React Native — transparent pixels are rendered correctly.


necolas:
Thanks! Been meaning to fix that. Applied in https://github.com/necolas/react-native-web/commit/e1da11fa1dabd238b1e8a3527155b5b408623e50

Issue #52, opened 2015-12-19
Should 'react' and 'react-dom' be bundled by default?


necolas:
React Native bundles a (forked) version of React with each release. That strategy would reduce compatibility issues between this framework and versions of React (as well as fewer packages to install). If this project were to do the same, it might make it a bit more complicated to determine which other `react-*` packages are compatible with a release though. That's probably the main reason I've avoided doing so for now.

Another option the bundling of react packages would provide is the ability to provide `render` methods that take care of the stylesheet creation.

necolas:
Leaving as peer dependencies

PR #59, opened 2015-12-22
Include react-dom package

necolas:
Example of including react-dom in react-native-web, and moving the stylesheet management into the library (no longer responsibility of app developer). Other methods on react-dom could also be included much like React Native's environment-specific additions.

Relates to #52

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,5 +1,8 @@
 import React from 'react'
+import ReactDOM from 'react-dom'
+import ReactDOMServer from 'react-dom/server'
 
+// api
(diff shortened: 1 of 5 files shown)


src/index.js
```diff
@@ -21,11 +42,10 @@ const ReactNative = {
   Touchable,
   View,
 
-  // apis
-  StyleSheet,
-
   // React
-  ...React
+  ...React,
```
necolas: could spread ReactDOM too, effectively returning to the 0.13 API since we're web-only

necolas:
merged

PR #64, opened 2015-12-28
Support array styles

### LINKED ISSUE #61: StylePropType: support array style

allofthenorthwood:
Fixes #61. This takes the simple approach of just flattening the arrays.

### DIFF
--- a/docs/apis/StyleSheet.md
+++ b/docs/apis/StyleSheet.md
@@ -29,10 +29,10 @@ Use styles:
 ```js
 <View style={styles.container}>
   <Text
-    style={{
-      ...styles.title,
-      ...(this.props.isActive && styles.activeTitle)
-    }}
+    style={[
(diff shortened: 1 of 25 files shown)


allofthenorthwood:
I updated the pull request to pass the Travis tests (mixed up some case-sensitive filenames, whoops), and to support the full range of style propTypes that React Native supports. In this version styles can be arbitrarily nested arrays, where the innermost values can be anything falsey or an object that matches the component style propType.

allofthenorthwood:
Travis will pass once Issue #69 is merged.

necolas:
Thanks a lot for this! We'd refactored `StyleSheet`, and I've just pushed another refactor that builds upon the architecture using in React Native. It now supports array styles. See 715c71b21564061f2c71917c53b49e9a6b92bd6a

PR #69, opened 2016-01-14
Add semicolons to property initializers

spicyj:
According to the current proposal, semicolons are required. Babel 6.4 was updated to require them (see https://github.com/babel/babel/pull/3225). This is required in order to fix the build -- master will be red on Travis as soon as the build is rerun because it will pick up the latest Babel parser.


ericmuyser:
+1 I had to do exactly this just now.

MoOx:
Must have indeed.

necolas:
Thank you! RIP semicolon-less future. Applied in a0187f9b1a9abcf6552c08b485dc1999caa21b5d

PR #79, opened 2016-03-04
Added support for Image.props.style.resizeMode

IjzerenHein:
As per the latest react-native API, the resizeMode for an image can also be specified through the style object. This is often more practical as all styling can be done in a style-sheet. The priority of resizeMode is as follows:
- props.resizeMode
- props.style.resizeMode
- 'cover'

### DIFF
--- a/src/components/Image/index.js
+++ b/src/components/Image/index.js
@@ -6,6 +6,7 @@ import ImageResizeMode from './ImageResizeMode'
 import ImageStylePropTypes from './ImageStylePropTypes'
 import React, { Component, PropTypes } from 'react'
 import StyleSheetPropType from '../../apis/StyleSheet/StyleSheetPropType'
+import flattenStyle from '../../apis/StyleSheet/flattenStyle'
 import View from '../View'
 
 const STATUS_ERRORED = 'ERRORED'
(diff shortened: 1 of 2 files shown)


src/components/Image/index.js
```diff
@@ -130,17 +130,21 @@ export default class Image extends Component {
       accessible,
       children,
       defaultSource,
-      resizeMode,
       source,
-      style,
       testID
     } = this.props
+    const style = flattenStyle(this.props.style)
```
necolas: this could be `StyleSheet.flatten` instead

src/components/Image/index.js
```diff
@@ -130,17 +130,21 @@ export default class Image extends Component {
...
 
     const isLoaded = this.state.status === STATUS_LOADED
     const defaultImage = defaultSource.uri || null
     const displayImage = !isLoaded ? defaultImage : source.uri
     const backgroundImage = displayImage ? `url("${displayImage}")` : null
 
+    const resizeMode = this.props.resizeMode || (style || {}).resizeMode || 'cover'
```
necolas: Rather than using the string `cover`, we can use `ImageResizeMode.cover`.

We should be able to avoid the potential creation of a new object in `render`, for example by including a default value for `style` in `defaultProps`. That would also avoid the need for the condition around `delete`.

necolas:
I made those changed and applied in 9a094565326de127e7925bd9564707e63820a7ab. thanks!

PR #80, opened 2016-03-04
Added support for directly requiring image assets

IjzerenHein:
Native-script allows for directly requiring image assets without the need of wrapping it in an uri object`{uri: 'someasset'}`. This commit adds this feature `<Image source={require('someasset.png')} />` as well as adds strong typing for uri-based sources:

```javascript
const ImageSourcePropType = PropTypes.oneOfType([
  PropTypes.shape({
    uri: PropTypes.string.isRequired
  }),
  PropTypes.string // Opaque type returned by require('./image.jpg')
])
```

final result, these both work:

```javascript
<Image source={require('../assets/myasset.png')} />
<Image source={{uri: require('../assets/myasset.png')}} />
```

### DIFF
--- a/src/components/Image/ImageStylePropTypes.js
+++ b/src/components/Image/ImageStylePropTypes.js
@@ -2,6 +2,7 @@ import { PropTypes } from 'react'
 import ColorPropType from '../../apis/StyleSheet/ColorPropType'
 import LayoutPropTypes from '../../apis/StyleSheet/LayoutPropTypes'
 import TransformPropTypes from '../../apis/StyleSheet/TransformPropTypes'
+import ImageResizeMode from './ImageResizeMode'
 
 const hiddenOrVisible = PropTypes.oneOf([ 'hidden', 'visible' ])
 
@@ -10,6 +11,7 @@ export default {
   ...TransformPropTypes,
   backfaceVisibility: hiddenOrVisible,
   backgroundColor: ColorPropType,
+  resizeMode: PropTypes.oneOf(Object.keys(ImageResizeMode)),
   /**
    * @platform web
    */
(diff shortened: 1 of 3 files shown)


src/components/Image/index.js
```diff
@@ -130,17 +137,20 @@ export default class Image extends Component {
       accessible,
       children,
       defaultSource,
-      resizeMode,
       source,
-      style,
       testID
     } = this.props
+    const style = flattenStyle(this.props.style)
```
necolas: would prefer `StyleSheet.flatten`

src/components/Image/ImageStylePropTypes.js
```diff
@@ -2,6 +2,7 @@ import { PropTypes } from 'react'
 import ColorPropType from '../../apis/StyleSheet/ColorPropType'
 import LayoutPropTypes from '../../apis/StyleSheet/LayoutPropTypes'
 import TransformPropTypes from '../../apis/StyleSheet/TransformPropTypes'
+import ImageResizeMode from './ImageResizeMode'
```
necolas: please could you remove the lines from the `resizeMode` patch

src/components/Image/index.js
```diff
@@ -14,35 +16,39 @@ const STATUS_LOADING = 'LOADING'
 const STATUS_PENDING = 'PENDING'
 const STATUS_IDLE = 'IDLE'
 
+const ImageSourcePropType = PropTypes.oneOfType([
+  PropTypes.shape({
+    uri: PropTypes.string.isRequired
+  }),
+  PropTypes.string // Opaque type returned by require('./image.jpg')
```
necolas: nit: move the EOL comment to above the line

necolas:
Thanks. Please could you update `docs/guides/react-native.md` with details about needing to use a webpack loader to support `require`-ing of image assets. And we should have a test for this feature.

IjzerenHein:
Awesome, I'll make the changes and create a new pull request for it.

IjzerenHein:
I've created a new pull request for this + added tests & doc section :+1: 
#84

PR #86, opened 2016-03-08
Added NativeModules shim (adds support for 'react-native-vector-icons')

IjzerenHein:
The NativeModules namespace is used by external libraries such as 'react-native-vector-icons'. That particular library throws an error because the NativeModules namespace doesn't exist. This commit adds the NativeModules namespace shim.

'react-native-vector-icons' can then be used with 'react-native-web' like this:

```html
<head>
   ...
  <style type="text/css">
    @font-face {
      font-family: 'Ionicons';
      src: url(node_modules/react-native-vector-icons/Fonts/Ionicons.ttf);
    }
  </style>
   ...
</head>
```

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -29,6 +29,9 @@ import TextInput from './components/TextInput'
 import Touchable from './components/Touchable'
 import View from './components/View'
 
+// modules
+import NativeModules from './modules/NativeModules'
+
 const ReactNative = {
   // apis
   Animated,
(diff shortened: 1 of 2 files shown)


necolas:
Thanks. This will prevent import errors but not all errors that can originate from use of `NativeModules`. I added a note to the "Known issues" guide and maybe we can encourage library authors to start using `Platform` to hide `NativeMethods` use from web.

PR #87, opened 2016-03-08
Added basic ListView component

IjzerenHein:
This commit adds a basic ListView component implementation. It fully supports the DataSource api which is exposed by ListView.DataSource. This allows for running your react-native ListView code using react-native-web.

Things that are supported:

[X] ListView.DataSource (fully supported)
[X] SectionHeaders (`ListView.props.renderSectionHeader` & `ListView.DataSource.cloneWithRowsAndSections`)
[X] Header (`ListView.props.renderHeader`)
[X] Footer  (`ListView.props.renderFooter`)
[X] Separators (`ListView.props.renderSeparator`)
[X] renderScrollComponent (`ListView.props.renderScrollComponent`)

Things are are not (yet) supported:

[ ] Infinite scrolling (only render content that is within a certain visible view)
[ ] Only re-render changed rows

The following props are defined but not (yet) used anywhere:

```js
{
  initialListSize: PropTypes.number,
  onEndReached: PropTypes.func,
  onEndReachedThreshold: PropTypes.number,
  pageSize: PropTypes.number,
  scrollRenderAheadDistance: PropTypes.number,
  onChangeVisibleRows: PropTypes.func,
  removeClippedSubviews: PropTypes.bool,
  stickyHeaderIndices: PropTypes.arrayOf(PropTypes.number)
}
```

### DIFF
--- a/src/components/ListView/ListViewDataSource.js
+++ b/src/components/ListView/ListViewDataSource.js
@@ -1,0 +1,194 @@
+import invariant from 'fbjs/lib/invariant'
+
+export default class ListViewDataSource {
+  constructor(params) {
+    invariant(
+      params && typeof params.rowHasChanged === 'function',
+      'Must provide a rowHasChanged function.'
+    )
+    this._rowHasChanged = params.rowHasChanged
+    this._sectionHeaderHasChanged = params.sectionHeaderHasChanged
+    this._getRowData = params.getRowData || defaultGetRowData
+    this._getSectionHeaderData = params.getSectionHeaderData || defaultGetSectionHeaderData
+
+    this._dataBlob = null
+    this._dirtyRows = []
+    this._dirtySections = []
+    this._cachedRowCount = 0
+
+    this.rowIdentities = []
+    this.sectionIdentities = []
+  }
+
+  cloneWithRows(dataBlob, rowIdentities) {
+    const rowIds = rowIdentities ? [rowIdentities] : null
+    this._sectionHeaderHasChanged = this._sectionHeaderHasChanged || defaultSectionHeaderHasChanged
+    return this.cloneWithRowsAndSections({s1: dataBlob}, ['s1'], rowIds)
+  }
+
+  cloneWithRowsAndSections(dataBlob, sectionIdentities, rowIdentities) {
+    invariant(
+      typeof this._sectionHeaderHasChanged === 'function',
+      'Must provide a sectionHeaderHasChanged function with section data.'
+    )
+    const newSource = new ListViewDataSource({
+      getRowData: this._getRowData,
+      getSectionHeaderData: this._getSectionHeaderData,
+      rowHasChanged: this._rowHasChanged,
+      sectionHeaderHasChanged: this._sectionHeaderHasChanged
+    })
+    newSource._dataBlob = dataBlob
+    newSource.sectionIdentities = sectionIdentities || Object.keys(dataBlob)
+    if (rowIdentities) {
+      newSource.rowIdentities = rowIdentities
+    } else {
+      newSource.rowIdentities = []
+      newSource.sectionIdentities.forEach((sectionID) => {
+        newSource.rowIdentities.push(Object.keys(dataBlob[sectionID]))
+      })
+    }
+    newSource._cachedRowCount = countRows(newSource.rowIdentities)
+
+    calculateDirtyArrays.call(newSource,
+      this._dataBlob,
+      this.sectionIdentities,
+      this.rowIdentities
+    )
+
+    return newSource
+  }
+
+  getRowCount() {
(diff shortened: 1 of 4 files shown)


src/components/ListView/ListViewDataSource.js
```diff
@@ -0,0 +1,194 @@
+import invariant from 'fbjs/lib/invariant'
+
+export default class ListViewDataSource {
```
necolas: why isn't this class copy-pasted from react-native?

IjzerenHein: That particular class was build using flow types incorporated into the code. I (for some reason..) thought that was not supported by the react-native-web toolchain (I had done some spot-checks and could not find any flow enabled code), so instead made a port of it (the code is nearly identical though).
I've just checked and the original code compiles just fine and works as expected. I agree we should use that. I only had to change the require's as the top of the file to use:

```js
var invariant = require('fbjs/lib/invariant');
var isEmpty = require('fbjs/lib/isEmpty');
var warning = require('fbjs/lib/warning');
```

Furthermore, the class throws a ton of linting errors because semi-colons are not allowed. I think it would be nice if we could drop code from 'react-native' in without having to remove all semi-colons (or change the require's).

necolas: react native is changing their fbjs requires to work like standard npm requires. you can add `/* eslint disable */` to the top of the file as is done in other modules mirroring RN

IjzerenHein: Alright cool, I'll do that. It's probably gonna be tomorrow before I have time for this.

IjzerenHein:
Hi, I've updated ListViewDataSource to the copy/pasted version of the original react-native code.
cheers

MoOx:
I am interested by this ListView. Anything missing @necolas?
Maybe https://bvaughn.github.io/react-virtualized/ can help for the scroll/optimized rendering part?

IjzerenHein:
For the record, I'm also working a new virtualized (infinite) scrollview implementation for javascript, supporting (sticky) section-headers, insert/remove/move animations, bottom alignment, etc.. I'll post progress here when I more to show.

MoOx:
Why you don't work on react-virtualized, which seems like a pretty solid start? They have an open PR for the v7 which will remove some legacy things https://github.com/bvaughn/react-virtualized/pull/206

IjzerenHein:
@MoOx Cause I'm working on a different feature-set, I have a set of specific goals in my mind. It's not specifically aimed at react-native-web, but when done we could have a look whether it is a match. In terms of features I'm recreating what can be seen in this demo here:
https://rawgit.com/IjzerenHein/famous-flex-chat/master/dist/index.html

codeheroics:
I'm interested by this ListView component. I tried merging the PR branch and the master branch and it worked like a charm for me after fixing a small conflict, which is great, compared to the current master which throws an error on `new ListView.DataSource()`.

Can't it be merged as it is, and the missing performance-related features added later?

mcampsall:
I love this project, thanks everyone for your contributions! 

I am having difficulties getting the ListView to compile though. I manually created the files in this PR in my project and get the following error when running it:

Module parse failed: /Users/.../node_modules/react-native-web/dist/components/ListView/index.js Unexpected character '@' (10:0)
You may need an appropriate loader to handle this file type.
SyntaxError: Unexpected character '@' (10:0)

it is pointing to this line:
`@NativeMethodsDecorator`

I am sorry if this is not the place for this, and thanks again for all your work.

MoOx:
@mcampsall you need a babel plugin in your stack that will handle decorators as they are not in es2015 (and not in es2016). See https://babeljs.io/docs/plugins/transform-decorators/

mcampsall:
Thanks @MoOx, that would likely work if I had added the files to the project correctly. What would the proper way to added this PR to my project?

necolas:
Thanks! I've merged this to unblock the work on the RN docs (install 0.0.28). We can iterate on ListView from here.

lacker:
Thanks man I owe you one high five 😃

damiaozi:
@MoOx  how to use react-virtualized on web？

MoOx:
Just consume it via npm. Default main of the package.json leads to a commonjs build. You should not have to do anything special. If you need more help, you should read the docs or contact the maintainer of this package, not here :)

fukemy:
hi, how to install this pull request through npm?

MoOx:
ListView has been remove from React Native, so it has been from React Native for web.
- https://github.com/necolas/react-native-web/commit/6c682bbb92920555b3cd7c25f5213d45cb34edc7
- https://github.com/necolas/react-native-web/issues/1352
- https://github.com/facebook/react-native/issues/23313

PR #94, opened 2016-03-11
Fixes to PanResponder + added mouse support

IjzerenHein:
Hi,

While working with PanResponder I've found a couple issues and ran into the problem that mouse-events are not yet supported. This pull request adds mouse-events support and also fixes these issues with touch-events. Most notably, `locationX` and `locationY` were not yet supported, as well as the `timestamp` was missing in the `touches` and `changedTouches` data. This caused `ResponderTouchHistoryStore` to not record the timestamp of the touches.

It now works as expected, what you see here in the demo is the `react-native-slider` component.

slider

cheers

### DIFF
--- a/src/apis/PanResponder/index.js
+++ b/src/apis/PanResponder/index.js
@@ -6,6 +6,7 @@
 
 "use strict";
 
+import normalizeNativeEvent from './normalizeNativeEvent';
 var TouchHistoryMath = require('./TouchHistoryMath');
 
 var currentCentroidXOfTouchesChangedAfter =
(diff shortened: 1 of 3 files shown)


src/apis/PanResponder/index.js
```diff
@@ -348,18 +354,19 @@ var PanResponder = {
         // Guard against the dispatch of two touch moves when there are two
         // simultaneously changed touches.
         if (gestureState._accountsForMovesUpTo === touchHistory.mostRecentTimeStamp) {
+          console.log('bail out')
```
necolas: need to remove this :)

src/apis/PanResponder/normalizeNativeEvent.js
```diff
@@ -0,0 +1,84 @@
...
+    // https://stackoverflow.com/questions/26177087/ios-8-mobile-safari-wrong-timestamp-on-touch-events
+    timestamp: Date.now(),
+    touches
+  }
+
+  if (changedTouches[0]) {
+    event.identifier = changedTouches[0].identifier
+    event.pageX = changedTouches[0].pageX
+    event.pageY = changedTouches[0].pageY
+    const rect = changedTouches[0].target.getBoundingClientRect()
```
necolas: This DOM API contributes to layout trashing. In this case it's being called for every event. I'd be interested finding out from anyone with more knowledge of the React/PanResponder events if there's a strategy we could use to mitigate potential perf issues. Any thoughts about this @vjeux?

necolas:
Thanks this is great! I'm ok with modifying `PanResponder` at this stage. We might be able to upstream any changes if Facebook intend to make it available in `React` at some point.

src/apis/PanResponder/normalizeNativeEvent.js
vjeux: Layout trashing is happening when you do write read write. Everytime you read it has to flush and do a layout pass.

If you do a read before doing any write it should be free and same for doing a read after all the writes of the frame. The second case will likely move the cost to you since it has to do layout but the browser would do it anyway just after.

This isn't a problem per se, what we need to figure out is where it is placed compared to the write operations.

necolas: Thanks, I see.

src/apis/PanResponder/index.js
IjzerenHein: Whoops, that accidentally slipped in :) I ran the linter and was expecting it to catch that case. I've removed it and updated the pull request :)

src/apis/PanResponder/normalizeNativeEvent.js
IjzerenHein: I agree that polling the DOM should be the absolute last option to try. However in this case I don't see any alternatives. If we don't then `locationX` and `locationY` will not be available on touch devices.

necolas:
How did you get react-native-slider working? Did you also patch `onLayout`?

IjzerenHein:
I did, but it's very much a work in progress. It's gonna be very difficult to get that to work correctly and efficiently as the browser is doing all the layout work.

necolas:
Yeah coordinating and minimizing when we do reads will be tricky. I thought we might be able to adapt some of what Polymer is doing to determine when an element's layout needs to be provided.

BTW, it looks like we can also drop the event normalization that was hacked into `View`.

IjzerenHein:
Yeah I've taken a short look at that too. I still need to dive into it more to figure out how and if we can integrate something like that.

Oh, was event normalization also necessary for View?

PR #109, opened 2016-03-23
Replace decorators with higher-order functions

tuckerconnelly:
Also, can you add *.log to gitignore?

OH and, how do you feel about moving the babel preset to babel-preset-react-native?


necolas:
Why should I merge this patch?
You can add `*.log` to your machine's gitignore.

tuckerconnelly:
Whoops, I thought for some reason decorators got cut from ES6 (just Babel 6 I guess). I'll close it.

PR #111, opened 2016-03-25
Update README.md to import from react-native-web

rogeliog:
Shouldn't this be `react-native-web` instead of `react-native`?


rogeliog:
Oops, never mind I just read [guides/react-native.md#react-native](https://github.com/necolas/react-native-web/blob/master/docs/guides/react-native.md#react-native).

PR #113, opened 2016-03-26
Support className

tuckerconnelly:
I'm writing a lib that needs className, but the current StyleSheet api overwrites it. This merges the current className with StyleSheet classNames instead of overwriting it.


VinceBT:
+1

necolas:
Fixed by a16e542bd8c2f3f359c09fd8bf6d62afee182749

PR #126, opened 2016-04-19
fix import statement in sample

kevinbarabash:


necolas:
This is intentional and mentioned in the rendering guide https://github.com/necolas/react-native-web/blob/master/docs/guides/rendering.md. The general idea is that you can work with and export packages built for React Native – this is a partial web backend basically.

kevinbarabash:
Sorry... I was skimming and copy/pasted the sample without reading the important parts.

PR #152, opened 2016-06-18
add Platform.select function to match that provided by react-native

bmcmahen:
you can see the equivalent code in the react-native repo here: https://github.com/facebook/react-native/blob/7c3364196a2a791a01e2a2d9d167bffb6f184ff6/Libraries/Utilities/Platform.ios.js


necolas:
Thanks. Fixed in 2cb68a45bebcb7048177e6a230998c3873197f9e

PR #159, opened 2016-07-03
Add an API to wrap and initialize animated

cesarandreu:
Both `TouchableBounce` and `TouchableOpacity` broke because they expect `Animated.View` to exist on the animated module. I added back `src/apis/Animated/index.js` so it's initialized in a single place.


necolas:
Thanks!

PR #165, opened 2016-07-12
Allow other valid button `type`s when `accessibilityRole` is `button`

apalm:
I might be missing something, but it doesn't currently seem possible to specify e.g. `type='submit'` in conjunction with `accessibilityRole='button'`. This patch adds a check to see if `type` is one of the other possible valid values. The values are taken from [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/button). Let me know if I'm doing the check in the wrong place or you have issues with this in general. Thanks!

### DIFF
--- a/src/modules/createReactDOMComponent/__tests__/index-test.js
+++ b/src/modules/createReactDOMComponent/__tests__/index-test.js
@@ -29,6 +29,18 @@ suite('modules/createReactDOMComponent', () => {
     assert.equal(element.is('button'), true)
   })
 
+  test('prop "type"', () => {
+    const accessibilityRole = 'button'
+    let element = shallow(createReactDOMComponent({ accessibilityRole }))
+    assert.equal(element.prop('type'), accessibilityRole)
+    assert.equal(element.is('button'), true)
+
+    const type = 'submit'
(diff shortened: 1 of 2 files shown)


necolas:
I'm not sure any other type is needed as you can use js to submit

apalm:
Fair enough. Feel free to close if you don't want to take this.

MoOx:
I am interested about having option for submitting via JS directly.
It's would be pretty convenient to be able to use a sort of type=submit on TouchableHighlight in a form. Using onPress on a TouchableHighlight to call the onSubmit callback of the form is not straightforward as native onSubmit event on form receive form data (and onPress, obviously, does not). Maybe I am just missing something because I am tired today :) Any comment will be appreciated.

MoOx:
To give more explanation:
I am using https://github.com/mozilla-services/react-jsonschema-form and the provided "Form" component which use a specific "onSubmit" callback, that send data directly on the callback. So I cannot just call my handleSubmit on both onSubmit (Form) and onPress (TouchableHighlight) because it expects data. That's why for me the "type=submit" behavior would be more than handy (otherwise I have to use vanilla js with a ref and an onPress that call ref form native submit(), which kind of suck, a lot ^^).

PR #174, opened 2016-07-27
RTL layout support

necolas:
Adds support for RTL

1. Auto-flips specific properties
2. Opt-out of auto-flip by adding `$noflip` suffix to prop name

TODO:

- [x] tests
- [x] documentation (inc $noflip properties, `I18nManager`, `dir` prop)
- [x] how to set RTL state in `I18nManager`

### DIFF
--- a/docs/apis/I18nManager.md
+++ b/docs/apis/I18nManager.md
@@ -1,0 +1,27 @@
+# I18nManager
+
+Control and set the layout and writing direction of the application. You must
+set `dir="rtl"` (and should set `lang="${lang}"`) on the root element of your
+app.
+
+## Properties
+
+**isRTL**: bool = false
+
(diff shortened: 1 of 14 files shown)


necolas:
OK I think this is ready to land. I'm not sure about the `noflip` name, perhaps `noI18n` would be better.

giuseppeg:
<blockquote>I'm not sure about the noflip name, perhaps noI18n would be better.</blockquote>
what about `i18n_preserve`, `preserveDir` or just `preserve` also wouldn't it be better to use a comment instead of producing an invalid value?

necolas:
This was added in 6416166bc3c40cb6a48c99f9fe270b0b459056c3 and fb5406e4ec28fa21b398e89ac5672a1a655540c9

PR #179, opened 2016-08-02
Remove blue outline on Chrome for Touchables and add userSelect none for TouchableWithoutFeedback

VinceBT:


necolas:
My concern with removing the outline is losing a basic web accessibility feature.

Why should `userSelect:none` be added to `TouchableWithoutFeedback`?

VinceBT:
It only happens in Chrome and people usually want to remove the outline because it doesn't look good in applications.
All others Touchables have userSelect: none and I think that when an element is clickable even if it has no feedback, should not be selectable.

necolas:
Like I said, this is an important accessibility feature on the web and most people are not going to think about reintroducing it. In the meantime, you can remove the `outline` in your app code.

`TouchableWithoutFeedback` has many different use cases and it's not obvious to me that selection should be disabled by default.

PR #200, opened 2016-09-07
export modules and propTypes to enable development of RTCView:

robblovell:

I needed to develop a react-native web tag for a WebRTC implementation for react-native.  I wanted to use the same tag with react-native-web.  The new tag is called "RTCView" and resolves to a <video> html element.  This new tag parallels a react-native-webrtc project that has a tag developed for WebRTC in Android and iOS (https://github.com/liivevideo/react-native-webrtc). 

In order to implement <RTCView> outside the react-native-web project and within a reat-native-web-rtcweb project, I needed to have access to two modules and two propTypes within react-native-web: 

/modules/createDOMElement
/modules/applyNativeMethods
/propTypes/StyleSheetPropType
/propTypes/BaseComponentPropTypes

**This pull request**

I have include these exports in index.js so that they can be accessed by the new RTCVideo tag in react-native-web-webrtc.  See https://github.com/liivevideo/react-native-web-webrtc for the implementation of RTCVideo.  See https://github.com/liivevideo/react-native-webrtc for the sister project that implements WebRTC in react-native.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -38,12 +38,16 @@ import View from './components/View';
 
 // modules
 import NativeModules from './modules/NativeModules';
+import createDOMElement from './modules/createDOMElement'
+import applyNativeMethods from './modules/applyNativeMethods'
 
 // propTypes
 
 import ColorPropType from './propTypes/ColorPropType';
 import EdgeInsetsPropType from './propTypes/EdgeInsetsPropType';
 import PointPropType from './propTypes/PointPropType';
+import StyleSheetPropType from './propTypes/StyleSheetPropType';
+import BaseComponentPropTypes from './propTypes/BaseComponentPropTypes';
 
 const ReactNative = {
   // top-level API
@@ -88,11 +92,15 @@ const ReactNative = {
 
   // modules
   NativeModules,
+  createDOMElement,
+  applyNativeMethods,
 
   // propTypes
   ColorPropType,
   EdgeInsetsPropType,
-  PointPropType
+  PointPropType,
+  StyleSheetPropType,
+  BaseComponentPropTypes
 };
 
 module.exports = ReactNative;


RangerMauve:
Oh perfect! This means I won't have to do this stuff myself when I get to it. :grinning:

necolas:
Your RTCView looks like it copy-pastes a lot of the implementation of 'Image' and isn't functional. I don't think I want to export this many modules that are currently part of the framework's implementation details right now, and I don't think you'd need them all. If you help me understand what the complete user-facing API is meant to be I can think about this a bit more.

RangerMauve:
The proptypes for RTCView come from [here](https://github.com/oney/react-native-webrtc/blob/master/RTCView.js).

It looks like it should just support everything that `View` supports, plus:
* `mirror`: Can be done via CSS with `transform: scaleX(-1)`
* `objectFit`: Just use the CSS property for this
* `streamURL` : This is the only thing that might need to change
* `zOrder`: Just use `z-index` in CSS

robblovell:
Yes, I admit that the RTCView implementation is a copy past job due to my lack of understanding at this point in time.

I will refactor the RTCView as RangerMauve has suggested to get rid of the extra exports.

I will also try a regular video tab inside the View tab.  Thank you for the suggestions. I will reopen if I can't make it work.

necolas:
I do think something like 'createDOMElement' should be exported. There's another issue about that. I may add it as an undocumented export in 0.0.46

robblovell:
The basic need is to have a "<video>" tag with src=streamURL:string and autoPlay:bool. There are other attributes should be supported as suggested (mirror: bool, zoreder: number, objectFit: contain/cover).  

I don't know enough to know what to tear out and what to inherit from the framework. It's basically working, but with a lot of waste that should be cut out of the implementation as I see it now.

I made it work by copying the Image tag and adding some props, which was a start on learning how it all works. Perhaps I should have started with an minimum implementation first...

createDOMElement would be useful ...

PR #203, opened 2016-09-14
Corrected Image docs for resizeMode prop default

lprhodes:
Updating the Image doc as the default resizeMode value appears to be cover, not stretch.
See https://github.com/necolas/react-native-web/blob/master/src/components/Image/index.js#L96

**Test plan**

**This pull request**

- [x] includes documentation
- [ ] includes tests

### DIFF
--- a/docs/components/Image.md
+++ b/docs/components/Image.md
@@ -46,7 +46,7 @@ Invoked when load either succeeds or fails,
 
 Invoked on load start.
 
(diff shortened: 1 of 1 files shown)


necolas:
Do you know what the default in React Native is? rnplay.com is down so I can't check easily. It might be that I've got the default resizeMode wrong in the code.

necolas:
Thanks. Applied in 4576b4236569a2088c045e5451408c84085d18ce

PR #204, opened 2016-09-15
ListView fixes for contained ScrollView

PaulLeCam:

The current ListView implementation (v0.0.45) fails to proxy the relevant calls to its contained ScrollView, this seems due to two things:

- the `_setScrollViewRef()` function is not bound to the class, so `this` points to `window` rather than the ListView instance
- no props are passed to the ScrollView, the arguments provided to [Lodash's `pick()`](https://lodash.com/docs/4.15.0#pick) do not match the expected signature

This PR addresses these two issues.

**Test plan**

No existing test for ListView to update, so only local test usage

**This pull request**

- [ ] includes documentation
- [ ] includes tests

### DIFF
--- a/src/components/ListView/index.js
+++ b/src/components/ListView/index.js
@@ -2,10 +2,11 @@ import applyNativeMethods from '../../modules/applyNativeMethods';
 import ListViewDataSource from './ListViewDataSource';
 import ListViewPropTypes from './ListViewPropTypes';
 import pick from 'lodash/pick';
-import keys from 'lodash/keys';
 import ScrollView from '../ScrollView';
 import View from '../View';
 import React, { Component } from 'react';
+
+const scrollViewProps = Object.keys(ScrollView.propTypes);
 
 class ListView extends Component {
   static propTypes = ListViewPropTypes;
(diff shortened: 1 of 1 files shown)


src/components/ListView/index.js
```diff
@@ -88,14 +89,14 @@ class ListView extends Component {
       }
     }
 
-    const props = pick(ScrollView.propTypes, this.props);
+    const props = pick(this.props, keys(ScrollView.propTypes));
```
necolas: Please use `Object.keys` and move the creation of the array outside of `render` since we don't need to recompute it. See https://github.com/necolas/react-native-web/blob/d64df129b2a7f0003e9614549ae70fbcfb734f9e/src/components/TextInput/index.js#L15

PaulLeCam: Updated

necolas:
Thanks!

codeheroics:
@necolas Do you plan on doing a new react-native-web release including this fix soon? It fixes a big display problem on a test app I'm working on thanks to props being passed down correctly now :)

necolas:
I've published 0.0.46 with this fix

codeheroics:
Thanks !

PR #219, opened 2016-10-07
Add doc section on rendering within a React webapp

flegall:
This section was missing and we spent some time implementing a very simple wrapper that setup/teardown React Native Web in componendDidMount() / componentWillUnmount. 
It turned out to be unnecessary since it actually just worked out of the box !

**This pull request**

- [X] includes documentation

We use React Native Web as a preview tool in our back-office application at fluo.com so that our back-office operators can preview how it will render on the mobile app ! It works really great !


necolas:
Hey, I'm not sure what this documentation is addressing. The library is an interface on top of React/ReactDOM. Going to close this for now.

Issue #227, opened 2016-10-19
Image: scroll performance issues caused by background-image CSS property


faceyspacey:
Because `Image` uses `background-image`, when scrolling lots of images, performance is destroyed. Here's a non-RNW post regarding the issue with this css property in general:

http://stackoverflow.com/questions/33229840/chrome-is-lagging-when-using-css-background-sizecover-in-scrolling-background-i

The `translateZ` hardware-acceleration trick mentioned there hardly does anything in this case. And the issue isn't just about `background-size: cover`. I've created my own `<Image>` class and tried every combination. Just having `background-image` is enough to kill performanec for me. I have 50 images on the page. With under 20 images you won't notice it unless you have something to compare to. If you do try the alternative and create your own `<Image>` class which doesn't use background images, you will definitely notice a major different on just those 10 images.  *Now*, for all 50, the difference is between a page that barely scrolls and one that scrolls properly. It's a major problem--you basically can't use the `<Image>` component that comes with RNW for more than a few images. 

The only option I see is rolling your own that can't properly center an image like you do with `background-size: cover`, or otherwise RNW will be incompatible with the `resizeMode` expectations of standard RN.

necolas:
The joys of the web platform. Thanks for tracking this issue down. Is it an issue in all browsers or just Chrome?

necolas:
FWIW, we use background-image for all media previews on mobile.twitter.com. Perhaps we don't experience this issue because we avoid rendering off-screen timeline items.

RangerMauve:
The stackoverflow link says that doing a 3D transform to move the layer to the GPU is supposed to help. Would that be a good idea to pursue, or should the `background-image` properly just be replaced entirely?

necolas:
Didn't you mention that the 3D transform didn't help? Some of the display modes require background-image to work (e.g., repeat)

faceyspacey:
@RangerMauve I've done the `translateZ` thing and in this case, I couldn't notice anything. If you check the *Paint Flashing* box in Chrome's *Rendering Settings* you will see the paint still flashing when scrolling.  Basically with the `background-image` property, it requires re-paints on scrolling. It seems as simple as that. 

Here's some screenshots of the Rendering Settings in action:

https://www.dropbox.com/s/u4akheugmj331yh/Screenshot%202016-10-19%2014.52.21.png?dl=0

[...]

PR #228, opened 2016-10-20
merge translateZ: 0 into Image container

faceyspacey:
The `background-image` style property is a performance bottleneck for rendering. By adding `translateZ: 0` we can trigger hardware acceleration, which is what this PR provides. 

NOTE: `translateZ: 0` is only added if not already present, so as not to override any value it might already contain.

REFERENCE ISSUE: https://github.com/necolas/react-native-web/issues/227

### DIFF
--- a/src/components/Image/index.js
+++ b/src/components/Image/index.js
@@ -96,6 +96,13 @@ class Image extends Component {
     const backgroundImage = displayImage ? `url("${displayImage}")` : null;
     let style = StyleSheet.flatten(this.props.style);
 
+    //translateZ property triggers hardware acceleration for better rendering of the background-image property
+    const hasTranslateZ = style.transform && style.transform.reduce((acc, t) => acc || t && t.hasOwnProperty('translateZ'), false);
+    if(!hasTranslateZ) {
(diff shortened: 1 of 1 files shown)


necolas:
I can't reproduce the problem this is meant to fix. Test case: [100 images in a scroll view](https://codepen.io/necolas/full/XjOaVw). Does this issue only manifest if you're animating images while scrolling?

faceyspacey:
Not only with the animations. I turned them off and tried it too. I have more tests coming up (as a natural progression of the actual project). The images are very large in size, so that could contribute to. So after I have them shrunk and compressed, I'll check it again. But that said, I'm pretty certain that the more you throw at it, performance degrades at a lot faster rate than images with sizes for their width and height. So, it probably won't end up being one thing that contributes to the problem.

RangerMauve:
Wouldn't it make sense to put translateZ on the container that's a parent to _all_ the images so that they all get placed on one layer? Or am I misunderstanding how that optimization works?

necolas:
`Image` was rewritten to only use a single div. This fixed a major perf issue in Safari. Try 0.0.50 and see if things are any better. Closing this for now

PR #231, opened 2016-10-20
fix for placeholder not returning when programmatically cleared

faceyspacey:
currently the placeholder does not come back if `state.value` is programmatically set to '' or if `clear()` is called

### DIFF
--- a/src/components/TextInput/index.js
+++ b/src/components/TextInput/index.js
@@ -64,6 +64,7 @@ class TextInput extends Component {
   }
 
   clear() {
+    this.setState({showPlaceholder: true})
     this.setNativeProps({ text: '' });
   }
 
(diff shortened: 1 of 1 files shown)


src/components/TextInput/index.js
```diff
@@ -64,6 +64,7 @@ class TextInput extends Component {
   }
 
   clear() {
+    this.setState({showPlaceholder: true})
```
RangerMauve: Should this maybe also set text to `""`?

faceyspacey: the nativeProps? it does the line below, as it did before.

necolas:
This shouldn't be an issue anymore after 89f5a13891a5c99f5a25bfb95ee8d54a91595b4f

PR #232, opened 2016-10-21
proposal for touchables as position: static

faceyspacey:
In RN proper, Touchables can be positioned absolutely or statically by positioning the children of a `Touchable`. It's as if the Touchable container serves no other purpose other than to animate the children (opacity, highlight) and provide `onPress` handlers. However, in RNW, all views are position `relative`, therefore children when absolutely positioned move relative to the generated touchable `<button>`, rather than the first position relative parent view that contains the touchable component. This change will need to be applied to TouchableHightly, TouchableWithoutFeedback, etc. It will also be a breaking change for developers who utilized absolute positioning in relation to the incorrectly position relative button, but it will make RNW consistent with RN.

### DIFF
--- a/src/components/Touchable/TouchableOpacity.js
+++ b/src/components/Touchable/TouchableOpacity.js
@@ -200,7 +200,8 @@ var TouchableOpacity = React.createClass({
 var styles = StyleSheet.create({
   root: {
     cursor: 'pointer',
-    userSelect: 'none'
+    userSelect: 'none',
+    position: 'static',
   },
   disabled: {
     cursor: 'default'


necolas:
Hey @vjeux, what is going on in React Native for touchable children to behave this way?

vjeux:
This is definitely not intended to behave that way in React Native. We should fix it there.

necolas:
Ah good to know! Thanks

@faceyspacey, if I've understood the issue correctly you may want to report this against the react-native repo. Thanks for helping to identify these inconsistencies!

Issue #151, opened 2016-06-17
AppState should not be using functions as property names


ckknight:
Given the following code:
```js
function run() {
  function handler() { /* ... */ }
  AppState.addEventListener('change', handler);
  return function () {
    AppState.removeEventListener('change', handler);
  }
}
```

Because `AppState` is managing its listeners in [an object](https://github.com/necolas/react-native-web/blob/6a73d77030a8b48afb5f3db780409e451145d2a9/src/apis/AppState/index.js#L19), the `handler` above will use its `.toString()` value as the key of `listeners`, which is equivalent between function calls.

One solution would be to use a `Map`, thus removing the implicit `[[ToString]]`.

PR #246, opened 2016-11-03
[fix] NetInfo event handler registration

paularmstrong:

Similar to #151 and [this commit](https://github.com/necolas/react-native-web/commit/af40f98f2380861a95f8219557384b497c0dee95), `NetInfo.isConnected` was creating new functions when adding event listeners and, subsequently, removing them, thus, never actually removing the listeners.

**Test plan**

Added tests.

**This pull request**

- [ ] includes documentation
- [x] includes tests


necolas:
Thanks! Merged 4beae0dd78b82b9e6d7d98bf431582606c220d16

PR #248, opened 2016-11-04
[feature] Use yarn for dependency management

paularmstrong:

Getting up and running with npm takes 5+ minutes to resolve and install dependencies.

Using yarn took 58s.

**Test plan**

* [install yarn](https://yarnpkg.com/en/docs/install)
* run `yarn` or `yarn install`

**This pull request**

- [x] includes documentation
- [ ] includes tests


RangerMauve:
I'm not sure what the point in having a library that's meant to be incorporated as part of a larger project tell you to use a particular npm CLI. This really seems like something that users of the package should decide on. Wouldn't the lock file interfere with de-duplication of other packages that users install once they've installed `react-native-web`?

paularmstrong:
@RangerMauve This shouldn't cause any changes for people using `react-native-web` in their projects. This is for development of the `react-native-web` project only. You'll notice that the dependencies in `package.json` have not changed at all due to this and `yarn` will continue to update `package.json` accordingly, just as `npm install --save` would.

RangerMauve:
@paularmstrong with regards to it being intended for use in development, is the lockfile really required to be part of the version control? And what benefit does it bring to having everybody using yarn when they develop?

paularmstrong:
@RangerMauve it doesn't *need* to be checked in, but is recommended by the [documentation](https://yarnpkg.com/en/docs/yarn-lock) 

> All yarn.lock files should be checked into source control (e.g. git or mercurial). This allows Yarn to install the same exact dependency tree across all machines, whether it be your coworker’s laptop or a CI server.

(also note on the link provided information about top-level yarn.lock file: this lock file will be ignored by projects that are using react-native-web as a dependency)

RangerMauve:
Are there good descriptions on when things should be updated and how? For instance, when users install the project, they might have slightly different versions of modules, which would mean that whoever is debugging the problem might need to install different versions. Then again, I guess the user filing the bug could share their lockfile.

necolas:
This only effects those of us developing the library, not people installing the package from npm

necolas:
@paularmstrong LGTM. Going to wait for web-tools to land Yarn internally

necolas:
See 8352c7cbda18679c9c885736340ca86eb31a94e8

PR #255, opened 2016-11-17
Upgrade to react@15.4.0

MoOx:

I tried to update react to latest 15.4.0 to be able to use jest without [this blocking issue ](https://github.com/facebook/react/issues/7386) but it seems RNW use some /lib/* stuff.

**Test plan**

I just upgraded all reference to react-dom. With react 15.4.0, some lib/* redundant in both react and react-dom packages but I think it's better to rely on a single package so I just switched all references to `react/lib/` to `react-dom/lib/` and it worked.

**This pull request**

- [ ] includes documentation
- [x] includes tests

Note: Tests were breaking before this update, with the changes included in this PR they pass.


necolas:
I tried this locally and found that the touchables stopped working in the examples.

necolas:
The problem seems to be that the `topLevelType` in [ResponderEventPlugin](https://github.com/facebook/react/blob/308e0b77863071f31cff7aec74e359603526274e/src/renderers/shared/shared/event/eventPlugins/ResponderEventPlugin.js#L319) - used to be `topMouseDown` or `topTouchStart` but is now `topClick`. The `isStartish` [helper](https://github.com/facebook/react/blob/308e0b77863071f31cff7aec74e359603526274e/src/renderers/shared/shared/event/eventPlugins/ResponderEventPlugin.js#L325) now returns `false` (because it's only check for mouse/touch events) in 15.4 where it used to return `true`.

necolas:
Sigh. Because React@15.4.0 is going to be a giant ball of pre-built code, we might need to update the usage instructions to suggest aliasing 'react' to 'react/lib/React'. I think this will let your bundler de-dupe all the modules shared between 'react' and 'react-dom'.

remon-georgy:
Thanks for your effort to make the upgrade work. The update is especially necessary because running tests while using react < 15.4 is currently [impossible.](https://github.com/facebook/jest/issues/1353)

necolas:
IIRC, not impossible if you include `jest.mock('react-dom')`

remon-georgy:
I just realised the @MoOx raised the issue originally for the same purpose. Thanks for the tip @necolas. Hopefully we get a feedback about facebook/react#8370 soon.

nicholeuf:
Any idea when this will get merged?

necolas:
When facebook/react#8370 is resolved

necolas:
@MoOx would you mind rebasing and updating the branch. May make it easier to debug the (possible) React issue.

MoOx:
PR updated. 

(There are some changes in the yarn.lock file, I guess it's normal since we changed some packages - I just "yarn"ed once before running the tests).

PR #256, opened 2016-11-17
Provide source files (allow npm install via git with a building step)

MoOx:
Without this change, if you use npm to install the repo (or a fork 😉) you get a folder with just a package.json, which sucks.

With this change, you get the source files too and allow to build yourself “dist” (using `cd node_modules/react-native-web && npm install`, build is in prepublish and prepublish is run in install with npm@2/3, with npm@4 you need to add `npm run build`)

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,7 +4,9 @@
   "description": "React Native for Web",
   "main": "dist/index.js",
   "files": [
-    "dist"
+    "dist",
+    "src",
+    "!**/__tests__"
   ],
   "scripts": {
     "build": "del ./dist && mkdir dist && babel src -d dist --ignore **/__tests__",


MoOx:
Tests fails because of react@15.4.0. See #255.

FYI, "npm run build" fail because of missing webpack.config.js, but at least you have the files transpiled. Alternatively, I guess we can consume the src files directly.

RangerMauve:
Wouldn't this make the package size a lot larger? And couldn't people wanting to build from source just make use of the git repo?

MoOx:
> Wouldn't this make the package size a lot larger?

Obviously yes, but just .js files, so not that much imo.

> And couldn't people wanting to build from source just make use of the git repo?

It's not easy when you share your project with teammates (or I don't have an easy answer)

necolas:
Why do you need the source files?

MoOx:
`npm install https://github.com/MoOx/react-native-web#react-15.4.0` works, but take in consideration the package.json "files" section, so it remove everything that is not "dist" (except LICENSE, package.json and README).

necolas:
I don't understand why you're trying to build the package in your project. If you're trying to use a subset of the modules, you can import them directly from 'dist' into a local module and alias that to `react-native`. For tree-shaking / es-next support I'm planning to include modules for that in the package at some point. Closing this for now.

MoOx:
If you want to use a fork (eg react 15.4 partial support - I don't use touchable), the easiest way is to install from git. But when installing from git, .npmignore is considered so you end up with an empty folder.

necolas:
OK doesn't hurt to make that easier. Added in 945fff00159f521becdb499439e38538889313cd

Issue #272, opened 2016-12-01
Console warnings for shadow* props on View elements


thetable:
**What is the current behavior?**

For a `View` element like this:

```
<View style={styles}>
  <Text>Text</Text>
</View>
```

where `styles` are:

```
const styles = StyleSheet.create({
  shadow: {
    shadowColor: '#000000',
    shadowOpacity: 0.2,
    shadowRadius: 10,
    shadowOffset: {
      height: 10,
      width: 0
    }
  }
});
```

react-native-web prints console warnings like this:

`"Warning: 'shadowColor' is not a valid style property.`

as demonstrated here: http://codepen.io/thetable/pen/bBaNaG?editors=1011

**What is the expected behaviour?**

As shadow props are supported properties on `View` elements (see https://facebook.github.io/react-native/docs/view.html) and react-native-web correctly translates these into CSS, no warnings should be shown.  

**Environment (include versions)**

OS: MacOS 
Device: Mac
Browser: Chrome
React Native for Web (version): latest
React (version): 15.3.1

necolas:
These warnings are intentional because the RN shadow props aren't implemented and don't map cleanly to CSS's `box-shadow`.

PR #273, opened 2016-12-01
[fix] Add shadow style prop types to View styles

thetable:

Fixes console warnings when supplying `shadow*` style props to a `View` element (https://github.com/necolas/react-native-web/issues/272)


necolas:
This only hides the warnings but doesn't produce valid CSS. The problem with the shadow style props is that they don't map cleanly to CSS `box-shadow`

necolas:
Closing because this only hides the warnings, but doesn't implement the shadow props. The warnings are there to make clear that shadow props don't do anything on Web.

PR #279, opened 2016-12-07
Include support to react v15.4

thiagofelix:
Breaking changes with react v15.4.
The package react-dom isn't bundled with react anymore therefore its modules can't be accessed via `require('react/lib/*')`

**Test plan**
Previously react-native-web apps wouldn't work when running along side react v15.4+ version.
Test plan would be simply trying to run any of the existing example/demo apps and expect it not to break.

**This pull request**

- [ ] includes documentation
- [X] includes tests

I am not sure if these changes need to be informed somewhere in the documentation, please let me know if something is missing.

Thanks


MoOx:
Dup of #255

thiagofelix:
Thanks @MoOx for pointing out. 
Closing this here.

PR #281, opened 2016-12-08
Added react-native-web-webpack-starter

agrcrobles:
Hi,

Adding starter.

**This pull request**

- [ x ] includes documentation


necolas:
Any reason why your project can't be combined with the other starter kit? I think CRA and the existing RNW starter are enough at the moment. Thanks though

agrcrobles:
because I want to test it as a hosted app on production and hot-reload on development and add flow as it is TicTacToe.js

PR #283, opened 2016-12-12
Uses css classes for styling instead of inline styles

tuckerconnelly:

Compatibility with [Uranium](https://github.com/tuckerconnelly/uranium) and ultimately [Carbon UI](https://github.com/tuckerconnelly/carbon-ui).

More generally, adds compatibility with classnames. Currently, any library that uses classnames to apply styles, or any user-provided css classes, will get overridden by RNW's inline styles.

This also pretty significantly reduces the css footprint, and I wouldn't be surprised if it also improved performance. material-ui noticed a [25% decrease in performance](https://github.com/callemall/material-ui/blob/master/ROADMAP.md#summarizing-what-are-our-main-problems-with-css) when moving to inline styles from classes.

**Test plan**

Updated the jest test snapshots, and confirmed it worked with the [carbon ui documentation site](https://carbon-ui.com).

**This pull request**

- [ ] includes documentation
- [x] includes tests


necolas:
> material-ui noticed a 25% decrease in performance when moving to inline styles from classes

Noticed similar on an internal mobile.twitter.com branch. Now that this library essentially has full support for RN styles and i18n, we can go back to introducing a more efficient `className` implementation that operates at the declaration level (which styletron recently emulated).

remon-georgy:
I'm still new to React-native, so my question is mainly to learn. 

I had the impression that js inline styling solves some CSS issues; such as eliminating dead code, avoid global scoping and minification. Also IMO it's a gateway to lots of fun programmable styling, such as [react-native-extended-stylesheet](https://github.com/vitalets/react-native-extended-stylesheet).

I'm wondering what advantages does CSS bring. Also it's still possible with this merge request to use inline styling besides CSS.

remon-georgy:
Alright, checked the merge request as well as the underlying library (classnames). Seems that inline styling will still work and that library is actually interesting. Thanks!

necolas:
I'm closing this because it doesn't really solve the problem and creates others. Any styles set via RNW will still override the styles set via className. It only works if you use `AppRegistry`. And the classnames aren't obfuscated. I'll be working on a solution to the problem you raise. Thanks

necolas:
Hey, try using 0.0.62. Styles are resolved to className where possible—which should allow you to use Uranium—and rendering performance is about 13x better against the benchmarks.

tuckerconnelly:
Hey yup, it works :) Thanks. It is a little difficult to debug directly though, would be nice if in the future there could be a single classname per element.

PR #284, opened 2016-12-12
Add WebkitTapHighlightColor to ViewStylePropTypes

tuckerconnelly:

Adds [WebkitTapHighlightColor](https://developer.mozilla.org/en-US/docs/Web/CSS/-webkit-tap-highlight-color) to valid ViewStylePropTypes. This is necessary for a good mobile-web experience because the default color is grey and is applied even if you're providing touch feedback yourself (like with TouchableRipple in carbon-ui).

### DIFF
--- a/src/components/View/ViewStylePropTypes.js
+++ b/src/components/View/ViewStylePropTypes.js
@@ -35,5 +35,6 @@ module.exports = process.env.NODE_ENV !== 'production' ? {
   transition: string,
   userSelect: string,
   visibility: hiddenOrVisible,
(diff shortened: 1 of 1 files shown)


necolas:
Shouldn't the tap highlight color be getting removed by https://github.com/necolas/react-native-web/blob/5a5707855b236e4cae5983aedcb75cd077007088/src/apis/StyleSheet/css.js#L11?

tuckerconnelly:
It totally is :P Thanks

PR #301, opened 2016-12-27
mock for libraries that uses BackAndroid

vaukalak:
react-native [has a mock for BackAndroid in ios version](https://github.com/facebook/react-native/blob/9ee815f6b52e0c2417c04e5a05e1e31df26daed2/Libraries/Utilities/BackAndroid.ios.js), my version is just a copy/paste of it.

A lot of libraries at least ([react-native-router-flux](https://github.com/aksonov/react-native-router-flux) and [react-native-action-sheet](https://github.com/exponent/react-native-action-sheet)). This patch helps those libraries working smooth in react-native-web (avoiding runtime errors).

### DIFF
--- a/src/apis/BackAndroid/index.js
+++ b/src/apis/BackAndroid/index.js
@@ -1,0 +1,28 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ * All rights reserved.
+ *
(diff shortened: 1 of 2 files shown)


necolas:
Added by ba75acb66a6403df20020f7541462e17518f1990

PR #311, opened 2017-01-01
[fix] Support newline in Text using {'\n'}

### LINKED ISSUE #302: Support {'\n'} in Text

Unlike RN, RNW doesn't support creating newlines using `{'\n'}`.

Link to minimal test case: http://facebook.github.io/react-native/docs/text.html

This shows `{'\n'}` not working in official RN docs.

**What is the expected behaviour?**

`{'\n'}` within `Text` should create a newline, i.e. `{'\n'}` should be replaced with something like `<br />`.

nthtran:

Supports newlines in Text using `{'\n'}`.

Fixes #302.

**Test plan**

**This pull request**

- [ ] includes documentation
- [ ] includes tests

### DIFF
--- a/src/components/Text/index.js
+++ b/src/components/Text/index.js
@@ -78,7 +78,8 @@ const styles = StyleSheet.create({
     margin: 0,
     padding: 0,
     textDecorationLine: 'none',
-    wordWrap: 'break-word'
+    wordWrap: 'break-word',
+    whiteSpace: 'pre-wrap'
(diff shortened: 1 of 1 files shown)


necolas:
Thanks :). This change was already made in d87f71ebc11d4ac986ec39a039ec4f7c96916e7d

nthtran:
Awesome thanks!

PR #316, opened 2017-01-04
Fixes for ListView

PaulLeCam:

The ListView component makes calls to undefined functions: `isEmpty()` and `merge()` that it tries to get from React (replaced by Lodash in this PR but could probably be using fbjs) and `ListViewDataSource.getRowAndSectionCount()` (added in this PR copying React-Native implementation).

**Test plan**

**This pull request**

- [ ] includes documentation
- [ ] includes tests

### DIFF
--- a/src/components/ListView/index.js
+++ b/src/components/ListView/index.js
@@ -4,7 +4,7 @@ import ListViewPropTypes from './ListViewPropTypes';
 import ScrollView from '../ScrollView';
 import StaticRenderer from '../StaticRenderer';
 import React, { Component } from 'react';
-import { isEmpty, merge } from 'lodash';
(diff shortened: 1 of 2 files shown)


src/components/ListView/index.js
```diff
@@ -3,7 +3,8 @@ import ListViewDataSource from './ListViewDataSource';
 import ListViewPropTypes from './ListViewPropTypes';
 import ScrollView from '../ScrollView';
 import StaticRenderer from '../StaticRenderer';
-import React, { Component, isEmpty, merge } from 'react';
+import React, { Component } from 'react';
+import { isEmpty, merge } from 'lodash';
```
necolas: If you look over the commit history you'll see I removed lodash in favor of platform features and smaller alternatives

PR #325, opened 2017-01-10
Fixes error in NativeMethodsMixin when style props could not be resolved

psychoo118:
When applying nativeProps to View component it could happen that `nativeProps.style` is `undefined`. In this case calling `StyleRegistry.resolve( nativeProps.style )` will return `undefined`. But in the next line you try to extract className property from the resolvedProps variable(which is undefined at the moment) it will cause an Exception.

Now persistance checking of required properties are made in order to avoid exceptions.


necolas:
Thanks for reporting. I've put a fix up in #327

Issue #260, opened 2016-11-23
ListView implementation


necolas:
Catch-all issue for `ListView`. Leave comments here if you want to report issues or have PRs that improve it.

**What is the current behavior?**

`ListView` is partially implemented and missing many features. It has been reported that this component has performance issues on native platforms, making it unlikely that the core of the RN implementation will perform well on web. I'm not actively working on `ListView`, nor do I plan to work on it given that it doesn't currently help us solve our problems on Web.

There's also an experimental `ListView` replacement being built within RN, much like the replacement for `Navigator`.

gethinwebster:
@necolas, do you have any details on the experimental ListView in RN? On a quick search I wasn't able to find any references to this, and would be interested to find out more, we're currently looking at options for a RN and RNW shared ListView component.

necolas:
It's called `WindowedListView` - https://github.com/facebook/react-native/blob/master/Libraries/Experimental/WindowedListView.js

mkdotcom:
New listview is very much faster, nice 👍 
But I still have this layout issue :
https://github.com/necolas/react-native-web/issues/234

gethinwebster:
@mkdotcom I've done a quick bit of testing, and applying `flexBasis: 0` to the style of the ListView seems to fix it. I've come across similar issues from time to time with other components (and fixed it in the same way), and am not sure whether this is a systematic difference between react-native and react-native-web, or something more subtle than that - I seem to remember that setting `flexBasis: 0` across the board caused even more issues.

Anyway - hopefully that helps a little, here's a pen with the fix applied:

https://codepen.io/gethinwebster/pen/PbgjYw

mkdotcom:
@gethinwebster nice, thanks !

MarkMurphy:
Is `ListView` supposed to only render enough rows to fill the viewport of the component + the `scrollRenderAheadDistance`? (or something along those lines) If so, it doesn't seem to do that...I tried adding a couple thousand rows and it seems to render all of them at once even though it only takes 20 or so to fill the viewport.

MarkMurphy:
Also, getting this error/warning while using `ListView`:

```
[...]

PR #337, opened 2017-01-26
[fix] Prevent props warnings from ScrollView in ListView

gethinwebster:
As discussed in #260

### DIFF
--- a/src/components/ListView/index.js
+++ b/src/components/ListView/index.js
@@ -107,20 +107,42 @@ class ListView extends Component {
   render() {
     const children = [];
 
-    const dataSource = this.props.dataSource;
+    const {
+      dataSource,
+      renderHeader,
+      renderFooter,
+      renderSectionHeader,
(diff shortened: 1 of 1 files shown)


necolas:
We do something similar in other components. I think the preferred way to do this is to use restructuring to create an object without the `ListView` props, and spread that when rendering `ScrollView`:

```
const {
  aListViewProp,
  anotherListViewProp,
  ...other
} = this.props

<ScrollView {...other} />
```

gethinwebster:
As not all of the ListView props are used within the render method (initialListSize, renderRow etc), this will end up creating quite a few unused variables - what would be your preferred way to deal with those?

necolas:
https://github.com/necolas/react-native-web/blob/master/src/components/ScrollView/index.js#L104

necolas:
Thank you. Applied in e06d7a96500c7a674bffac10ed345935f03668a6

PR #343, opened 2017-02-01
Feature - JSS Support

jtag05:
This pull request creates a new module that utilizes the JSS library for the StyleSheet api when loaded though `react-native-web/jss`

- [x] includes documentation
- [x] includes tests
- [x] includes benchmark reports


RangerMauve:
Will this have any impact on code shared between React Native and React Native Web?

jtag05:
@RangerMauve I may not be understanding your question completely. But if you're asking if the StyleSheet API interface has changed and would require rework to be implemented, the answer is no.

RangerMauve:
Awesome! :D

jankjr:
The web target for our project really needs this, as 'hover', 'focus' and the like are pretty important for pointer based input devices. So I would love for this to get merged.

👍

necolas:
Hey, thanks for trying this out. I pulled your branch and had a look. Compared to the existing implementation…

* JSS adds 15 KB to the gzipped size of a bundle
* JSS branch doesn't render all the storybook examples and interactions correctly
* JSS benchmark render times are 5-20% slower*
* JSS benchmark initial heap size is ~10x larger (65 MB vs 6.5 MB)
* JSS benchmark maximum heap size is larger (200 MB vs 32 MB)

The JS heap size and Nodes count often continues to grow throughout the benchmark runs. Suggests a memory leak somewhere.

*MacBook Pro (13-inch, Early 2011) :: 2.7 GHz Intel Core i7 :: 16 GB 1600 MHz DDR3

ndbroadbent:
@jankjr - You can use `onFocus` / `onBlur` props to simulate `:focus`, and `onMouseEnter` / `onMouseLeave` for `:hover`.

necolas:
`className` + normal CSS is also an escape hatch you can use to apply hover styles

jtag05:
@necolas Thank you for the feedback. Having gone forward and used the JSS branch on a project I've noticed some of the issues you've reported and will attempt to correct them at some point. In the meantime, I have used the main project and found your recommendations to be more than acceptable for my needs.

That all being said, I'm not sure I follow your latest comment regarding "className + normal CSS" for pseudo class support. Perhaps that could find it's way into an example or documentation? I think that may be one of the few features supported by JSS that aren't obvious in your StyleSheet implementation.

necolas:
> I'm not sure I follow your latest comment regarding "className + normal CSS" for pseudo class support

You can do what @ndbroadbent suggested and apply a set of styles to an element in response to an event callback. That is the most idiomatic and reliable approach, plus you can encapsulate the behaviour in a component, e.g., `<TouchableWithHover style={styles.default} hoverStyle={styles.hover} />`.

The `className` prop is also supported (not on all components yet) as an escape hatch – for migrating existing codebases, etc. – so you could apply a static className from any CSS file:

```css
.some-class-with-hover-styles:hover {
  opacity: 0.8;
}
```

```js
<Touchable className="some-class-with-hover-styles" />
```

necolas:
Closing this now. Thanks for investigating

Issue #341, opened 2017-01-31
Regression with UIManager measurement methods


bloodyowl:
**What is the current behavior?**

https://codepen.io/anon/pen/JEpbjg

The new calculation method of UIManager using `offset${ X }` properties doesn't return the expected values, because `offset${ X }` methods stop calculation at the closest positioned ancestor.

It breaks `measureLayout` behaviour by returning the wrong offsets.

**What is the expected behaviour?**

<img width="491" alt="screen shot 2017-01-31 at 15 03 59" src="https://cloud.githubusercontent.com/assets/1688645/22467876/a5d95d1e-e7c6-11e6-970d-b5f15e58aede.png">

necolas:
Thanks, I suppose we could iterate up through the tree and accumulate offsets. I'm a little concerned about how that could impact performance vs a single call to get client rects (which unfortunately includes the effects of transforms).

bloodyowl:
from what I can recall, most DOM libraries à-la-jQuery moved to `getBoundingClientRect` at some point both for performance and browser consistency issues. the only bug I can recall was with inline-block elements within inline ones, which shouldn't be an issue in RNW.

I think a loop iterating over `offsetParent` should do.

would you like a PR for this?

necolas:
> would you like a PR for this?

sure, thanks!

PR #345, opened 2017-02-01
Make getRect iterate over the ancestor chain

bloodyowl:
Fix for #341

Didn't add any test because jsdom doesn't seem to support `offset*` properties at all 😕

### DIFF
--- a/src/apis/UIManager/index.js
+++ b/src/apis/UIManager/index.js
@@ -2,10 +2,15 @@ import asap from 'asap';
 import CSSPropertyOperations from 'react-dom/lib/CSSPropertyOperations';
 
 const getRect = (node) => {
-  const height = node.offsetHeight;
-  const left = node.offsetLeft;
-  const top = node.offsetTop;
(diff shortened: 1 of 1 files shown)


MoOx:
I previously had to mock those things indeed "by hand" https://github.com/MoOx/phenomic/blob/bcf8456b47a636637af8baeabc48ea2ea38ec2c4/src/client/__tests__/should-update-scroll.js not a super experience :)

src/apis/UIManager/index.js
```diff
@@ -2,10 +2,15 @@ import asap from 'asap';
 import CSSPropertyOperations from 'react-dom/lib/CSSPropertyOperations';
 
 const getRect = (node) => {
-  const height = node.offsetHeight;
-  const left = node.offsetLeft;
-  const top = node.offsetTop;
-  const width = node.offsetWidth;
+  let height = node.offsetHeight;
+  let width = node.offsetWidth;
```
MoOx: use const for those ;)

PR #360, opened 2017-02-10
basic end 2 end example of setting up react-native-web

### LINKED ISSUE #359: end2end example
Attempting a basic ReactNative app with webpack config

https://gist.github.com/drewwells/27411bd34c270a37e5274184742c7c2e

The app builds, but throws require is not defined errors: `module.exports = require("url");`

**What is the expected behaviour?**
App loads as expected within react-native environment

**Steps to reproduce**

1. webpack-dev-server webpack.config.js
2. open localhost:8080

**Environment (include versions)**

OS: Sierra
Device: Mac
Browser: Chrome
React Native for Web (version): 0.0.70
React (version): 15.4.2

drewwells:
- ignore upstream babel
- add starter page
- add react and babel-react support

fixes #359 
**Test plan**
webpack-dev-server
**This pull request**

- [x] includes documentation
- [/] includes tests
- [/] includes benchmark reports
- [/] includes an interactive example
- [x] includes screenshots/videos


RangerMauve:
I've found that using create-react-app works out of the box, as well.
I'd actually recommend using that instead of anything custom that needs to be constantly kept up to date manually

drewwells:
Generators are Great for starting new projects. However, those that have react native code already can't use a generator effectively. 

Also this demo being in the demo and breaking is sort of the point. The demo stands as living documentation that is versioned along with the library that it runs on.

I spent several hours piecing together several webpack configs to get here. If there was a full working example with dependency versions and the correct config parameters, it would have saved me a lot of time.

RangerMauve:
You can easily add react-scripts to an existing project, actually. There's a guide on it [here](http://fredrik.anderzon.se/2016/12/04/adding-create-react-app-scripts-to-an-existing-project/)

The changes you'd need to make to an existing react-native project are the addition of a `public` folder with the `index.html` file described in the tutorial, and an `src/index.js` file for the entry point for the web.

necolas:
Hey, I think you had a rough time configuring webpack which you set up to build for node rather than web. I won't be including this code. There's little here that isn't covered in these repos "React Native" docs. But I can add mention of using babel-loader and babel-preset-react-native. Thanks

RangerMauve:
By the way, `create-react-app` isn't actually good enough for this after all. They don't put their dependencies through babel, so if you want to use a library that makes use of JSX or es6, you're out of luck and need to create your own scripts.

PR #377, opened 2017-02-28
Use prettier@0.19

necolas:
Installs prettier@0.19 and reformats the contents of the `src` directory.

cc @paularmstrong @itszero @madnl @sieverk @vjeux


src/apis/AppState/index.js
```diff
@@ -2,7 +2,7 @@ import ExecutionEnvironment from 'fbjs/lib/ExecutionEnvironment';
 import findIndex from 'array-find-index';
 import invariant from 'fbjs/lib/invariant';
 
-const EVENT_TYPES = [ 'change' ];
+const EVENT_TYPES = ['change'];
```
paularmstrong: When I tried prettier on Normalizr, this is the one change that annoyed me to no end. I'm sure I could get used to it,  but the latter is the style that we've been using for so long that it's just too ingrained in habit.

vjeux:
<3 <3 <3

Let me know if there's anything that you find odd looking, I can either fix it or provide context around why it is that way.

src/apis/AppState/index.js
```diff
@@ -32,18 +32,29 @@ class AppState {
 
   static addEventListener(type: string, handler: Function) {
     if (AppState.isSupported) {
-      invariant(EVENT_TYPES.indexOf(type) !== -1, 'Trying to subscribe to unknown event: "%s"', type);
+      invariant(
```
paularmstrong: Also, may want to configure line lengths?

src/apis/AppState/index.js
vjeux: We used to add spaces around [] in the bracket spacing option but a lot of people complained so we removed it. May be worth revisiting. cc @jlongster for a data point.

jlongster: ¯\_(ツ)_/¯ Something like prettier is so hard because everyone has ingrained habits like that, and they all conflict. I can't find the issue where we talked about it, but it was a complaint. I have no idea what to do except... say sorry? :/

I mean, we *could* make the bracketSpacing option more configurable but we really, really don't want to add many more configuration options unless we really need to. 

If you want to continue discussing it, please file an issue! I think, even if we've talked about it before, it's worth talking with newcomers because it helps them understand how we got there and hopefully makes them feel better about it. And maybe you could even change our minds!

paularmstrong: Actually, I've had to go back and forth multiple times depending on the company and project. I'm just tired of switching :)

necolas: Same, I find it harder to read without spaces. I guess people will complain either way but I was expecting spacing to be added for both `{` and `[` for consistency.

src/apis/AppState/index.js
necolas: It's configured to 100 chars

paularmstrong: Ah, was expecting 120. Whatever works!

src/apis/AppState/index.js
paularmstrong: This was honestly the only extra thing that I felt needed an option. It's just jarring to go back and forth on a silly little thing. Sure, probably kind of petty in the long run, but it's almost like the tabs-vs-spaces argument to me. You're just never going to convince firm believers that one is better than the other.

jlongster: Wouldn't it be nice to get used to a single format though instead of jumping back and forth when going between companies and projects, like you said? That'd going to happen if we make it an option.

necolas: Why not remove all the options then :P

So spaces for `[` and `{` are inconsistent because that's what airbnb's style guide requires?

src/apis/AppState/index.js
necolas: Oh yeah we've used 120, I can change that

performance/implementations/glamor/Box/index.js
```diff
@@ -6,12 +6,7 @@ import View from '../View';
 const Box = ({ color, fixed = false, layout = 'column', outer = false, ...other }) => (
   <View
     {...other}
-    style={[
-      styles[`color${color}`],
-      fixed && styles.fixed,
-      layout === 'row' && styles.row,
-      outer && styles.outer
-    ]}
+    style={[styles[`color${color}`], fixed && styles.fixed, layout === 'row' && styles.row, outer && styles.outer]}
```
necolas: @vjeux @jlongster this formatting change is unfortunate for RN styles / components. Thinking out loud, I wonder if proxies for reading complexity of code – e.g., use of conditionals and multiple elements – could eventually be used for context-specific formatting.

vjeux: prettier relies heavily on column count for this, which works well for 80 columns. Since you pushed it to 120, you are going to be able to cram a lot more inside.

One solution that we are considering is to expand the array if it was expanded in the original source. We already do this for objects and it solved a ton of hard problems to algorithmically decide on.

https://github.com/prettier/prettier/issues/817#issuecomment-283075178

necolas: I think we only bumped up to 120 at Twitter because our imports (especially with relative paths) can break over 2 or more lines with the 80 char limit. You might not experience that so much at FB with your module system :). OK so I might drop back to 80 for now since it's so easy to do!

Really impressed with the results so far - great work! This will be a useful case study to include in a proposal to adopt `prettier` at Twitter. Let me know if you have any recommendations / warnings related to doing that.

src/apis/AppState/index.js
jfsiii: One option if you want to deviate from prettier's preferences is to use [`prettier-eslint`](https://github.com/prettier/prettier-eslint)

performance/implementations/glamor/Box/index.js
vjeux: We also ignore imports and require > 80 lines at fb. There's an issue opened for that on prettier in order not to break them into multiple lines.

As for using it at Twitter, some thoughts:
- it is still very early and hasn't been battle tested in prod with fb employees. I'm only adding it to the codebase.
- if you could run the command with --debug-check instead of --write on your codebase, it does prettyprint(prettyprint(file)) === prettyprint(file). I would love if you could see if there's any that trigger on twitter codebase so that we can fix them.
- the printer itself at this point is pretty solid, there's still edge cases reported frequently but they are now pretty small and usually easy to fix.
- the biggest shaky piece is integration. Do you integrate it as a lint rule, as a precommit hook...? How do you keep the editor version with the linter version in sync? many of the integrations are currently calling the node "binary" every time which is slower as you have to pay for node startup time + parsing all the files...
- I'm working on prettier full time until it is widely used at Facebook, so expect a lot of polish in the next month or two :)

src/apis/AppState/index.js
Pomax: A suggestion: if you have an editor that can work with JS styling profiles (Sublime or the like) you could tell it to load in code and automatically apply this one rule that makes your developer life easier, making sure that everything you load in has spaces, with a precommit hook that runs prettier on your src dir first, to ensure that the code you check in is what `prettier` generates. That way you get to enjoy the styling you like while you work on the code personally, but the project as a whole only ever sees prettier-"conformant" code.

performance/implementations/glamor/Box/index.js
necolas: Thanks, I'll get back to you

necolas:
This was merged and the reformatting done in a2f25a46c495ca53a75e728cfb14dbdf67cdb342

PR #378, opened 2017-02-28
[add] appearance 'View' style

bloodyowl:

Adds support for `appearance`, which is useful to reset styles in `button` in order to make it positionable absolutely. 

**Test plan**

**This pull request**

- [ ] includes documentation
- [ ] includes tests

### DIFF
--- a/src/components/View/ViewStylePropTypes.js
+++ b/src/components/View/ViewStylePropTypes.js
@@ -28,6 +28,7 @@ module.exports = {
   /*
    * @platform web
    */
+  appearance: string,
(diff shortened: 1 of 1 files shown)


necolas:
Should appearance just be set to 'none' by default for buttons?

bloodyowl:
Not sure, my first thoughts is that it can only make them simpler to style

necolas:
458c534 should ensure that the `appearance` is automatically reset when the underlying node is a `button`. It also falls back to the RN equivalents of `accessibilityRole` when determining if the element is a `button`.

PR #382, opened 2017-03-03
TouchableWithoutFeedback tabIndex override

holmesal:

If you wrap a `TextInput` in a `TouchableWithoutFeedback`, you cannot tab to the input without tabbing through the Touchable.

This is because `TouchableWithoutFeedback` [explicitly sets a `tabIndex` of `0`](https://github.com/necolas/react-native-web/blob/master/src/components/Touchable/TouchableWithoutFeedback.js#L182) when not disabled.

This patch solves the problem by allowing `props.tabIndex` to override the default value when the `TouchableWithoutFeedback` is not `disabled`

When the `TouchableWithoutFeedback` is `disabled`, the existing behavior is maintained (cannot tab to it).

**This pull request**

- [ ] includes documentation
- [ ] includes tests

### DIFF
--- a/src/components/Touchable/TouchableWithoutFeedback.js
+++ b/src/components/Touchable/TouchableWithoutFeedback.js
@@ -179,7 +179,7 @@ const TouchableWithoutFeedback = React.createClass({
       onResponderTerminate: this.touchableHandleResponderTerminate,
       style,
       children,
-      tabIndex: this.props.disabled ? null : '0'
+      tabIndex: this.props.disabled ? null : (this.props.tabIndex || '0')
     });
   }
 });


necolas:
Please could you explain the use case for wrapping an Input in a Touchable?

holmesal:
Sure - say you're building an input like this:

image

The top label, right icon, and possibly even bottom underline will be separate components. If you want those to focus the input when clicked, you need to wrap the whole thing in a Touchable* and focus the input onPress.

mattdlh:
This would help us a great deal! @necolas can this fix be merged?

necolas:
What happens on native when you use a screen reader like TalkBack or VoiceOver?

`tabIndex` is a web-specific property so I'd rather not expose that in the API if possible, as they problem you describe seems unlikely to be limited to web.

holmesal:
RE: screen readers on native

On iOS, you'd implement the `UIAccessibility` protocol and mark yourself as "not accessible.
On Android, you'd set the `android:isImportantForAccessibility` attribute to "no".

RE: exposing platform-specific property

`react-native` does a pretty good job of using cross-platform properties where possible, and only using platform specific properties when the feature is only supported on one platform.

You also do this on `react-native-web` where it makes sense - for example, `View` has a `cursor` style properly, which only makes sense on `react-native-web`.

IMO `tabIndex` is a property like this. It's only useful in `react-native-web`, where you have a physical `tab` key.

necolas:
> On Android, you'd set the android:isImportantForAccessibility attribute to "no"

So you're saying that to avoid tabbing into the touchable in RN (for Android) you set `android:isImportantForAccessibility='no'`? From this it sounds like `accessible={false}` does this too https://facebook.github.io/react-native/docs/accessibility.html#accessible-ios-android

I think `tabindex` can be abstracted away and I have a patch that does that. There are only 2 useful values: `0` and `-1`. Just need to be sure which props to expose to users.

holmesal:
Awesome! Abstracting away `tabIndex` works for our use-case, as long as it's something that `accessible={false}` sets under the hood.

necolas:
Hey @vjeux, did you have any thoughts about how RN's `accessible` (and perhaps `isImportantForAccessibility`) would be implemented on web?

I read through the RN docs on `accessible`; is it fair to say this roughly maps to whether an element is "focusable" and can be implemented with `tabindex`? In the example given here (nested touchables), using `accessible={false}` would removed the touchable from the tab flow but not hide content from screenreaders.

To hide content from screenreaders (using `aria-hidden`), I was looking at adding web support for `isImportantForAccessibility`. However, there is no web equivalent that I'm aware of for `isImportantForAccessibility="no"` (which @holmesal mentioned as one way to avoid a wrapping touchable from being focusable by a SR) - what if "no" was mapped to `accessible={false}`?

vjeux:
If I remember correctly (it's been a while) `accessible` on RN iOS actually means, "it is visible to screen readers". If you don't have it, then screen readers won't be able to see it.

A lot of the native UI components already implement accessibility so they may not respect the `accessible` flag (not sure).

Sorry I can't be more helpful :(

necolas:
No worries, is there anyone on the React Native team that you think could help me figure this out?

necolas:
Also @mattdlh and @holmesal, please can you share the React Native code patterns you're using to achieve the desired result on iOS and Android?

necolas:
Please give 0.0.83 a try. I've updated the component and accessibility docs with details. Let me know

PR #398, opened 2017-03-17
Add default option for Platform.select

brunolemos:
See: https://github.com/facebook/react-native/commit/f30ab35e9278f120247f49d6a355e263cc357946


necolas:
Applied in e6f00f7592c7e29610bda34aff930cc49d186419. Thanks

PR #400, opened 2017-03-21
implement callback behavior for AsyncStorage.getAllKeys()

### LINKED ISSUE #399: AsyncStorage.getAllKeys() does not behave like React Native's API
The documentation for React Native's AsyncStorage API's `getAllKeys()` says that the function should take a callback as the first argument which is a "Function that will be called the keys found and any error."

However, the react-native-web version does not accept any parameters.

While it is possible to use the API with the promise object returned by the function, it probably should also handle the callback, as documented.

I'll open a PR for this in the morning with a fix, unless there is some reason for the omission.

aaroncraig10e:
Resolves #399

### DIFF
--- a/src/apis/AsyncStorage/__tests__/index-test.js
+++ b/src/apis/AsyncStorage/__tests__/index-test.js
@@ -58,7 +58,7 @@ describe('apis/AsyncStorage', () => {
       expect(keys).toEqual(['UID123']);
     });
 
-    test.only('should have same callback behavior as react-native', async () => {
+    test('should have same callback behavior as react-native', async () => {
       let keys = null;
       let err  = true;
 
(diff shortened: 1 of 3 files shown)


necolas:
All the functions support the callback style usage in RN: https://facebook.github.io/react-native/docs/asyncstorage.html

src/apis/AsyncStorage/__tests__/index-test.js
```diff
@@ -21,25 +21,92 @@ const waterfall = (fns, cb) => {
...
+    beforeEach(async () => {
+      window.localStorage = mockLocalStorage;
+      await AsyncStorage.setItem('UID123', JSON.stringify(UID123_object));
+    });
+
+    afterEach(() => {
+      window.localStorage = originalLocalStorage;
+    });
+
+    test('should have same Promise behavior as react-native', async () => {
```
necolas: Please can you use more descriptive test names than "same as react-native"

src/apis/AsyncStorage/index.js
```diff
@@ -40,8 +40,17 @@ class AsyncStorage {
           const key = window.localStorage.key(i);
           keys.push(key);
         }
+
+        if (cb) {
+          cb(null, keys);
+        }
+
         resolve(keys);
```
necolas: Is the expected behaviour that the callback is called AND the promise resolved?

aaroncraig10e: That's my understanding from the documentation (and using AsyncStorage in native).

nthtran:
Would be great to support callbacks. Libs like `redux-persist` rely on this behaviour.

aaroncraig10e:
@nthtran That's how I discovered this issue, actually -- we're using `redux-persist` and in fact, it breaks without this patch.

aaroncraig10e:
@necolas any chance of getting this merged in?

necolas:
I'll accept a PR that adds the callback interface to every function that supports it in the RN AsyncStorage API

aaroncraig10e:
Fair enough -- I'll get that together over the next week or so.

Issue #91, opened 2016-03-10
Components backlog


necolas:
- [ ] ListView
- [ ] Picker
- [ ] ProgressBar
- [ ] Switch
- [ ] ViewPager
- [ ] (Navigator once new API is stable)

necolas:
From #195 by @MoOx

https://gist.github.com/MoOx/96169e8af11894d3e74a2aa29b2ccac5

RangerMauve:
Are you accepting PRs for any of these?

necolas:
Yep

smkhalsa:
When can we expect FlatList and SectionList? Are there any blockers to a web implementation at the moment?

RangerMauve:
@smkhalsa Check out #464

joncursi:
@necolas I started on an initial version of `FlatList` since I needed it for my own project. It's rather simple as I'm not using `VirtualizedList` or anything fancy under the hood. It also doesn't support all of the props, just the main ones for now as that's all I needed. Is something like this worth commiting to this repository?

```jsx
/* @flow */

import * as React from 'react';

import type { StyleFlowType } from '../../../types/style';
import { ScrollView, View } from 'react-native-web';

type PropsFlowType = {
  ItemSeparatorComponent?: React.ComponentType<any>,
  ListFooterComponent?: React.ComponentType<any>,
  ListHeaderComponent?: React.ComponentType<any>,
  contentContainerStyle?: StyleFlowType,
  data?: Array<any>,
  horizontal?: boolean,
  keyExtractor?: (item: any) => string,
  renderItem: (obj: { index: number, item: any }) => ?React.Element<any>,
  style?: StyleFlowType,
};

const FlatList = ({
  ItemSeparatorComponent,
  ListFooterComponent,
  ListHeaderComponent,
  contentContainerStyle = {},
  data = [],
  horizontal = false,
  keyExtractor,
  renderItem,
  style = {},
}: PropsFlowType = {}): React.Node => (
  <ScrollView
    contentContainerStyle={contentContainerStyle}
    horizontal={horizontal}
    style={style}
  >

    {ListHeaderComponent &&
      <ListHeaderComponent />
    }

    {data.map((item, index): React.Node => (
      <View
        key={keyExtractor ? keyExtractor(item) : index}
      >

        {renderItem({ index, item })}

        {(ItemSeparatorComponent && index < data.length - 1) &&
          <ItemSeparatorComponent />
        }

      </View>
    ))}

[...]

PR #411, opened 2017-04-04
Add Modal component

jhen0409:

#91 Modal component

* The `shortAnimTime` is referenced from [RN for Android](https://github.com/facebook/react-native/tree/cfae3e376d96653e5ad48e0d0398674a954d3637/ReactAndroid/src/main/res/views/modal/anim) anim duration, that used for modal. ([`config_shortAnimTime`](https://github.com/android/platform_frameworks_base/blob/15e69df9218575a3695c97a856322de3bf54e8da/core/res/res/values/config.xml#L119): 200)
* Referenced [official documentation](https://facebook.github.io/react-native/docs/modal.html) (havn't add deprecated `animated` prop and platform only props)

**Test plan**

* Simple [snapshot tests](https://github.com/necolas/react-native-web/blob/9d6fec8f7e7cbcd99d3ffcc831bfa2c4d06e7c2d/src/components/Modal/__tests__/index-test.js) for `visible` prop. [index-test.js.snap](https://github.com/necolas/react-native-web/blob/9d6fec8f7e7cbcd99d3ffcc831bfa2c4d06e7c2d/src/components/Modal/__tests__/__snapshots__/index-test.js.snap)
* Added examples to storybook, [deployed on my fork](http://jhen0409.github.io/react-native-web/storybook/?selectedKind=component%3A%20Modal&selectedStory=animationType%3A%20none&full=0&down=1&left=1&panelRight=0&downPanel=kadirahq%2Fstorybook-addon-actions%2Factions-panel)

**This pull request**

- [x] includes documentation
- [x] includes tests

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,6 +56,7 @@ Exported modules:
   * [`Button`](docs/components/Button.md)
   * [`Image`](docs/components/Image.md)
   * [`ListView`](docs/components/ListView.md)
+  * [`Modal`](docs/components/Modal.md)
   * [`ProgressBar`](docs/components/ProgressBar.md)
   * [`ScrollView`](docs/components/ScrollView.md)
   * [`Switch`](docs/components/Switch.md)
--- a/docs/components/Modal.md
+++ b/docs/components/Modal.md
@@ -1,0 +1,61 @@
+# Modal
+
+The Modal component is a simple way to present content above an enclosing view.
+
+## Props
+
+**animationType**: oneOf('none', 'slide', 'fade') = 'none'
+
+The `animationType` prop controls how the modal animates.
+
+* `slide` slides in from the bottom
+* `fade` fades into view
+* `none` appears without an animation
+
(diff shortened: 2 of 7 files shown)


ndbroadbent:
Oh awesome, I was just looking for this so that I could get [react-native-material-dialog](https://github.com/hectahertz/react-native-material-dialog) working on web.

I would love to help test this, but it looks like there's a conflict that needs to be fixed.

necolas:
I think the animations could be implemented as CSS transitions (for performance).

While this is a good start, I think the `Modal` implementation will be more complicated than this as it needs to be rendered outside the tree of your main app (i.e., probably requires working into the AppRegistry root view) in order to get the stacking correct and to support launching modals from within any layer context (e.g., a ScrollView or anything else using `translate3d`)

jhen0409:
Move to #471.

Sorry I haven't so much time for look this. :|

PR #418, opened 2017-04-09
comparison table formatting

threepointone:
just a typo fix.

qq - aphrodite is interesting, because it batches and inserts its styles asynchronously (so this table is probably underreporting). what would that mean for these numbers?

(on a broader point, how will these numbers look once fiber comes into play? would love to hear your thoughts.)

### DIFF
--- a/performance/README.md
+++ b/performance/README.md
@@ -29,9 +29,10 @@ Typical render timings*: mean ± two standard deviations
 Other libraries
 
 | Implementation                       | Deep tree (ms)    | Wide tree (ms)    |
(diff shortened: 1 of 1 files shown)


necolas:
I'm not sure about fiber. Still trying to determine what RNW changes are needed to deal with the bundle changes they are planning.

I think Aphrodite numbers are probably ok as the styles will have been injected by then. But happy to make changes if needed.

These crude benchmarks were mainly to give me some visibility of RNW performance. I want to get as close to css-modules as possible, as that's what we use in Twitter Lite (via something called react-native-shim) and what I'm aiming to replace under experimentation - hopefully comparable performance but superior feature set :)

necolas:
I fixed this while I was make some changes

PR #421, opened 2017-04-11
Added animated example

kwelch:
Added example for API that doesn't have one

**Test plan**
No tests required

**This pull request**

- [ ] includes documentation
- [ ] includes tests


necolas:
Thanks. There is already an Animated example under "View". I'm currently rewriting the examples and will better surface Animated ones soon

kwelch:
Sounds good. This helped me get past an issue I was happening and thought I would see if this added value. 👍

PR #428, opened 2017-04-14
updated the script hook for publishing to match it's intent when it was originally added

JoeCortopassi:
The `prepublish` npm hook has changed in functionality a few times since it was utilized by this script. This has caused branching from `master` to have an extra/un-needed step of building every time before working on it. This update is to remove that step by trying to restore to what appears to be the original intention of adding `prepublish`

**Test plan**
locally link branch, and you won't have to run `npm run build` in the `node_modules/react-native-web` directory every time you use it in your project


JoeCortopassi:
accidentally committed the wrong file. Closed pr and created new one

PR #431, opened 2017-04-18
[change] render Image 'source' immediately if previously loaded

necolas:
Maintain a record of loaded images. If an image has already been loaded,
bypass the JS loading logic and render it immediately. This prevents
flashes of placeholder state when moving between screens or items in a
virtualized list.

### DIFF
--- a/src/components/Image/ImageUriCache.js
+++ b/src/components/Image/ImageUriCache.js
@@ -1,0 +1,60 @@
+class ImageUriCache {
+  static _maximumItems: number = 256;
+  static _entries = {};
+
+  static has(uri: string) {
+    const entries = ImageUriCache._entries;
+    const isDataUri = /^data:/.test(uri);
+    return isDataUri || Boolean(entries[uri]);
+  }
+
+  static add(uri: string) {
+    const entries = ImageUriCache._entries;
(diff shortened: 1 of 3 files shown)


src/components/Image/ImageUriCache.js
```diff
@@ -0,0 +1,58 @@
...
+  static has(uri: string) {
+    const entries = ImageUriCache._entries;
+    const isDataUri = /^data:/.test(uri);
+    return isDataUri || Boolean(entries[uri]);
+  }
+
+  static add(uri: string) {
+    const entries = ImageUriCache._entries;
+    if (entries[uri]) {
+      entries[uri].refCount += 1;
```
paularmstrong: Why not also update a `lastAccessed` timestamp? Small concern that as implemented, this isn't *least recently used*, but *first added*. Just because something doesn't exist on screen doesn't mean that it isn't a high priority and often used item (I'm thinking logged-in user avatar, for example)

necolas: Good idea

necolas:
1f80e4c10595ce3c6fb357ce298f8f6fb65d338e

PR #438, opened 2017-04-22
Allow filter property in CSS

ndbroadbent:

It allows the `filter` CSS rule.


necolas:
Update the View docs too please

ndbroadbent:
My apologies, I didn't see those. Updated

PR #445, opened 2017-04-25
Added an example app with an awesome webpack configuration

ndbroadbent:

It took me a while to figure this stuff out, so I think it would be very helpful for others.

**Test plan**

```bash
cd example_app
yarn
npm run web
```

**This pull request**

- includes documentation
  * [Project README](https://github.com/ndbroadbent/react-native-web/blob/example-app/README.md)
  * [Getting Started](https://github.com/ndbroadbent/react-native-web/blob/example-app/docs/guides/getting-started.md)
  * [Example App README](https://github.com/ndbroadbent/react-native-web/blob/example-app/example_app/README.md)
- [includes an interactive example](https://github.com/ndbroadbent/react-native-web/tree/example-app/example_app)

P.S. I really want to get [react-hot-loader](https://github.com/gaearon/react-hot-loader) working, but I was struggling with that. [I think I need to use `ReactNative.render`](https://github.com/gaearon/react-hot-loader/tree/master/docs#webpack-2) instead of `AppRegistry.runApplication`. But when I do that, the layout gets kind of messed up, and I'm not sure why that's happening. UPDATE: I can [reproduce it in this example app, so I opened an issue](https://github.com/necolas/react-native-web/issues/446).

Also, how would you recommend providing the Promise and Array.from polyfills? It was something that I overlooked (and still haven't set up), so I think it would be great if `react-native-web` just included those by default.


necolas:
You'll need to include the babel-runtime for those polyfills, and probably hook up the promise polyfill yourself.

This patch is huge and I'm not going to maintain such a complex app setup here. There are already many starter kits for react and react native, and webpack boilerplates. If there are some smaller changes I could make, like a note about polyfills or linking to external examples, I'd be happy to do that instead.

ndbroadbent:
Cool, no problem! I moved it over here: https://github.com/ndbroadbent/react-native-web-webpack

ndbroadbent:
Hi again, it's been a [real nightmare](https://github.com/webpack/webpack/issues/3974#issuecomment-297096211) trying to get `babel-polyfill` or `babel-runtime` working, so I ended up just importing those specific polyfills from `core-js`:

```
  entry: {
    app: [
      'core-js/fn/promise',
      'core-js/fn/array/from',
      path.join(__dirname, '../index.web.js'),
    ],
  },
```

When you say:

> It is recommended that your application provide a Promise and Array.from polyfill.

Do you mean those are the exact polyfills that are required for all of the supported browsers? 

Sorry to bother you with this stuff! I think it would be great if there were some more notes about polyfills, and how you would recommend setting those up.

necolas:
Using babel-runtime should automatically use helper functions to deal with `Promise` etc in your app code. Most JS libraries already expect you to globally polyfill `Promise` in your app.

```
if (typeof global.Promise !== 'function') {
  global.Promise = require('babel-runtime/core-js/promise').default;
}
```

I believe a future version of React will also expect you to polyfill `requestAnimationFrame`

PR #464, opened 2017-05-03
Flat List and Virtualized List

tazsingh:
Currently there isn't a `FlatList` or `VirtualizedList` implementation in React Native Web, as was introduced officially on React Native 0.43. For compatibility, I've added initial implementations of these here from a virtualized list component that I've created our web app which I've adapted to the `VirtualizedList` API.

Note that my implementation takes a different approach than what I've seen elsewhere. I've used an intermediate `ListPage` component to house each `ListCell`. This has the benefit of being smarter about determining which page to render, not strictly requiring knowledge of each cell's height in advance, and can maintain scroll positions with ease.

There are a few caveats with these implementations that I'm looking to refine over time. I'd like to ship these with a Release Candidate tag in the docs and resolve any issues as they come up.

**Test plan**
This PR does include tests for the features that are supported. Please read through the tests as some features were not testable as Jest does not utilize a real DOM with real scroll events and style calculations. I've tested this manually for the features included in the test suite and it is all working as expected.

We are continuing to implement RN's `FlatList` and `VirtualizedList` on our code base in conjunction with these components. As we notice incompatibilities or features that need to be added, we will add it here.

**This pull request**

- [x] includes documentation
- [x] includes tests


package.json
```diff
@@ -32,6 +32,7 @@
     "fbjs": "^0.8.8",
     "hyphenate-style-name": "^1.0.2",
     "inline-style-prefixer": "^3.0.2",
+    "lodash": "^4.17.4",
```
necolas: Please remove the lodash dependency. It adds too much to the bundle size even when using specific modules

package.json
```diff
@@ -83,6 +86,7 @@
     "web"
   ],
   "jest": {
+    "setupTestFrameworkScriptFile": "./node_modules/jest-enzyme/lib/index.js",
```
necolas: What is this doing?

src/components/FlatList/FlatListPropTypes.js
```diff
@@ -0,0 +1,13 @@
...
+
+export default {
+  ItemSeparatorComponent: PropTypes.oneOfType([
+    PropTypes.func,
+    PropTypes.element
+  ]),
+  ListFooterComponent: PropTypes.oneOfType([PropTypes.func, PropTypes.element]),
+  ListHeaderComponent: PropTypes.oneOfType([PropTypes.func, PropTypes.element]),
+  columnWrapperStyle: PropTypes.any,
+  extraData: PropTypes.any,
```
necolas: Please alpha-sort props

src/components/FlatList/__tests__/index-test.js
```diff
@@ -0,0 +1,167 @@
...
+    test('renders 1 header', () => {
+      const wrapper = render(
+        <FlatList {...defaultProps} ListHeaderComponent={HeaderComponent} />
+      );
+
+      expect(wrapper.find('.header')).toHaveLength(1);
+    });
+  });
+  // describe("columnWrapperStyle"); // not supported
+  // describe("numColumns"); // not supported
```
necolas: Is this a limitation of the implementation design?

src/components/FlatList/__tests__/index-test.js
```diff
@@ -0,0 +1,167 @@
...
+      const mockFn = jest.fn(list.listRef.scrollToEnd);
+
+      list.listRef.scrollToEnd = mockFn;
+
+      list.scrollToEnd();
+
+      expect(mockFn).toHaveBeenCalled();
+    });
+  });
+  describe('scrollToIndex', () => {
```
necolas: nit: blank lines between `describe` blocks

src/components/View/index.js
```diff
@@ -35,10 +37,15 @@ class View extends Component {
 
   getChildContext() {
     const isInAButtonView =
-      AccessibilityUtil.propsToAriaRole(this.props) === 'button' || this.context.isInAButtonView;
+      AccessibilityUtil.propsToAriaRole(this.props) === 'button' ||
+      this.context.isInAButtonView;
     return isInAButtonView ? { isInAButtonView } : emptyObject;
   }
 
+  bindRef = element => {
```
necolas: Revert these changes and get the ref in the relevant list component

src/core.js
```diff
@@ -13,7 +13,8 @@ import TextInput from './components/TextInput';
 import Touchable from './components/Touchable/Touchable';
 import TouchableHighlight from './components/Touchable/TouchableHighlight';
 import TouchableOpacity from './components/Touchable/TouchableOpacity';
-import TouchableWithoutFeedback from './components/Touchable/TouchableWithoutFeedback';
+import TouchableWithoutFeedback
```
necolas: revert

webpack.config.js
```diff
@@ -1,19 +1,20 @@
-const path = require('path')
-const webpack = require('webpack')
-const BundleAnalyzerPlugin = require('webpack-bundle-analyzer').BundleAnalyzerPlugin;
+const path = require("path");
```
necolas: revert all these changes please

necolas:
Thanks for giving this a shot! Please could you break this into 2 reviews, the first for the virtual list which I hope @madnl and @itszero can take a look at too. It would also be good to document the differences in the implementation that you describe in the code itself, and what you think is wrong with the React Native implementation.

tazsingh:
@necolas Thanks for the feedback!

Quick update on my end - I've submitted this from code extracted from our codebase and adapted to these APIs. At work we've decided to go a different route as the RN FlatList/VirtualizedList just aren't working out for us for a critical piece of functionality.

Once I've finished working on that, I'll loop back around to this and address these comments as we'll still need FlatList and VirtualizedList in RNW to use elsewhere in our app. Will likely be some point next week.

src/components/VirtualizedList/index.js
```diff
@@ -0,0 +1,336 @@
...
+        return memo.concat(page.ref.calculateItemsOnScreen(listDimensions));
+      }
+
+      return memo;
+    }, []);
+  }
+
+  assignItemPageRef(index) {
+    return element => {
+      this.state.pages[index].ref = element;
```
MarkMurphy: Not sure why state is being mutated here? Why not use `setState`

brunolemos: Probably because it is not supposed to trigger a rerender

brunolemos:
Apparently it's not a drop in replacement yet

image

tazsingh:
Thank you everyone for your feedback!

Have a look over at #474 for the description in the PR there. Let me know if it makes sense to continue adapting this implementation or if a straight port for compatibility makes more sense.

IMO, it's better to continue with this implementation as the performance is way better in my tests. But I want to hear from you and your experience with the two. Let me know if you have any difficulty running this code.

webpack.config.js
tazsingh: This was made by `prettier`. I noticed that some parts of the code are using `prettier` but others aren't. What's the distinction that I should make with this? Does it make sense to do a commit to `master` that `prettier`ify's everything?

src/components/FlatList/__tests__/index-test.js
tazsingh: Not a limitation at all. Just not supported right now. Can work on it in a future iteration.

package.json
tazsingh: I needed it for some reason but can't remember right now. Will try removing and see what it breaks depending on feedback on this vs #474

brunolemos:
@tazsingh IMO api and final result must be exactly the same (no brainer drop in replacement, like the rest of this project), the underlying implementation not necessarily. Don't know necolas thoughts.

necolas:
Going to close this as we have #474 to introduce `VirtualizedList` and can follow up with another PR to introduce `LazyList` etc. Thanks!

PR #466, opened 2017-05-04
chore(flow): Set up config, added third party libdefs

peggyrayzis:
Here's the first Flow PR, more to follow!

- Set up our `.flowconfig`
- Added flow CLI to dev dependencies, added flow script
- Installed third-party libdefs via `flow-typed`
- Upgraded `fbjs` to fix a flow bug

Moves us closer toward resolving all Flow errors. Now that our config is correct, we only have 41 errors as opposed to the 70+ we had before.

**Test plan**
`npm run flow`

**This pull request**
- [ ] includes documentation
- [ ] includes tests


.gitignore
```diff
@@ -1,3 +1,4 @@
 node_modules
 dist
 dist-examples
+*.log
```
necolas: You should ignore this in your global `.gitignore`

flow-typed/npm/@kadira/storybook_vx.x.x.js
```diff
@@ -0,0 +1,263 @@
...
+ *   '@kadira/storybook'
+ *
+ * Fill this stub out by replacing all the `any` types.
+ *
+ * Once filled out, we encourage you to share your work with the
+ * community by sending a pull request to:
+ * https://github.com/flowtype/flow-typed
+ */
+
+declare module '@kadira/storybook' {
```
necolas: Why does storybook need to be typed?

peggyrayzis:
Not that I know of. When you run `flow-typed install`, it reads from your pkg.json and adds libdefs for all dependencies. If flow-typed doesn't have the libdef in its repository, it auto-generates a stub, which is what you're seeing with `storybook`. I think there's discussion to ignore dev dependencies in the future, but this feature doesn't exist yet.

Cherry picking the libdefs requires a lot of manual work. Not only do you have to cherry pick with a fresh install, you also have to cherry pick each time you update the libdefs to their latest version.

Check out [the docs](https://github.com/flowtype/flow-typed) for more info.

necolas:
OK. Why does `flow-typed` need to be used in RNW rather than in projects that depend on it?

peggyrayzis:
RNW needs to use `flow-typed` because it relies upon third-party dependencies, such as React's `prop-types` & `create-react-class` modules.

When you run flow without `flow-typed` definitions, you get 38 errors. With the `flow-typed` libdefs, you get 41 errors. Having third-party libdefs ensures that the project is more type safe because it guards against using third-party modules incorrectly.

Here's an example of some of the errors `flow-typed` catches:
```
src/components/Touchable/TouchableWithoutFeedback.js:38
 38: const TouchableWithoutFeedback = createReactClass({
                                      ^ function call. Expected a React PropType instead of
 84:     pressRetentionOffset: EdgeInsetsPropType,
                               ^^^^^^^^^^^^^^^^^^ object type

src/components/Touchable/TouchableWithoutFeedback.js:38
 38: const TouchableWithoutFeedback = createReactClass({
                                      ^ function call. Expected a React PropType instead of
 93:     hitSlop: EdgeInsetsPropType
                  ^^^^^^^^^^^^^^^^^^ object type
```

necolas:
Do you know why Facebook doesn't export types for their own packages (e.g., fbjs, prop-types, create-react-class, etc) if this is important for the React ecosystem? It would be good to get some input from the Facebook Flow team, because I'm not excited about how adding Flow support to libraries needs this much work (and commits 1000s of lines of new code)

peggyrayzis:
Idk, I tweeted about it so we'll see if anyone can give us some answers. I also think it's a little confusing that FB isn't using the tools that the Flow docs recommend.

flow-typed/npm/@kadira/storybook_vx.x.x.js
kkemple: `flow-typed` goes through all dependencies and pulls in their libdef or creates stubs so that you don't get 100s of type errors from not having type defs for whatever you include. While not a big deal for those who just run flow from the command line since its not in your face, those users with IDE integrations will have all of these errors constantly nagging them in their editors. so storybook needs to be typed because when you use RNW components in SB flow wants to check types on SB to ensure those APIs are being used properly. NOTE: this isn't for users of RNW but for people working on RNW.

ryyppy:
Hey!
I am not a facebook employee or flow team member per se, but I have been contributing to the flow-typed ecosystem ever since, so I want to give some input:

> Do you know why Facebook doesn't export types for their own packages (e.g., fbjs, prop-types, create-react-class, etc) 

It's because those packages predate flow by a long time... typing deprecated packages and concepts doesn't bring any business value. E.g. why would you type prop-types, when flow is designed to replace them completely via a babel-plugin? 

They introduce a lot of flow types in newer projects though (like yarn or prepack). I would not generally say they don't use their own tools. Time is a huge constraint here.

>  I'm not excited about how adding Flow support to libraries needs this much work

The biggest problem of flow is that library definitions cannot be dependent on each other without manual administration (plugins etc.) and `.flowconfig` files do not compose... so usually, if you have a complex library with many dependencies, a project depending on your typed library needs to make sure to add the same library definitions as well, otherwise things will get out of hand...

When we tried to ship flow typed source code with `styled-components`, we realized it's almost impossible to communicate the changes the users have to do in their codebase to make things type . So we did a fallback on flow-typed library definitions (which still need to be maintained manually).

> ...(and commits 1000s of lines of new code)

I would not care too much about that... it has a lot of benefits to track the library definitions in your repository, since you are basically version locking your definitions (they might change upstream, because some maintainer fixes a bug,... bug fixes are only communicated in commit hashes, not in version numbers).
Also you can easily adapt your definitions, if your codebase is a special snow flake with special needs (like using Bluebird as Promise impl).

peggyrayzis:
@ryyppy thanks so much for your input! Your response was really helpful and even taught me a few things. 😀 

@necolas do you still have any reservations about integrating flow-typed?

necolas:
> why would you type prop-types, when flow is designed to replace them completely via a babel-plugin?

They provide runtime checks for everyone who isn't using flow, which means they're still commonly used in type-checked modules. But adding those definitions manually seems [simple enough](https://github.com/facebook/react-native/tree/master/flow).

> do you still have any reservations about integrating flow-typed?So we did a fallback on flow-typed library definitions

There's an enormous amount of data that is for dev dependencies like cli tools and webpack (both here and in the styled-components repo). Surely that isn't necessary? And almost all the definitions are of type `any`, so it seems that you're keeping flow quiet but it's not adding useful type checking.

> if your codebase is a special snow flake with special needs

I'd appreciate that you not using that kind of language (and its current associations with the far-right) here.

> do you still have any reservations about integrating flow-typed?

I don't mind automating the addition of type definitions but I don't want to commit 12K LOC without knowing that it's all relevant and necessary.

ryyppy:
> There's an enormous amount of data that is for dev dependencies like cli tools and webpack (both here and in the styled-components repo). Surely that isn't necessary? And almost all the definitions are of type any, so it seems that you're keeping flow quiet but it's not adding useful type checking.

Yeah, `flow-typed` might need to change their mind of automatically installing `devDependency` libdefs by default... there was also https://github.com/flowtype/flow-typed/issues/453 regarding this .. maybe there should really be an option to opt-in installing devDependencies...

> I'd appreciate that you not using that kind of language (and its current associations with the far-right) here.

Understood. I am not American, so I didn't think about that connotation tbh 😅

peggyrayzis:
@necolas Sure, what libdefs would you like me to manually add? Include dependencies but exclude dev dependencies? Exclude anything that is stubbed?

peggyrayzis:
To get the ball rolling, I added only what is currently in React Native. 👍

MoOx:
My experience: I just cherry pick defs from flow-typed repo by hand when flow is saying "required module not found" (-> error throw from acutal covered code - by @ flow directive). Boring but avoid noise. If no def is existing, i write a dumb "any" def or a more complicated one if relevant.

necolas:
Thanks, applied in be3c78f31794c3292f706718aaa893e09a1c40d6

PR #471, opened 2017-05-08
Modal Component

jefferscn:
#91  Modal Component 
* use react-transition-group to implement animation.
* for some problem when modal renders in a fixed element in browser(specially Safari),We render the modal root element out of the main app tree.Create a new div under body,then render all modal conent into this div.
* document is copied from #411 

**Test plan**
* default value test
* root element test

**This pull request**

- [x] includes documentation
- [ ] includes tests


vaukalak:
Does anyone know what is the status of this PR, and if the help is needed on that?

piranna:
Can somebody fix the conflicts an merge it upstream?

piranna:
By clossing this, does it means it's upstream already and published on npm?

PR #486, opened 2017-05-21
[Linking] Add implementations for listener + getInitialURL

Minishlink:
`react-native-web` doesn't currently provide any workable implementations for `Linking.addEventListener`, `Linking.removeEventListener` and `Linking.getInitialURL`.

*addEventListener/removeEventListener*
In `react-native`, the URL listener is called each time another app deep links to your app. This can also be used by developers to pass some data to the app using fragments. In `react-native-web`, I propose that the listener is called when a foreign `window` (// app) posts a message (// deep-links) to your `window` (// app). While it surely isn't the same concept as for apps, this implementation at least provides a workable way for developers to develop a cross-platform app.

*getInitialURL*
Instead of returning an empty string, this implementation returns the current URL.

**Test plan**
Added tests and ran `yarn test`.

**This pull request**

- [x] includes documentation
- [x] includes tests

Thanks for your time :)

EDIT: While adding the examples, I had to pay more attention to the `openURL` current implementation. It forbids users to open a new page that contains a `window.opener` object. This is quite critical for things like authentication from websites (for example Trello authentication). I understand the security concerns, but I'm not sure whether react-native-web should do this by default?


necolas:
I think the listeners should probably remain as no-ops

necolas:
`getInitialURL` returning the current URL also sounds semantically incorrect

Minishlink:
Thanks for you review.

> I think the listeners should probably remain as no-ops
Doesn't it have more benefits to have a workable implementation (yet not perfect I agree) rather than nothing? I took example from the `AppState` listeners' implementation.

> getInitialURL returning the current URL also sounds semantically incorrect
Again not a perfect solution but it fits most use cases as `react-native-web` apps are mostly single page applications. We could then iterate and find a way to store the initial URL when the app initializes, don't you think?

necolas:
> Doesn't it have more benefits to have a workable implementation (yet not perfect I agree) rather than nothing?

Nothing is sometimes better than something. In this case, there is no comparable functionality and the stub is there to avoid errors being thrown in components/apps that didn't really consider non-iOS/Android platforms.

> We could then iterate and find a way to store the initial URL when the app initializes, don't you think?

You could assume that `window.location.href` will be correct when the code is first run.

PR #487, opened 2017-05-22
Fix grey circle animation

mamaz:
Current implementation: the grey circle / shadow box does not automatically disappear during movement from `active` to inactive, vice versa. 

This pull request fix this issue by using `mouseUp`, `mouseDown`, `touchStart`, and `touchEnd` event instead of `focus`/ `blur`.

**Updates**:
Now it only uses `mouseUp` to fire the grey circle, and `onChange` to dismiss it.

**Test plan**
run `yarn examples` find the Switch on `component: Switch`, tap and hold the switch.

**This pull request**

- [ ] includes documentation
- [ ] includes tests
- [x] includes screenshots/videos

Examples:

switch

### DIFF
--- a/src/components/Switch/__tests__/index-test.js
+++ b/src/components/Switch/__tests__/index-test.js
@@ -45,7 +45,7 @@ describe('components/Switch', () => {
 
   describe('grey circle', () => {
     test('it should show grey circle when user touch / clicks', () => {
-      const component = shallow
+
     });
 
     test('it should dismiss grey circle immediately', () => {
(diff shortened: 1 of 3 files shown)


necolas:
That's because the control is still focused. Why is this a problem for you?

mamaz:
@necolas 
because the grey circle does not automatically disappear, it needs to be unfocused by tapping somewhere on the browser, 
I hope I can make it like the switch from React Native Material Kit `https://github.com/xinthink/react-native-material-kit`

mamaz:
@necolas 
I have finished this, could you have a look and give feedback for this. Thank you.

necolas:
Thanks. I don't think this approach will work for mouse + touch + keyboard without getting pretty messy. Probably best to rely on the `modality` functionality and convert that into a focus manager of sorts, so components like `Switch` can determine whether a keyboard is in use and requires the focus state. I'll look into that at some point.

PR #474, opened 2017-05-11
Port over VirtualizedList from react-native

tazsingh:
**This patch solves the following problem**
Hey everyone, can you give me some feedback on this?

I created #464 and can move forward with that but would like some feedback here first.

Since FlatList & VirtualizedList are JS implementations, they should be able to be ported over directly from React Native itself. That's what I've done here.

However what I've noticed is that the VirtualizedList performance is horrendously slow with the straight port. This makes me wonder if it's better to continue working on #464 as I know that the performance there is really good and can support all features over time, or to stick with this PR for a true port of the React Native functionality for compatibility but sacrificing performance?

**Test plan**

Test out #464 and this PR to let me know about performance. I've tested it on a list, and this PR is extremely slow to render initially and very janky during scrolling. But your experience may be different and I'd like to hear about it!

**This pull request**

- [ ] includes documentation
- [ ] includes tests
- [ ] includes benchmark reports
- [ ] includes an interactive example
- [ ] includes screenshots/videos

PR #490, opened 2017-05-25
Low level perf optimizations from Chrome flamegraphs

tazsingh:
While exploring performance related to #474, these are some issues that I noticed in Chrome's Performance flamegraphs & JavaScript Profiler:

1. `createDOMProps`:
I noticed that Babel was compiling the function with calls to `arguments` when a default parameter value is used. This dramatically slows down the function execution due to the JavaScript VM not being able to fully optimize the function as it has to check its arguments on every invocation. Changing this to a falsey check avoids this performance pitfall with the same result. This produced a 2x speedup once the function was optimized.
`typeof` calls are also generally slow. I've replaced these with calls to `constructor` which provided an additional 20-30% speedup.

2. `onLayout` callback:
I noticed that the `requestAnimationFrame` was being scheduled many times during the list's scrolling. This was taking a lot of time to schedule work to be done later. Then once the work itself was being processed, it was doing a tiny amount of work, yielding back to the JS VM to schedule the next amount of tiny work, and so on. This would leave a lot of gaps in time which would cause performance to suffer. Moving this over to a single invocation of `requestAnimationFrame` to queue up the work to be batched executed later made this about 2-2.5x faster in my not-overly-scientific testing (i.e. difficult to measure accurately due to the variability of the test scenarios).

**Test plan**
Let me know what you think. I've done some basic testing of `VirtualizedList` and all works fine. I don't believe anything in this PR has a different result than what was there before and therefore shouldn't introduce any issues.

**This pull request**

- [ ] includes documentation
- [ ] includes tests
- [x] includes benchmark reports

### DIFF
--- a/src/apis/UIManager/index.js
+++ b/src/apis/UIManager/index.js
@@ -20,7 +20,7 @@ let measureLayoutQueue = [];
 
 const processLayoutQueue = () => {
   measureLayoutQueue.splice(0, 250).forEach((item) => {
-    const [node, relativeToNativeNode, callback] = item
+    const [node, relativeToNativeNode, callback] = item;
 
     const relativeNode = relativeToNativeNode || (node && node.parentNode);
     if (node && relativeNode) {
@@ -33,7 +33,7 @@ const processLayoutQueue = () => {
   })
 
   if (measureLayoutQueue.length > 0) {
-    setImmediate(processLayoutQueue)
+    setImmediate(processLayoutQueue);
   }
 }
 
--- a/src/apis/UIManager/index.js
+++ b/src/apis/UIManager/index.js
@@ -1,4 +1,5 @@
 import requestAnimationFrame from 'fbjs/lib/requestAnimationFrame';
+import setImmediate from 'fbjs/lib/setImmediate';
 import setValueForStyles from '../../vendor/setValueForStyles';
 
 const getRect = node => {
@@ -14,8 +15,13 @@ const getRect = node => {
   return { height, left, top, width };
 };
 
-const measureLayout = (node, relativeToNativeNode, callback) => {
-  requestAnimationFrame(() => {
+let hasRequestedAnimationFrame = false;
+let measureLayoutQueue = [];
+
+const processLayoutQueue = () => {
+  measureLayoutQueue.splice(0, 250).forEach((item) => {
+    const [node, relativeToNativeNode, callback] = item
+
     const relativeNode = relativeToNativeNode || (node && node.parentNode);
     if (node && relativeNode) {
       const relativeRect = getRect(relativeNode);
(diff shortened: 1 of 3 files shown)


necolas:
Thanks, sounds promising. I'll test this out today. cc @hzoo for the Babel issue; not the first time I've encountered this issue in RNW or Twitter Lite either

hzoo:
Not a Babel issue, the reason is spec behavior. The reasoning is for function arity I believe, https://github.com/babel/babel/issues/612.

Basically for

```js
function a(b, c = 1) {}
```

```js
a.length // 1
```

Since default parameters don't affect the function arity.

So if we remove the argument in the transpiled code, the only way to get the value is by using `arguments`.

I guess we could have a `loose` option for the defaults transform (but since this is non-spec behavior it would need to be an issue/PR if y'all can make that contribution).

hzoo:
In the meantime, I would suggest https://github.com/babel/babel-preset-env and maybe looking into multiple bundles such that browsers that support default parameters (or whatever JS) can native support. Although you might find native ES6+ to be slower as well. But simplest thing to do is to add a loose option assuming/understanding that it means you don't care about function length

tazsingh:
Thanks @hzoo for the explanation of this and the reco! I tried out the `env` preset within the Babel REPL on my browser (Chrome) and noticed that it still compiles the same even though default values are supported by Chrome: [link to REPL](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=env%2Creact%2Cstage-1&targets=&browsers=&builtIns=false&debug=false&experimental=true&loose=false&spec=false&playground=false&code=const%20foo%20%3D%20%22oh%20hey%22%0A%0Afunction%20bar%20(arg1%20%3D%20foo)%20%7B%0A%20%20console.log(arg1)%0A%7D) and [link to compat table](https://babeljs.io/repl/#?babili=false&evaluate=true&lineWrap=false&presets=env%2Creact%2Cstage-1&targets=&browsers=&builtIns=false&debug=false&experimental=true&loose=false&spec=false&playground=false&code=const%20foo%20%3D%20%22oh%20hey%22%0A%0Afunction%20bar%20(arg1%20%3D%20foo)%20%7B%0A%20%20console.log(arg1)%0A%7D)

Overall I definitely think `babel-preset-env` would be good to employ for this project. But not sure if it directly applies here as the project does support browsers where the default parameters aren't supported, unless we do different builds for each browser (up to @necolas but may be difficult to distribute and for user understanding).

hzoo:
screen shot 2017-05-25 at 8 39 29 am

Works fine for me, but yes you would need to figure that out if you want to ship ES6 (and then if you minify figure that out too). Should measure native default parameter perf in browser versions too though. I was thinking about a "perf-table" instead of compat-table but it would be a lot of work that would require community effort.

tazsingh:
Ah yeah looks like if you set the Target to Chrome 58, it compiles appropriately 👍

necolas:
I'm going to stick with a single build and the browser support required means pretty much everything needs to be converted to ES5. Will probably just avoid default argument values

hzoo:
Ok just saying we can make a change on Babel's side but that sounds good in the meantime.

necolas:
Oh I see! Thanks 👍

tazsingh:
@necolas Can you let me know if anything further needs to be done here? Would like to get this merged such that we can use it within our apps. Thanks!

necolas:
Thanks. Available in 0.0.96

PR #496, opened 2017-05-31
Update README.md

msand:
Add link and description of related project https://github.com/msand/estructor
A React UI Builder, and a stack to enable in-browser cross platform development of components, pages and universal apps.
It has a component marketplace, CI/CD pipeline, and examples of most everything you need for simple projects and prototypes. It is based on Este, Structor and React Native for Web. It also integrates Apollo and CodePush into Este.


necolas:
Thanks for sharing your project. It's not what I'm looking to include among related projects though

msand:
Alright, thought I would give it a shot :)

PR #514, opened 2017-06-13
added support to -webkit-user-drag

vaukalak:
In a app you can have an image as background. If you drag something over this image, it starts dragging in chrome.

**Test plan**

There's appropriate section in storybook.

**This pull request**

- [ ] includes documentation
- [ ] includes tests

### DIFF
--- a/docs/storybook/components/Image/ImageExample.js
+++ b/docs/storybook/components/Image/ImageExample.js
@@ -291,6 +291,39 @@ const examples = [
     }
   },
   {
+    title: 'Webkit User Drag',
+    description: 'You can modify -webkit-user-drag property using style userDrag: \'none\'',
+    render() {
+      return (
+        <View style={styles.horizontal}>
+          <View style={{ marginRight: 100 }}>
+            <Text>Draggable</Text>
+            <Image
+              source={smallImage}
+              style={[
+                styles.base,
+                styles.background,
+              ]}
+            />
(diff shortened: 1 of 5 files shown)


necolas:
What are the conditions? Mouse use? Maybe setting `draggable=false` on the internal `img` would be simpler

vaukalak:
Hey @necolas yeah, this also can be used. You'd like `Image` to expose a boolean property `draggable`? Or you'd like to set that value in base of `userDrag` style?

necolas:
Let's make it an internal implementation detail for now, no change to the API

vaukalak:
@necolas in this case ALL images will not be draggable, which means, that user will not be able to pick an image and drag and drop it on the desktop. Are you sure, this is a good solution?

necolas:
Ah, so what are some use cases for disabling drag?

vaukalak:
@necolas in my project I have an image and a draggable (by panresponder) overlay over an image. This works fine, but the background image is dragged as a default d&d chrome behavior for images if users clicks slightly outside of the overlay...
e3fd56c2fb9313fd7b1dfba670d1a108
I think draggable overlays over images is the most common use case for disabling background images, but with hight probably there a lot of similar use cases.

necolas:
Thanks. OK let's add a web-specific `draggable` prop to `Image` for this

Issue #517, opened 2017-06-14
CSS rendering much slower than html output in iOS webview


Kinka:
I'm encoutering a rendering problem in ios browser, ios version 10.6. This is the test page: https://fudao.qq.com/rn4web/discover.html

The problem is that there is before the content with CSS styles applied I can saw the version without CSS styles yet. Android browser has no such problem. I tried extracting the css using `StyleSheet.renderToString`, and copy them into page's html, then got a better performance.

So it seems to be a problem css rendering slower than html output?

necolas:
I thought this might come up at some point. Try commenting out the requestAnimationFrame wrapper in `node_modules/react-native-web/dist/apis/StyleSheet/StyleManager.js`

Kinka:
Yes, commenting out makes better result. So what's your concern adding requestAnimationFrame, the problem I mention above occurs almost every time.

necolas:
Yeah we'll want to remove the use of requestAnimationFrame

Kinka:
I think of another question, is mobile.twitter.com using react-native-web?  I wonder how does it merge so many classnames into just several ones. Too many classnames would influence the rendering performance?

necolas:
mobile.twitter.com was already using css-modules, so it uses a limited fork of react-native-web that replaces the library's styles with css-modules definitions. There doesn't appear to be a problem with using many class names

PR #519, opened 2017-06-15
Remove requestAnimationFrame from StyleManager

paularmstrong:
**This patch solves the following problem** Using `requestAnimationFrame` while registering styles/classes slows down iOS/Mobile Webkit. By removing it, it's possible that we add a little bit of overhead, slowing down current cycles, but should be mostly unnoticed.

**Test plan** Manually tested running `npm run docs:start` and checking that everything looked okay in the storybook. Also did a little perf evaluation using Chrome's devtools to see if this caused any extra jank or thrashing, but didn't see anything sticking out. Not sure what to look for, exactly, but happy to find specific points if anyone has anything in mind that this would directly affect.

Fixes gh-517

**This pull request**

- [ ] includes documentation
- [ ] includes tests

### DIFF
--- a/src/apis/StyleSheet/StyleManager.js
+++ b/src/apis/StyleSheet/StyleManager.js
@@ -1,7 +1,6 @@
 import { canUseDOM } from 'fbjs/lib/ExecutionEnvironment';
 import generateCss from './generateCss';
 import hash from './hash';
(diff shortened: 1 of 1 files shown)


necolas:
Thanks. Applied as 65d5a89040f8a611b9c533b2b8d45cda17e56f8f

Issue #524, opened 2017-06-16
TextInput: Support shift + enter to add new line when multiline is enabled?


AlmaasAre:
**What is the current behavior?**

When multiline is enabled in TextInput, shift + enter adds a new line but also triggers onSubmitEnd-callback. 
Link to minimal test case: [codepen](https://codepen.io/anon/pen/WOoPRy?editors=0010)

**What is the expected behaviour?**
shift + enter adds a new line without triggering onSubmitEnd-callback.

OS: 
Device:
Browser: Chrome
React Native for Web (version): 0.0.94
React (version): 15.5.4

PR #526, opened 2017-06-16
Prevent shift+enter in multiline TextInput from triggering onSubmitEditing

paularmstrong:
When multiline is enabled in TextInput, shift + enter should add a new line only, but also triggers onSubmitEnd-callback.

Fixes gh-524.

**Test plan**

* Added test
* Manually tested *Multiline blur on submit* example in storybook

**This pull request**

- [ ] includes documentation
- [x] includes tests

### DIFF
--- a/src/components/TextInput/__tests__/index-test.js
+++ b/src/components/TextInput/__tests__/index-test.js
@@ -170,6 +170,19 @@ describe('components/TextInput', () => {
         done();
       }
     });
+
+    test('multi-line input', done => {
+      const input = findNativeTextarea(
+        mount(<TextInput defaultValue="12345" multiline onSubmitEditing={onSubmitEditing} />)
+      );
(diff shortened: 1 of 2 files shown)


necolas:
There are also these notes in the RN docs that we should make sure remain true:

> Note that for multiline fields, setting `blurOnSubmit` to `true` means that pressing return will blur the field and trigger the `onSubmitEditing` event instead of inserting a newline into the field.

>  `onSubmitEditing`: Invalid if `multiline={true}` is specified.

paularmstrong:
@necolas Doesn't seem that there's any warning or error to be thrown for `onSubmitEditing` invalid if `multiline={true}`. So I just left it suppressed. LMK if there's anything else I missed here.

necolas:
Thanks, applied in 535a7b702712f50120a3d507fe6c7752da8959d9

Issue #518, opened 2017-06-14
TextInput: onKeyPress does not return correct values


sondremare:
**What is the current behavior?**

onKeyPress does not trigger when backspace is pressed.
Link to minimal test case: https://codepen.io/anon/pen/PjGJyx?editors=0010

**What is the expected behaviour?**
That onKeyPress triggers when backspace is pressed. 

I know that onKeyPress is not triggered for backspace on regular html input-fields.
However, with those one has the option to listen to onKeyDown, which is triggered for backspace. 

**Steps to reproduce**

1. Open dev tools, more specifically the console.
2. Focus the TextInput and type something, every key is logged.
2. Hit backspace, nothing is logged. 

**Environment (include versions)**

OS: macOS Sierra 10.12
Device: Macbook Pro
Browser: Chrome 58
React Native for Web (version): 0.0.94
React (version): 15.5.4

necolas:
It looks like `onKeyPress` should be returning the string value of each key too

PR #537, opened 2017-06-22
feat: improve handling of keyboard events on text inputs

kennethlynne:
This is a breaking change! 

Match the current react-native API for key press events on text inputs. See https://jsfiddle.net/kennethlynne/dL3ng4cc/ to test how different keyboard input is mapped to text

React native's API:
https://facebook.github.io/react-native/docs/textinput.html#onkeypress

Closes https://github.com/necolas/react-native-web/issues/518


.gitignore
```diff
@@ -1,2 +1,4 @@
 node_modules
 dist
+dist-examples
+.idea
```
necolas: Please revert these changes

src/components/TextInput/index.js
```diff
@@ -228,6 +228,7 @@ class TextInput extends Component {
       onChange: normalizeEventHandler(this._handleChange),
       onFocus: normalizeEventHandler(this._handleFocus),
       onKeyPress: normalizeEventHandler(this._handleKeyPress),
+      onKeyDown: normalizeEventHandler(this._handleKeyDown),
```
necolas: It looks like we don't need to use `normalizeEventHandler` here or for `onKeyPress`, as we're not using the native event or passing it to any prop handlers within the component.

kennethlynne:
It does not look like you use semver correctly, so I have no idea how to bump the package version to indicate that this is a breaking change

src/components/TextInput/index.js
```diff
@@ -275,12 +276,42 @@ class TextInput extends Component {
...
+      // space
+      else if (e.which === 32) {
+        keyValue = ' ';
+      }
+      else {
+        // we trim to only care about ones that has a textual representation
+        if (e.shiftKey) {
+          keyValue = String.fromCharCode(e.which).trim();
+        } else {
+          keyValue = String.fromCharCode(e.which).toLowerCase().trim();
```
necolas: I wonder if it's worth caching the results rather than recomputing each key stroke

necolas:
> It does not look like you use semver correctly

I haven't published a stable release yet

> I have no idea how to bump the package version to indicate that this is a breaking change

That's ok as it's not your responsibility to do that in a PR

kennethlynne:
Ok, thanks for the clarification @necolas, makes sense! :)

.gitignore
kennethlynne: Done

src/components/TextInput/index.js
kennethlynne: Fixed

kennethlynne:
Fixed all requested changes and successfully ran `yarn lint && yarn fmt` @necolas

kennethlynne:
I'm working on tests now

kennethlynne:
Tests added and all requested changes fixed  @necolas

necolas:
Thanks!

necolas:
I wonder if we should add support for `Shift+Enter` as that's quite commonly used in textareas on web

PR #540, opened 2017-06-23
Support arrow key press events on text input

### LINKED ISSUE #539: Forward arrow key presses on text input
React-native will forward key press events like enter and backspace as "Enter" and "Backspace" for the key property of the nativeEvent payload. 

React-native-web probably needs to support some more use cases since it is used on desktop. 

I currently miss being able to listen for left and right arrow key presses.

I suggest firing events for arrow presses, with the key payload set to "Left arrow", "Up arrow", "Down arrow" and "Right arrow" respectively. I can send a PR for this ASAP if this is agreed on.

kennethlynne:
Forward arrow keys on key press in text inputs

**This pull request**

- [x] includes documentation
- [x] includes tests

Closes #539

### DIFF
--- a/docs/components/TextInput.md
+++ b/docs/components/TextInput.md
@@ -104,7 +104,8 @@ Callback that is called when the text input is focused.
 **onKeyPress**: ?function
 
 Callback that is called when a key is pressed. This will be called with `{
-nativeEvent: { key: keyValue } }` where keyValue is 'Enter` or 'Backspace' for
+nativeEvent: { key: keyValue } }` where keyValue is 
+'Enter', 'Backspace', 'Left arrow', 'Right arrow', 'Down arrow' and 'Up arrow' for
 respective keys and the typed-in character otherwise including ' ' for space.
 Fires before onChange callbacks.
 
--- a/src/components/TextInput/__tests__/index-test.js
+++ b/src/components/TextInput/__tests__/index-test.js
@@ -174,6 +174,42 @@ describe('components/TextInput', () => {
       }
     });
 
+    test('up arrow', done => {
+      const input = findNativeInput(mount(<TextInput onKeyPress={onKeyPress} />));
+      input.simulate('keyDown', { which: 38 });
+      function onKeyPress(e) {
+        expect(e.nativeEvent.key).toEqual('Up arrow');
+        done();
+      }
+    });
+
+    test('down arrow', done => {
+      const input = findNativeInput(mount(<TextInput onKeyPress={onKeyPress} />));
+      input.simulate('keyDown', { which: 40 });
+      function onKeyPress(e) {
+        expect(e.nativeEvent.key).toEqual('Down arrow');
+        done();
+      }
+    });
+
+    test('left arrow', done => {
+      const input = findNativeInput(mount(<TextInput onKeyPress={onKeyPress} />));
+      input.simulate('keyDown', { which: 37 });
+      function onKeyPress(e) {
+        expect(e.nativeEvent.key).toEqual('Left arrow');
+        done();
+      }
+    });
+
+    test('right arrow', done => {
+      const input = findNativeInput(mount(<TextInput onKeyPress={onKeyPress} />));
+      input.simulate('keyDown', { which: 39 });
+      function onKeyPress(e) {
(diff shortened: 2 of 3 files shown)


necolas:
Pressing arrow keys within a text input will change the selection state. What is the use case for needing to know about them within onKeyPress?

necolas:
Closing as it's not clear what the use case is. Let me know otherwise. Thanks

thenewvu:
I'm working on an use case that relates to this PR, it's to implement a text editor, a paragraph contains many TextInput instances, each presents a rich text, handling arrow keys allows to navigate the cursor in the entire paragraph, because, for example, pressing `Arrow Left` on the leftmost of `TextInput`s doesn't trigger a selection state change.

Of course, this PR is specific for web platform, on native platforms, there's no arrow keys.

necolas:
Using arrow keys to navigate between text inputs within a paragraph that's part of a text editor? Does that mean you're trying to implement a code editor using TextInput? If so, that's the wrong tool for that job

thenewvu:
@necolas What does make you think it's the wrong tool for the job ? Just curious and opened to new ideas, not trying to argue with you. "The job" actually is my side-project, I've tried `contentEditable` and plain `input`, but they did not meet my target.

RangerMauve:
@thenewvu TextInput can't hold formatted text, for one

thenewvu:
I'm not sure why this doesn't appear here:

```
@thenewvu TextInput can't hold formatted text, for one
```

I think you're misunderstood (what I said is not clear enough), I'm using `Text` to highlight texts actually, it will be replaced by a `TextInput` when it gains focus. Moreover, `TextInput` does support formatted text ([isn't it here](http://necolas.github.io/react-native-web/storybook/?selectedKind=Components&selectedStory=TextInput&full=0&down=1&left=1&panelRight=0&downPanel=kadirahq%2Fstorybook-addon-actions%2Factions-panel)). Anyway, just want to tell you my case, this PR is useful, at least for me.

aaroncraig10e:
I have a use case which is to implement key navigation for a drop down menu like control.

I have a text box which I can type something into.  It populates a ListView below it with values derived from the current value in the TextInput.

To avoid requiring the user to move their hands off the keyboard and to the mouse, I'd like to listen for arrow key events to highlight items in the list view.

The TextInput should not lose focus so that if the user decides to type more, they can.

It's a very common UI pattern.  Just type something into the Chrome location bar.

emgroves:
@necolas i'm also looking to do the same as @aaroncraig10e - if there's a different way I should be looking to accomplish this... I'm all ears

alexandergunnarson:
I have another use case for this. I'm trying to create a spreadsheet app and I want the user to be able to navigate easily with arrow keys, but the up arrow key causes the selection to go to the beginning of the input and the down arrow key causes it to go to the end; only *after* this does the focus move to the cell above or below. `event.preventDefault()` with `onKeyDown` works to avoid this undesirable behavior, but that's only if I use a raw `<input>` rather than a `TextInput`, which ignores the `onKeyDown` handler. This is not handled by taking charge of the `selection` property (which seems not to work in practice, and based on [the source](https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/TextInput/index.js) it seems to be disabled), because there's no way to capture arrow key events in `onKeyPress` in order to change the selection. Thoughts?

alexandergunnarson:
A related issue is that I can't `.preventDefault` on "Enter" or "Tab" with `onKeyPress`; for it to actually override default behavior, that needs to happen in an `onKeyDown` handler. React Native doesn't support `keyDown` behaviors and seems to replace the `keyUp` behavior with `keyPress`. 

I see at least two alternatives to address this:
1) Allow web-specific custom `onKeyDown` handlers
2) Propagate all non-printable keys (including arrow keys) to `onKeyPress`

Number 2 seems better because it doesn't change the API. What do you think @necolas?

achtan:
@necolas can we reopen or even merge this PR, see @aaroncraig10e 's usecae, i have the same

SBS-randerson:
Could really use this. Trying to disable use of the arrow keys in the TextInput.

PR #542, opened 2017-06-23
Fix @providesModule for a few APIs

fmoo:
AppState and AsyncStorage are tagged as "AppRegistry" instead of their
own names.  BackAndroid is also listed as "BackHandler", but I'm not sure
if that was intentional or not, and there are no conflicts, so I left it alone.

Fixes issues with static analysis tools that expect providesModule directives to be unique

**Test plan**

Used git grep to verify consistency across all remaining modules.

**This pull request**

- [ ] includes documentation
- [ ] includes tests


necolas:
Thanks

PR #556, opened 2017-07-08
[change] a11y and layout: button role and DOM props

### LINKED ISSUE #241: Touchable: fix cross-browser flex styling
HTML's `<Button>` tag does not support flex styling in some browsers (e.g. safari), causing the flex styles on `<touchable>` to fail.

Suggest to wrap touchable's content with a `<div>` and assign the flex styles to it so that flex will work.

necolas:
**Problems:**

HTML's native <button> tag doesn't support flex styling in all browsers,
causing layout bugs. And buttons or links created by "createDOMElement"
(without an accessibility role) do not have the correct props.

**Solution:**

The "button" role is rendered to a "div[role=button]" that is focusable
and responds to the same keyboard events as a native button. A native
button can still be rendered using "createDOMElement".

Make "createDOMProps" aware of the component context to ensure style
resets and other props are correctly applied when an accessibility role
is not defined.

**Additionally:**

This patch also adds a new "label" role to support accessible forms.
It maps to a native label element.

Close #241


necolas:
Applied in ad3dee0204df683642ab8cc66bfa9899b27f616a

PR #572, opened 2017-07-23
Update getting started guide

L1fescape:

For beginners it's not apparent what `create-react-app` is doing behind the scenes. It might be helpful to detail some of those steps so they have a better understanding of `react-native-web` and how it works with `react-native`.

**Test plan**

**This pull request**

- [x] includes documentation
- [ ] includes tests

Also I noticed that the link to the Contributing Guidelines links to https://github.com/necolas/react-native-web/CONTRIBUTING.md which is a 404. I believe it should link to https://github.com/necolas/react-native-web/blob/master/.github/CONTRIBUTING.md

### DIFF
--- a/docs/guides/getting-started.md
+++ b/docs/guides/getting-started.md
@@ -8,12 +8,11 @@ with React Native for Web.
 If you haven't done so already, initialize your project:
 
 ```
-$ npm i -g yarn
-$ yarn global add react-native
+$ npm i -g react-native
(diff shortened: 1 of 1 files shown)


necolas:
Thanks, I made some small changes but I won't be merging this as it repeats much existing information.

L1fescape:
Gotcha, makes sense.

PR #583, opened 2017-07-31
Add tests for button component

L1fescape:
Let me know if there are any additional tests you'd like implemented! I was unable to test default and disabled style values because `StyleSheet.create` returns a hash of references to the classnames generated for those styles rather than the actual style values. Idk a way around that.

### DIFF
--- a/src/components/Button/__tests__/index-test.js
+++ b/src/components/Button/__tests__/index-test.js
@@ -4,43 +4,21 @@ import React from 'react';
 import Button from '..';
 import { mount, shallow } from 'enzyme';
 
-const findNativeButton = wrapper => wrapper.find('[role="button"]');
+const findNativeButton = wrapper => wrapper.getDOMNode();
 
 describe('components/Button', () => {
-  describe('rendered element', () => {
-    test('is a "div"', () => {
-      const onPress = () => {}
-      const button = findNativeButton(mount(<Button onPress={onPress} title="" />));
-      expect(button.type()).toBe('div');
-    });
-
-    test('can display text', () => {
-      const onPress = () => {}
-      const text = 'Click me';
-      const component = mount(<Button onPress={onPress} title={text} />);
-      expect(component.text()).toEqual(text);
-    });
+  test('prop "title"', () => {
+    const onPress = () => {}
+    const text = 'Click me';
+    const component = mount(<Button onPress={onPress} title={text} />);
+    expect(component.text()).toEqual(text);
   });
 
   test('prop "color"', () => {
     const onPress = () => {}
     const color = 'blue';
     const button = findNativeButton(mount(<Button color={color} onPress={onPress} title="" />));
-    expect(button.prop('style')).toEqual({backgroundColor: color});
-  });
-
-  test('prop "disabled"', () => {
-    const onPress = () => {}
-    // true
-    let component = shallow(<Button disabled={true} onPress={onPress} title="" />);
-    expect(component.prop('disabled')).toEqual(true);
-    component = shallow(<Button disabled onPress={onPress} title="" />);
-    expect(component.prop('disabled')).toEqual(true);
-    // false
-    component = shallow(<Button disabled={false} onPress={onPress} title="" />);
-    expect(component.prop('disabled')).toEqual(false);
-    component = shallow(<Button onPress={onPress} title="" />);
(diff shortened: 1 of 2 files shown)


src/components/Button/__tests__/index-test.js
```diff
@@ -1,5 +1,52 @@
...
+import React from 'react';
+import Button from '..';
+import { mount, shallow } from 'enzyme';
+
+const findNativeButton = wrapper => wrapper.find('[role="button"]');
+
 describe('components/Button', () => {
-  test.skip('NO TEST COVERAGE', () => {});
+  describe('rendered element', () => {
+    test('is a "div"', () => {
```
necolas: This test is not necessary, as it is testing the implementation details of `Button` dependencies

src/components/Button/__tests__/index-test.js
```diff
@@ -1,5 +1,52 @@
 /* eslint-env jasmine, jest */
 
+import React from 'react';
+import Button from '..';
+import { mount, shallow } from 'enzyme';
+
+const findNativeButton = wrapper => wrapper.find('[role="button"]');
```
necolas: Can you use something like `getDOMNode` instead of leaking knowledge about how `accessibilityRole` is implemented into this test?

src/components/Button/__tests__/index-test.js
```diff
@@ -1,5 +1,52 @@
...
+  });
+
+  test('prop "color"', () => {
+    const onPress = () => {}
+    const color = 'blue';
+    const button = findNativeButton(mount(<Button color={color} onPress={onPress} title="" />));
+    expect(button.prop('style')).toEqual({backgroundColor: color});
+  });
+
+  test('prop "disabled"', () => {
```
necolas: Is this really testing anything?

src/components/Button/index.js
```diff
@@ -28,7 +28,8 @@ class Button extends Component {
   };
 
   render() {
-    const { accessibilityLabel, color, disabled, onPress, testID, title } = this.props;
+    const { accessibilityLabel, color, onPress, testID, title } = this.props;
+    const disabled = this.props.disabled !== undefined ? this.props.disabled : 'disabled' in this.props;
```
necolas: Why this change?

L1fescape: To match the html spec for boolean attributes. The presence of a boolean attribute results in a true value regardless of what that value is set to, which I now realize means `diabled={false}` should also equal true. This is probably undesirable behavior so I'll just revert this change.

src/components/Button/__tests__/index-test.js
L1fescape: Haha nope, not really. I could update it to test that the `disabled` prop prevents the button from being pressed:

```javascript
  test('prop "disabled" should prevent "onPress" from firing', () => {
    const onPress = jest.fn();
    const component = shallow(<Button disabled={true} onPress={onPress} title="" />);
    component.simulate('press');
    expect(onPress).not.toHaveBeenCalled();
  });
```

however I'm getting this failure when I do that: 

```
expect(jest.fn()).not.toHaveBeenCalled()

    Expected mock function not to be called but it was called with:
      Array []
```

Could this be an issue with simulating press events in enzyme?

src/components/Button/__tests__/index-test.js
```diff
@@ -1,5 +1,30 @@
...
+    const text = 'Click me';
+    const component = mount(<Button onPress={onPress} title={text} />);
+    expect(component.text()).toEqual(text);
+  });
+
+  test('prop "color"', () => {
+    const onPress = () => {}
+    const color = 'blue';
+    const button = findNativeButton(mount(<Button color={color} onPress={onPress} title="" />));
+    expect(button.style.backgroundColor).toEqual(color);
```
L1fescape: Would you prefer this test use snapshots instead?

src/components/Button/__tests__/index-test.js
```diff
@@ -1,5 +1,30 @@
...
+  });
+
+  test('prop "onPress"', () => {
+    const onPress = jest.fn();
+    const component = shallow(<Button onPress={onPress} title="" />);
+    component.simulate('press');
```
doxiaodong: @L1fescape Sorry for the question.  How do do make press simulate work in enzyme or jest?

PR #586, opened 2017-08-01
Fix async race condition in Touchable mixin

### LINKED ISSUE #585: TouchableOpacity intermittently fails to call onPress handler
Thanks for all the great work! We're using this project extensively and it has been a pleasure to work with.

Report a bug.

When a TouchableOpacity is used to toggle a change that changes the position of the `TouchableOpacity` view, subsequent clicks may fail to trigger the `onPress` handler when using a Mac touchpad. 

This bug is not reproducible with an external mouse, or when a "tap" on the touchpad is used.

**If the current behavior is a bug, please provide the steps to reproduce.**

[https://codepen.io/sam-10e/pen/RZajrV?editors=0010](https://codepen.io/sam-10e/pen/RZajrV?editors=0010)

1. Open the demo, click the black box using a Mac touchpad (make sure to click and not tap)
2. The box should move to a new position, click again.
3. Repeat until the box fails to move. It will still change opacity.
4. Click again and the box will move.

The TouchableHighlight should consistently detect presses, regardless of input type (mouse, touchpad, etc).

* OS: OSX Version 10.12.6 (16G29)
* Device: Macbook Pro
* Browser: Chrome, Firefox (latest)
* React Native for Web (version): 0.0.114+ (may also happen on older versions)
* React (version):

** Additional Notes **

This _only_ happens with TouchableOpacity and _not_ TouchableHighlight. It appears that when `touchableGetHighlightDelayMS()` returns 0, the `this._handleDelay` method is called immediately, and it may complete before the callback passed to `UIManager.measure` in  `this._remeasureMetricsOnActivation` has been executed. Thus the check to determine if the action occurred within the bounds of the TouchableOpacity component fails, and it does not call the `onPress` handler.

sam-10e:
When a Touchable component returns 0 for `touchableGetHighlightDelayMS()` the mixin may occasionally try to detect if the press event occurred on the hit area _before_ the hit area has been determined. Clearing out this value ensures that `positionOnActivate` is recalculated before that takes place.

This fixes #585

### DIFF
--- a/src/components/Touchable/Touchable.js
+++ b/src/components/Touchable/Touchable.js
@@ -399,6 +399,7 @@ const TouchableMixin = {
     if (delayMS !== 0) {
       this.touchableDelayTimeout = setTimeout(this._handleDelay.bind(this, e), delayMS);
     } else {
+      this.state.touchable.positionOnActivate = null;
       this._handleDelay(e);
     }
 


necolas:
You might want to submit this in the original implementation in React Native too

sam-10e:
Thanks- I'll look into seeing if I can reproduce it in React Native.

PR #589, opened 2017-08-02
Fix example in README

MaxPleaner:
All I know is that it wasn't working to read the dependencies from the `react-native` package. I had to use `react-native-web` instead.

Otherwise, I got `Cannot find module` errors when attempting to read any properties of ReactNative.

**Before submitting a pull request,** please make sure you have followed the steps the CONTRIBUTING guide.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,7 @@ Guides:
 
 ```js
 import React from 'react'
-import { AppRegistry, Image, StyleSheet, Text, View } from 'react-native'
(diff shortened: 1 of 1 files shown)


RangerMauve:
You're supposed to alias `react-native` to point to `react-native-web` in your build configuration. That way you can build the same code for react-native using the react native build tools, and use whatever browser build system to build the code for the web.

necolas:
Read the docs please

PR #597, opened 2017-08-08
Kim/normalizenativeevent check before bind

kim-10e:
The two functions normalizeTouchEvent() and normalizeMouseEvent() in `module/normalizeNativeEvent/index.js` assume the native event object contains the following three functions:

- preventDefault
- stopImmediatePropagation
- stopPropagation

However, events simulated by React Test Util (e.g. via Enzyme) do not contain these three functions, causing a crash when React Native Web tries to normalize (simulated) native events.

Adding a simple check before attempting to bind said three properties fixes the problem and enables click and touch events to be simulated when testing React components using Enzyme and JSDOM.

* React Native for Web (version): 0.0.119
* React (version): 15.4.2

### DIFF
--- a/src/modules/normalizeNativeEvent/__tests__/index-test.js
+++ b/src/modules/normalizeNativeEvent/__tests__/index-test.js
@@ -1,0 +1,40 @@
+/* eslint-env jasmine, jest */
+
+import normalizeNativeEvent from '../index';
+
+describe('modules/normalizeNativeEvent', () => {
+  it('handles mouse events without preventDefault, stopPropagation and stopImmediatePropagation', () => {
+    const event = {
+      type: 'mouse'
+    };
+
+    normalizeNativeEvent(event);
+  });
+
+  it('handles mouse events with preventDefault, stopPropagation and stopImmediatePropagation', () => {
+    const event = {
+      type: 'mouse',
+      preventDefault: () => {},
+      stopPropagation: () => {},
+      stopImmediatePropagation: () => {}
+    };
+
+    normalizeNativeEvent(event);
+  })
+
+  it('handles touch events without preventDefault, stopPropagation and stopImmediatePropagation', () => {
+    const event = {};
+
+    normalizeNativeEvent(event);
+  });
+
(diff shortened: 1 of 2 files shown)


kim-10e:
This is a proposed fix for https://github.com/necolas/react-native-web/issues/598, which allows React Native Web components to be mounted and tested using Enzyme & JSDOM.

necolas:
Why do this rather than expecting people to create accurate mocks for events?

kim-10e:
Because the mock event objects are created by the Simulate functions provided by React Test Utils, which in turn are called by Enzyme's simulate function.

The scenario is this: Enzyme exposes a simulate() function on the wrapper returned when mounting a React component. This function allows you to simulate events on the component. Under the hood, Enzyme uses the Simulate functions provided by the React test utils add-on. Unfortunately, React test utils create mock events without declaring function stubs for preventDefault(), stopPropagation() and stopImmediatePropagation().

This PR provides a simple work-around for supporting events simulated by React test utils and thus Enzyme.

necolas:
I understand [the scenario](https://github.com/necolas/react-native-web/blob/ad3dee0204df683642ab8cc66bfa9899b27f616a/src/modules/createDOMElement/__tests__/index-test.js#L21-L27). But is this a problem unique to React Native for Web?

kim-10e:
So you prefer every invocation of click event to pass in stub functions to the normalize event function can carry on assuming these are always present despite the React test utils implementation clearly not providing them by default? To me that seems like React Native Web making assumptions about event objects that aren't generally shared by the base React implementation (or at least the folks that implemented React test utils)

necolas:
Before you start assuming anything about what I think, I want to understand if this is a problem unique to React Native for Web.

kim-10e:
I'm able to simulate click events with Enzyme and regular React components. I didn't see any open issues against React test utils regarding the lack of the three functions in question on the mock events. Once I saw that the React Native Web normalize event implementation assumes these three functions are always present it felt like the least impact solution to add a simple check and add stub functions when needed in the normalize event function.

Issue #602, opened 2017-08-12
ART does not work on web


naqvitalha:
Importing ART does not work, ideally we should fallback to react js ART

Issue #601, opened 2017-08-12
Setting pagingEnabled on scrollview does not enable the snap effect


naqvitalha:
Missing functionality, this is relevant to touch bases mobile devices

necolas:
Duplicate

PR #603, opened 2017-08-12
Added paging support to scrollview and ART

naqvitalha:
Addresses #601 , #602


naqvitalha:
Not sure how the new base should align with existing implementation such that everything remains composable. Help out, I have to add scrollTo logic

package.json
```diff
@@ -65,7 +65,8 @@
     "inline-style-prefixer": "^3.0.6",
     "normalize-css-color": "^1.0.2",
     "prop-types": "^15.5.10",
-    "react-timer-mixin": "^0.13.3"
+    "react-timer-mixin": "^0.13.3",
+    "@ecliptic/react-art": "0.15.4-fork.1"
```
rmevans9: Why the this fork instead of the one owned by FB?

https://www.npmjs.com/package/react-art

naqvitalha: Check out this issue reactjs/react-art#113 
The npm package no longer worked with latest react version, unfortunately no one is publishing it :(

necolas: I'll consider this once the official package can be used

.gitignore
```diff
@@ -1,2 +1,3 @@
 node_modules
 dist
+.idea
```
necolas: Please remove

src/components/ScrollView/PagingScrollViewBase.js
```diff
@@ -0,0 +1,224 @@
...
+            },
+            get width() {
+                return parent._parentWidth;
+            }
+        }
+    },
+    timeStamp: Date.now()
+});
+
+export default class PagingScrollViewBase extends Component {
```
necolas: Why is this a new component?

src/components/View/index.js
```diff
@@ -39,6 +39,7 @@ class View extends Component {
     const {
       hitSlop,
       style,
+      domRef,
```
necolas: ?

necolas:
Also, please break this into 2 PRs. One for ART and the other for paging. Thanks

package.json
naqvitalha: Why don't you fork and publish one yourself? We can use that

.gitignore
naqvitalha: Will do

src/components/ScrollView/PagingScrollViewBase.js
naqvitalha: Because it does not really use the base at all. Scroll events are raised using onTouchMove, you can snap browser scrolling. Switching the base seemed like a good idea to me. I'm open to ideas here since you know the code better

src/components/View/index.js
naqvitalha: I had wrtiten a comment around this, onLayout does not fire and measure doesn't work on one of the views that I created. Needed dom node for clientWidth, let me know the alternative

naqvitalha: Didn't have the time to look into the missing callbacks issue

naqvitalha:
I know there should be multiple PRs. I'm mostly trying to understand contributing to this project right now. Just reply to my question, we can close this PR post that. I'll raise two in the future.

necolas:
> Just reply to my question, we can close this PR post that.

Which question is that?

naqvitalha:
I mean respond to my replies to your review comments.

package.json
necolas: No thanks. Let's get the official package updated

src/components/ScrollView/PagingScrollViewBase.js
necolas: What about scrolling without touch?

src/components/View/index.js
necolas: OK. I don't think I will accept a patch that hacks around what is apparently another issue though

necolas:
I think you can break this into 2 PRs now

necolas:
Closing this PR for now. ART was added in 321051b72366666ba6e28985cff017cb6d3c506d.

PR #608, opened 2017-08-17
Add babel plugin to rewrite import paths

paularmstrong:
**Problem:**

1. Aliasing `react-native` to `react-native-web` currently requires using a setting in webpack.
2. Imports don't help tree-shaking, because everything is exported from the top-level of react-native-web.

**Solution:**

Add a babel plugin that:

1. Rewrites `react-native` to `react-native-web`
2. Finds all of the apis, components, and modules and rewrites imports to `react-native-web/dist/${type}/${importName}`

**Discussion, Problems:**

Only supports `require` variable declarator syntax that uses object patterns, ie `const { View } = require('react-native');` There are a lot of other ways to do this and supporting them all is significantly more work.

### DIFF
--- a/babel/__tests__/__snapshots__/rewrite-paths-test.js.snap
+++ b/babel/__tests__/__snapshots__/rewrite-paths-test.js.snap
@@ -1,0 +1,104 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`1. Rewrite react-native paths for react-native-web 1`] = `
+"
+import { View } from 'react-native';
+
+      ↓ ↓ ↓ ↓ ↓ ↓
+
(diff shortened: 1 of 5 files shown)


necolas:
Thanks Paul. I changed to a curated list as it's simpler at this stage. Might well use a flat directory of public exports later.

https://github.com/necolas/react-native-web/tree/gh-pr/608

I noticed that the `require` form is not transformed, e.g., `const { View } = require('react-native')`

paularmstrong:
> I noticed that the require form is not transformed, e.g., `const { View } = require('react-native')`

Ah yes... I always forget about those. I can update soon

necolas:
Thanks, this is now merged

PR #614, opened 2017-08-24
Change default flexBasis value

### LINKED ISSUE #574: resolvedStyle.flexBasis = '0%' or  resolvedStyle.flexBasis = 'auto'
loaction：..\react-native-web\src\apis\StyleSheet\createReactDOMStyle.js
on 172line

```
 case 'flex': {
        if (value > 0) {
          resolvedStyle.flexGrow = value;
          resolvedStyle.flexShrink = 1;
          resolvedStyle.flexBasis = 'auto';   //172line
        }
```
1、
  resolvedStyle.flexBasis = '0%';
I run web on iphone,but The layout of many interfaces are compressed on my  project。
on chrome and on android phone are no problem。 

2、
then I change react-native-web's  source code,
become resolvedStyle.flexBasis = 'auto';
I run web on iphone,The layout of many interfaces are Become very normal.
the same as on android phone。
but，The interface jump is very slow。After the button is clicked, it will take a long time to respond.
Is this a performance problem or a compatibility issue？
Or what's the other problem?

3、I've changed react-native-web's source code, and I'm not going to have a problem？
resolvedStyle.flexBasis = '0%' or  resolvedStyle.flexBasis = 'auto'
Will it affect the rest of the world?

dtdo-timfarley:
Fixes #574 by changing the default value of `flexBasis` applied to shorthand `flex: 1`.  Safari has a known bug displaying percentage-based values for `flex-basis` when `flex-direction: column` and `justify-content: center` [(details here)](https://github.com/philipwalton/flexbugs/issues/192).

Using `flexBasis: '0px'` instead of `flexBasis: 0%` fixes the issue.


necolas:
It used to be `auto`, but that's not the correct value and it means something different (causes different layout).

dtdo-timfarley:
@necolas I've updated the PR to fix the tests and to use `0px` instead of `auto` – this should result in the same layout behavior, but also fixes the Safari bug mentioned above.

dtdo-timfarley:
Changing it to `0px` did fix the issue in Safari, but it seems to break our layout in Chrome.  The single-property shorthand for `flex` is pretty strange, as it can influence different CSS properties, depending on what its value is!

[According to this guide](https://css-tricks.com/almanac/properties/f/flex/#article-header-id-5), it seems like `0px` should be the default value for `flex-basis`, although in Chrome I see that a shorthand of `flex: 1` expands to `flex-basis: 0%`.  And [the spec](https://www.w3.org/TR/css-flexbox-1/#valdef-flex-flex-basis) seems to suggest a unitless `0` as the default value when no third property is defined.

I'm going to close this PR – using `auto` does fix our layout in both browsers, but I agree that it isn't the correct value to set as a default.  I'm not sure `0%` is either, but that at least matches Chrome's interpretation of the `flex` shorthand, so won't cause unexpected behavior in the majority of browsers.

necolas:
Hmm, are you saying that `0%` and `0px` produce different rendering in Chrome (and/or other browsers)?

dtdo-timfarley:
Yeah, for our layout at least – using `0%` works in Chrome but not Safari (v10).  Using `0px` works in Safari, but not Chrome.  Using `auto` does work for us in both browsers, but I agree with you that it isn't the right default value, and may cause other layout issues.

Ideally, I think a style declaration of `flex: 1` in a RN component should translate to the web directly as `flex: 1`, instead of expanding it from the shorthand into specific values (as it seems each browsers handle the shorthand properly, but they don't all behave the same when explicitly setting  `flex-basis`. Yay standards.

However, I think the spirit of `react-native-web` is to have layout parity with RN, which I assume does work better with the explicit values, rather than the shorthand.  I closed the PR, because I'm not convinced there's anything broken here.

Our specific problem was also due to the fact that we hadn't upgraded in awhile (like almost a year 😬 ) and we still have places where our styles are assuming `flex: 1` is interpreted [in an outdated, incorrect way by RN](https://github.com/facebook/react-native/wiki/Breaking-Changes#039).

necolas:
Sigh, browsers

PR #641, opened 2017-09-15
[fix] event normalization

necolas:
Work better with simulated events, and avoid crashing if
'nativeEvent.target' isn't an element node.

Close #597


necolas:
997b598de88e86e9bd58a3041c09a03677df754f

PR #648, opened 2017-09-19
[fix] cross-browser flex styles

### LINKED ISSUE #616: ScrollView: content renders in Chrome but not Safari
Bug

Blank page in Safari

if a minimal demo of the problem via Glitch or similar (template:
https://glitch.com/edit/#!/react-native-web-playground).**

1. Go to https://mixed-lightning.glitch.me/ in Safari.
1. Observe blank screen.
1. Go to Chrome and notice renders correctly:
image
1. Source code is at https://glitch.com/edit/#!/mixed-lightning?path=src/App/index.js:1:0:

```
  cellStyle: {
    backgroundColor: 'white',
    width: 50,
    borderWidth: 1,
    borderStyle: 'solid'
  }
});

class App extends Component {
  render() {
    return (
      <View>
        <ScrollView horizontal={true}>
          <View style={styles.cellStyle}>
            <Text>hi</Text>
          </View>
          <View style={styles.cellStyle}>
            <Text>hi</Text>
          </View>
        </ScrollView>
      </View>
    );  
  }
}
```

Note that adding the following attribute to the outer `View` element fixes it:
`style={{display: 'block'}}`
But it doesn't seem this should be necessary, as we're using Flexbox.

Safari renders scroll view content like in Chrome.

* OS: macOS Sierra 10.12.6 (16G29)
* Device: MacBook Pro
* Browser: Chrome Version 60.0.3112.101 (Official Build) (64-bit), Safari Version 10.1.2 (12603.3.8)
* React Native for Web (version): .0.119
* React (version): 15.6.1

necolas:
Fix #616

### DIFF
--- a/src/apis/StyleSheet/__tests__/createReactDOMStyle-test.js
+++ b/src/apis/StyleSheet/__tests__/createReactDOMStyle-test.js
@@ -26,44 +26,44 @@ describe('apis/StyleSheet/createReactDOMStyle', () => {
     test('flex defaults', () => {
       expect(createReactDOMStyle({ display: 'flex' })).toEqual({
         display: 'flex',
-        flexShrink: 0,
-        flexBasis: 'auto'
+        flexShrink: '0 !important',
+        flexBasis: 'auto !important'
       });
     });
 
     test('flex: -1', () => {
       expect(createReactDOMStyle({ display: 'flex', flex: -1 })).toEqual({
         display: 'flex',
-        flexGrow: 0,
-        flexShrink: 1,
-        flexBasis: 'auto'
+        flexGrow: '0 !important',
+        flexShrink: '1 !important',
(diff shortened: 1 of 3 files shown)


src/apis/StyleSheet/__tests__/createReactDOMStyle-test.js
```diff
@@ -19,85 +19,89 @@ describe('apis/StyleSheet/createReactDOMStyle', () => {
   test('noop on DOM styles', () => {
     const firstStyle = createReactDOMStyle(reactNativeStyle);
     const secondStyle = createReactDOMStyle(firstStyle);
+    console.log(firstStyle);
+    console.log(secondStyle);
```
paularmstrong: logs

PR #661, opened 2017-09-28
Add react-nativeish

ericwooley:
I've been working on react nativeish for a long time now, finally feel like it's ready for use.


necolas:
Thanks for sharing your work. My intention with the starter kit links is to surface a shortlist of simpler, maintained setups rather than more opinionated and filled-out app architectures. I think react-nativeish falls into the latter, so I won't be including it here.

ericwooley:
mmk. Just out of curiosity, which part made it too opinionated? It doesn't seem any less opinionated than re-start? I guess I have navigation libraries that they don't have?

PR #665, opened 2017-10-01
Handle ImageLoader getSize failure

### LINKED ISSUE #561: ImageLoader 'getSize' never calls 'failure' callback
See: https://github.com/necolas/react-native-web/blob/02cfbf8/src/modules/ImageLoader/index.js#L22-L41

n3tr:
resolve #561 

- Call failure callback on Image.onerror, Also clearInterval and abort a request.


necolas:
Thanks!

PR #667, opened 2017-10-03
Set package.json module to use ES6 module for webpack/rollup.

AaronFriel:
This just points the `package.json` `module` field at the uncompiled ES6 code, for use with webpack or rollup and tree shaking.

### DIFF
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   "version": "0.1.3",
   "description": "React Native for Web",
   "main": "dist/index.js",
-  "module": "dist/module.js",
+  "module": "src/module.js",
   "files": [
     "babel",
     "dist",


necolas:
This won't work well because the source code needs to be compiled with a particular babel config. You're better off using the provided babel plugin at this stage

necolas:
> The provided babel-plugin causes issues with downstream dependencies as well

Such as?

AaronFriel:
I am submitting this patch for users who _would like_ to be able to use ES6 modules and tree shaking properly. A future patch could add a separate babel config and destination directory for ES5+ES2015 modules, or the `.mjs` format. This is a temporary fix to allow Webpack to use the ES6 sources in your library.

> If you want to address this properly you could patch React Native's babel-preset to allow the module-to-commonjs transform to be disabled via options.

I'm curious why you think this is a viable option for consumers who wish to use ES2015 modules and tree shaking? The `module` field in `package.json` exists for this purpose. As to your prior comment:

> This won't work well because the source code needs to be compiled with a particular babel config. You're better off using the provided babel plugin at this stage

`babel-loader` will take care of this automatically, due to [config chaining](https://github.com/babel/babel/blob/master/packages/babel-core/src/config/build-config-chain.js). It's rather annoying, but the `react-native` preset is again here somewhat painful to work around.

> The provided babel-plugin causes issues with downstream dependencies as well

When trying to use, e.g.: `native-base`, `react-native-easy-grid`, `reactxp`, and others with ES2015 modules and tree shaking, the result is that the `react-native-web/babel` plugin forces imports from the `dist` folder, which results in both the `src` and `dist` folders being placed in chunks by Webpack.

AaronFriel:
With the pull requests I submitted to a handful of projects last night, I am able to produce this set of dependencies on a minimal `react-native-web` project:

image

The important thing to notice is the `src` dependencies. The changes I'm submitting should also remove `lodash.js` shortly. The goal is to make `react-native-web` projects slimmer and hopefully get the gzipped size on the order of ~100kb when targeting modern browsers.

necolas:
Thanks, this is the kind of context that should have been included in the initial PR.

> I'm curious why you think this is a viable option for consumers who wish to use ES2015 modules and tree shaking?

Because it would allow this package to publish compiled modules that retain the ES2015 imports/exports.

> babel-loader will take care of this automatically, due to config chaining

Doesn't that require projects to configure babel to compile packages in `node_modules` anyway?

> When trying to use, e.g.: native-base, react-native-easy-grid, reactxp, and others with ES2015 modules and tree shaking, the result is that the react-native-web/babel plugin forces imports from the dist folder, which results in both the src and dist folders being placed in chunks by Webpack.

How are the RNW `src` files being included at all? And are you trying to use RNW to render `reactxp` components?

AaronFriel:
> Because it would allow this package to publish compiled modules that retain the ES2015 imports/exports.

But it doesn't.

> Doesn't that require projects to configure babel to compile packages in node_modules anyway?

Sure, but right now there's no other way to use the `module` field. This is an incremental step toward a proper `.mjs` build output that gets people using Webpack and Rollup partway there to using ES6 modules and tree shaking throughout.

>How are the RNW src files being included at all? And are you trying to use RNW to render reactxp components?

Attempting to use tree shaking (by referencing a `/src` path or using `module`) and `react-native-web/babel` results in both the `src` and `dist` folder being referenced.

PR #677, opened 2017-10-06
[Babel plugin] Support 'export' statement

n3tr:
Add support transpile `export` statement in babel plugin

```js
// from
export { View, Text } from 'react-native';
export { View as MyView } from 'react-native';
export { default as RNW } from 'react-native';

// to
export { default as View } from 'react-native-web/..';
export { default as Text } from 'react-native-web/..';
export { default as MyView } from 'react-native-web/..';
export { default as RNW } from 'react-native-web';
```

### DIFF
--- a/babel/index.js
+++ b/babel/index.js
@@ -127,6 +127,30 @@ module.exports = function({ types: t }) {
           path.replaceWithMultiple(imports);
         }
       },
+      ExportNamedDeclaration(path) {
+        const { source, specifiers } = path.node;
+        if (source.value === 'react-native' && specifiers.length) {
+          const exports = specifiers
+            .map(specifier => {
+              if (t.isExportSpecifier(specifier)) {
+                const exportName = specifier.exported.name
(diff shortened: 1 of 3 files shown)


necolas:
Thank you!

babel/index.js
```diff
@@ -127,6 +127,31 @@ module.exports = function({ types: t }) {
           path.replaceWithMultiple(imports);
         }
       },
+      ExportNamedDeclaration(path) {
+        const { source, specifiers } = path.node;
+        if (source.value === 'react-native' && specifiers.length) {
```
paularmstrong: This breaks when node chooses the `module` path from `package.json`, because the `dist/module.js` doesn't include a `source` for the `export { ... };` statement.

```
Module build failed: TypeError: project/tools/devserver/hmr-entry.js: Cannot read property 'value' of null
    at PluginPass.ExportNamedDeclaration (project/node_modules/react-native-web/babel/index.js:132:19)
```

PR #681, opened 2017-10-12
wip: Added Extend function to allow custom components

s2jesbak:
this functions allows a user to register a react component to replace or extend the default ones.

So this is possible
`const Modal = class Test extends Component {
  render(){
    return (
      <div className="modal" style={{
        display: this.props.visible ? 'block' : 'none'
      }}>
        {this.props.children}
      </div>
    )
  }
}
ReactNative.OverWrite('Modal',  Modal)`

### DIFF
--- a/src/apis/Extend
+++ b/src/apis/Extend
@@ -1,3 +1,0 @@
-export default function(name, module){
-  this[name] = module
-}
--- a/src/apis/Extend
+++ b/src/apis/Extend
@@ -1,0 +1,3 @@
+export default function(name, module){
+  this[name] = module
+}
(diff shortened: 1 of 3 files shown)


necolas:
Thanks but I don't think this is a necessary addition to the API

s2jesbak:
Ok I can appreciate that you don't think it's necessary. But if I am using the Modal in my react-native application and I use this to compile it to the web, I get an element with a red square. The unimplemented component.

So how is this a solvable solution? 

Do we have to wait for you to make the modal? can we not make the modal ourselves? there are multiple ways to implement modals, what if your implementation doesn't fit our app? what if we want to implement our own designs for components with less markup?

This will also be beneficial as it will allow users to prototype unimplemented components without forking this branch and then publishing their own NPM module (something I intend to do). As you use some sort of NPM alias, using this repo as a git submodule doesn't work so it's much harder to develop components.

Thanks.

PR #691, opened 2017-10-23
Component Slider

teabyii:


lyahdav:
Any reason this hasn't been merged? Any concerns with using it? Right now we're using [this slider](https://codepen.io/necolas/pen/WxkYpv) from @necolas but it doesn't seem to support RTL languages correctly.

lyahdav:
Looks like this was taken from https://github.com/jeanregisser/react-native-slider. Note I added a PR to that repo to support RTL: https://github.com/jeanregisser/react-native-slider/pull/93, but this repo won't work correctly in RTL languages currently.

PR #694, opened 2017-10-25
Fix TypeError: allStylePropTypes[prop] is not a function

brunolemos:
Got this error some times.

image

react-native-web 0.1.9
react 16


brunolemos:
Wait, this is another error. But I'm almost sure I got "allStylePropTypes is undefined" a couple times. Nevermind.

PR #705, opened 2017-11-06
Add Picker and Picker.Item to RNW

kennethkufluk:
Add Picker as a wrapper around <select>.
See #91 Components backlog.


necolas:
Thanks, this formed the basis for this commit to master b7e970f4e6b48ee2c063d01914df1df4556c7d94

necolas:
Ref #91

Issue #599, opened 2017-08-10
Bug: Animated.View is loading multiple copies of React when imported from a react library.


robertgonzales:
I'm using `react-native-web` in a standalone react library. However, I'm unable to import and use the library in another app since any component I export that uses `Animated.View` causes this error to show up:
```
Uncaught Error: addComponentAsRefTo(...): Only a ReactOwner can have refs. 
You might be adding a ref to a component that was not created inside a component's `render` method, 
or you have multiple copies of React loaded (details: https://fb.me/react-refs-must-have-owner).
```

Note: other components seem to work fine. If I replace `Animated.View` with `View` the error resolves.

I'm using `react-native-web@0.0.119` and `react@15.6.1` on the web.

necolas:
Do you have an example that reproduces the issue?

robertgonzales:
I made an example here: https://github.com/robertgonzales/react-native-web-animated-demo

To test, clone the repo and run these commands: `npm install`, `npm run prepublish`, and `npm link`. Then go to any separate react project and run `npm link react-native-web-animated-demo` and then try the following:
```jsx
import { WithAnimated, WithoutAnimated } from "react-native-web-animated-demo"

export default function App() {
 return <WithAnimated /> // error
 // return <WithoutAnimated /> // no error
}
```

ktj:
Maybe this has something to do with https://github.com/animatedjs/animated/blob/master/src/createAnimatedComponent.js
setting refs?

ktj:
The problem maybe in npm https://stackoverflow.com/questions/39371579/symlink-causing-2-versions-of-react

robertgonzales:
Yes, this solves the issue, though unfortunate that I have to eject my CRA project to get it to work. Thanks!

TomClarkson:
Could we use this in the meantime https://www.npmjs.com/package/react-dom-animated ?

PR #708, opened 2017-11-09
Use React 16 compatible version of animated

TomClarkson:


necolas:
It already is, as I don't use the DOM target

TomClarkson:
Sorry, I was receiving the multiple versions of React error a month ago from this issue https://github.com/necolas/react-native-web/issues/599 and I thought this was because the 0.2.0 of animated version wasn't compatible with react 16.

necolas:
Closing / folding this issue into #716

PR #712, opened 2017-11-10
Image: make onLoad fires when the image is cached

### LINKED ISSUE #452: Image: "caching" functionality skips loading callbacks
This probably breaks expectations that at least the `onLoad` callback is called whenever an image is mounted

itszero:
Closes #452.

# Problem

onLoad was not firing before when the image URL is in the URI cache. This breaks the expectation that onLoad should always fire after the image is loaded.

# Solution

Check for the case where the ImageLoader is skipped and fire onLoad accordingly.

### DIFF
--- a/src/components/Image/__tests__/index-test.js
+++ b/src/components/Image/__tests__/index-test.js
@@ -1,6 +1,7 @@
 /* eslint-env jasmine, jest */
 
 import Image from '../';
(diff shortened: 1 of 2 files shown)


itszero:
Thanks @necolas 😄

necolas:
Thank YOU!

PR #715, opened 2017-11-14
Update TextInputState when TextInput is focused

Minishlink:
Problem: when you click on a `TextInput`, the `TextInputState` is not updated, thus, when you call `Keyboard.dismiss`, the `TextInput` is not blurred.
This PR fixes it. I'm not sure if this is the best solution though. 

Maybe `TextInputState._currentlyFocusedNode` should always be `document.activeElement`?


necolas:
Thanks for noticing and reporting this. I pushed what I believe is a broader fix for the issue

PR #724, opened 2017-11-29
Fix NetInfo API to return data directly instead of the event wrapper

comp615:
**Problem**
Right now, when using addEventListener from the NetInfo API, the callback is called with an Event object. This means that end-users cannot use the same handler function as their initial call, and that further, it's not in conformance with the React Native spec. This was a bug in my initial implementation.

**Solution**
Create a wrapped event handler which gets the data out of the event and only passes that to the user-provided handler. The approach is similar to the existing connected functions. Since we are wrapping the handler, we create a key-value list of all handlers so that the user can use removeEventHandler with their initial handler reference. (Otherwise they'd have to take the return result and call .remove())

### DIFF
--- a/src/apis/NetInfo/index.js
+++ b/src/apis/NetInfo/index.js
@@ -43,6 +43,7 @@ const eventTypesMap = {
 const eventTypes = Object.keys(eventTypesMap);
 
 const connectionListeners = [];
+const netInfoListeners = [];
 
 /**
  * Navigator online: https://developer.mozilla.org/en-US/docs/Web/API/NavigatorOnLine/onLine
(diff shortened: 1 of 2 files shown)


src/apis/NetInfo/index.js
```diff
@@ -63,7 +64,9 @@ const NetInfo = {
       };
     }
 
-    connection.addEventListener(eventTypesMap[type], handler);
+    const wrappedHandler = e => handler(e.target);
```
necolas: An object of the wrong type is passed to the handler. `type: ConnectionType` is missing, and `onchange` should not be included. A type check or test would be helpful to catch this.

comp615:
@necolas thanks for finishing this!!! Sorry I was on a trip last weekend and did not have time to respond to the comments.

necolas:
No worries :)

PR #732, opened 2017-12-07
Fixed bug with animations and i18n

lyahdav:
Styles were not correctly getting flipped in `setNativeProps`.


necolas:
Thanks, please can you share a test case that exposes the problem?

lyahdav:
Sorry I didn't get a chance to write a test, haven't had time recently. Thanks for merging!

necolas:
Are you using react-native-web at Uber?

lyahdav:
Yes, I showed you a screen where we use it when I saw you at React Wednesdays at Airbnb a while back.

PR #736, opened 2017-12-20
[fix] corrupt hydrate import breaking v0.2.0

johakr:
I think, version v0.2.0 is broken due to named import of default export


necolas:
Oh thanks, yeah I messed up the switch away from the react-dom export.

necolas:
Should be fixed in 0.2.1. Thanks again

johakr:
Thank you so much for the blazing fast merge. Please keep up the amazing work with this library. ☺

PR #741, opened 2017-12-26
Standalone stylesheet

### LINKED ISSUE #145: StyleSheet: investigate publishing as separate module
The StyleSheet would be very useful on its own, even for those of us not using react-native-web. I have a basic implementation [over here](https://github.com/phuu/flight-redux-playground) but the implementation here is just much better.

I guess the options are:
1. Extract the StyleSheet API and then wrap it for react-native-web
2. Extract the utilities around normalizing, deduping and appending to the `style` element
3. Copy & paste the useful bits out of this repo

I prefer number 2 as it would allow sharing of best practices (DOM perf, prefixing...) without enforcing an API on either.

giuseppeg:
Fixes #145

This branch is WIP and not fully working, I am submitting an early PR to see whether there is any interest in moving this forward and to know what you, @necolas, think about this abstraction. 

Please keep in mind that I am not familiar with RNW and the StyleSheet implementation details and full API – only seen the public API.

I branched off `monorepo` since I thought it could save me some work once that branch is merged.
I am not quite familiar with Lerna's workflow though, especially around linking/bootstraping (which seems to not be working) and building. 

I moved most of the `modules` used exclusively by StyleSheet to its `src/lib` subfolder. The library though still depends on a few shared `modules` which probably should be distributed as an independent package or duplicated, unless we want the package to depend on `react-native-web` (need to double check whether there could be circular dependencies issues).

The library doesn't need to know about `I18nManager`, the user can set an `isRTL` helper.
The dependency from RNW's main components is weak, basically we only need the prop types for validation. Those can be registered in the consumer app or bundled with StyleSheet and the consumer library/app can import them.

TODOs

- [ ] pick up a name for the package
- [ ] fix the build step
- [ ] make tests work
- [ ] create a simple `createElement`
- [ ] bundle the standard propTypes for Image/Text/TextInput/View ?
- [ ] why doesn't `StyleSheet.create` yield `{ className, styles }` (i.e. calls `StyleRegistry.resolve`)? Without `createElement` resolving styles by hand and applying them is not fun. 
- [ ] What is `scripts/inline-style-prefixer/create.js` for?  Should I move it to `packages/stylesheet`?


necolas:
You can import `StyleSheet` from `react-native-web` for now. With the Babel plugin it will bundle only what `StyleSheet` needs.

giuseppeg:
Much easy to do that! I'll close this PR then.

PR #748, opened 2018-01-03
RFC: react-native-web module organization

necolas:
The patch reorganizes the top-level module division of the `react-native-web` project.

Previously, the package's exported modules were found in:

```
apis/*/index.js
components/*/index.js
components/*/*.js
modules/*/index.js
propTypes/*.js
```

This inconsistent location of modules complicated the Babel plugin used to rewrite imports. Plus, while working on the project I've found that I need faster answers to questions like "is this module exported by the package?", "where should this new module go?", "which of these imports are internals?".

Now, each part of the public API is found in:

```
exports/*/index.js
```

And anything not directly part of the public API is found in:

```
modules/*/index.js
vendor/*/index.js
```

When editing a file in `exports`, any `import` path that includes `modules` or `vendor` clearly indicates an internal dependency. 

While making this change it highlighted a few problems with relationships between certain modules. There are components that only depend on the public API and could eventually be moved into user-land (or a different package). There are a few components whose only internal dependency is on `StyleSheetPropType`, which I also would have liked access to when making a component library from Twitter Lite. Sometimes you want to expand (as `Picker` does) or limit the style properties for your component. This suggests to me that `StyleSheetPropType` and `StylePropTypes` could be added to the React Native API (perhaps hanging off `StyleSheet`?). Internally, `flattenStyle` is used in a couple of modules and might be more conveniently moved from `exports/StyleSheet/flattenStyle` to `modules/flattenStyle`.


packages/babel-plugin-react-native-web/src/index.js
```diff
@@ -1,89 +1,5 @@
...
-    case 'ViewPropTypes': {
-      return `${root}/components/View/${importName}`;
-    }
-
-    default:
-      return;
-  }
-};
+const getDistLocation = importName =>
+  importName ? `react-native-web/dist/exports/${importName}` : undefined;
```
necolas: The babel plugin is much simpler

packages/react-native-web/src/exports/ActivityIndicator/index.js
```diff
@@ -11,9 +11,9 @@
  */
 
 import applyNativeMethods from '../../modules/applyNativeMethods';
```
necolas: `../../modules` makes it obvious which dependencies are internal

packages/react-native-web/src/exports/Button/index.js
```diff
@@ -10,9 +10,9 @@
  * @flow
  */
 
-import ColorPropType from '../../propTypes/ColorPropType';
-import StyleSheet from '../../apis/StyleSheet';
-import TouchableOpacity from '../Touchable/TouchableOpacity';
+import ColorPropType from '../ColorPropType';
+import StyleSheet from '../StyleSheet';
+import TouchableOpacity from '../TouchableOpacity';
```
necolas: More obvious that `Button` is made of simpler primitives and not part of the "core"

packages/react-native-web/src/exports/CheckBox/index.js
```diff
@@ -11,12 +11,12 @@
  */
 
 import applyNativeMethods from '../../modules/applyNativeMethods';
```
necolas: This is the only internal dependency for `CheckBox`. In user-land, a `View` ref could be used to add these methods instead.

packages/react-native-web/src/exports/Picker/index.js
```diff
@@ -12,17 +12,19 @@
...
+import ColorPropType from '../ColorPropType';
+import createElement from '../createElement';
 import PickerItemPropType from './PickerItemPropType';
-import PickerStylePropTypes from './PickerStylePropTypes';
-import StyleSheetPropType from '../../propTypes/StyleSheetPropType';
-import StyleSheet from '../../apis/StyleSheet';
-import TextStylePropTypes from '../Text/TextStylePropTypes';
+import StyleSheetPropType from '../../modules/StyleSheetPropType';
+import StyleSheet from '../StyleSheet';
+import TextPropTypes from '../Text/TextPropTypes';
```
necolas: `TextPropTypes` should be exported by the package (as `ViewPropTypes` is)

packages/react-native-web/src/exports/Picker/index.js
```diff
@@ -12,17 +12,19 @@
...
+import ViewPropTypes from '../ViewPropTypes';
 import { arrayOf, bool, func, number, oneOfType, string } from 'prop-types';
 
-const pickerStyleType = StyleSheetPropType(PickerStylePropTypes);
-const itemStylePropType = StyleSheetPropType(TextStylePropTypes);
+const pickerStyleType = StyleSheetPropType({
+  ...ViewPropTypes.style,
```
necolas: Looks like a bug, missing import of `PickerStylePropTypes` but should be using that

packages/react-native-web/src/vendor/Animated/createAnimatedComponent.js
```diff
@@ -15,7 +15,7 @@
 const {AnimatedEvent} = require('./AnimatedEvent');
 const AnimatedProps = require('./nodes/AnimatedProps');
 const React = require('react');
-const ViewStylePropTypes = require('../../components/View/ViewStylePropTypes').default;
+const ViewStylePropTypes = require('../../exports/View/ViewStylePropTypes').default;
```
necolas: Another hint that `ViewStylePropTypes` (or `StylePropTypes`) should be added to the exported API, allowing modules like `Animated` and `Picker` to be built upon a smaller core API.

packages/react-native-web/src/vendor/Animated/nodes/AnimatedStyle.js
```diff
@@ -17,7 +17,7 @@ const AnimatedTransform = require('./AnimatedTransform');
 const AnimatedWithChildren = require('./AnimatedWithChildren');
 const NativeAnimatedHelper = require('../NativeAnimatedHelper');
 
-const flattenStyle = require('../../../apis/StyleSheet/flattenStyle').default;
+const flattenStyle = require('../../../exports/StyleSheet/flattenStyle').default;
```
necolas: Could be `StyleSheet.flatten` instead, if it were depending on the exported API

paularmstrong:
> There are a few components whose only internal dependency is on StyleSheetPropType, which I also would have liked access to when making a component library from Twitter Lite.

I was actually looking for this recently for another project and was upset that I couldn't get at it.

Overall, project structure makes sense. I still prefer having `components` and `modules` split out, just for a little extra organization.

PR #751, opened 2018-01-06
Add missing babel plugin dependency

cdosborn:
The transform-runtime plugin is used in the sample webpack config, but not suggested as a package to install.


necolas:
Thanks, I decided to remove mention of that plugin for now - b78206d

PR #754, opened 2018-01-08
Add missing lint:cmd

MoOx:
lint:cmd is used in lint-staged config. No failure if everything is already up to date, only happen when you made change that is updated by prettier.


MoOx:
Not sure why this is failing on jest :/

necolas:
It just timed out. I've noticed that happen a couple of times

necolas:
What I'll do is remove `lint:cmd` and use `eslint` in your other patch adding husky.

MoOx:
lgtm :)

PR #757, opened 2018-01-08
Use prettier config instead of CLI args

MoOx:
This allow IDE plugins that rely on prettier config (introduced in [1.6.0](https://github.com/prettier/prettier/pull/2434)) to detect prettier and run it automatically with the correct config!

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "docs:release": "cd website && yarn release",
     "flow": "flow",
     "fmt": "find packages scripts types website -name '*.js' | grep -v -E '(node_modules|dist|vendor)' | xargs yarn fmt:cmd",
-    "fmt:cmd": "prettier --print-width=100 --single-quote --write",
+    "fmt:cmd": "prettier --write",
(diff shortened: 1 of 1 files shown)


necolas:
Do they all work with package.json or would a prettierrc be better?

paularmstrong:
I believe SublimeText JSPrettier requires `.prettierrc` or a manually set file location (that is not compatible with `package.json` format) https://github.com/jonlabelle/SublimeJsPrettier

MoOx:
prettier rely on [cosmiconfig](https://github.com/davidtheclark/cosmiconfig) which support multiple format (depending on the options when instantiated). I guess it should not be a big deal at the end if plugins use prettier API instead of custom stuff to parse option + CLI. I used package.json to avoid the .*rc fatigue :)

MoOx:
It seems SublimeJsPrettier is taking prettier configuration files: https://github.com/jonlabelle/SublimeJsPrettier#prettier-configuration-files

PR #762, opened 2018-01-10
[fix] KeyboardAvoidingView missing this binding

johakr:
Fixes `Uncaught TypeError: Cannot set property 'frame' of undefined`


jsp3536:
I am seeing the same problem. It was causing the rest of my app not to function correctly.

necolas:
Thank you!

PR #772, opened 2018-01-17
Include Object.assign polyfill recommendation

paularmstrong:
**Problem:** Since react-native-web@0.3.0, if you target older browsers, you will need to polyfill `Object.assign`. This was previously not a problem.

**Solution:** Add to the list of recommended polyfills in the getting started doc.


necolas:
This is already fixed but I haven't pushed the commit

PR #305, opened 2016-12-30
added alert api

vaukalak:
**This patch solves the following problem**

When developing mobile application, there a lot of cases when `Alert` api is used. Anyway it's not available in `react-native-web` and developer have to rewrite a lot of code. This patch solves the issue.

**Test plan**

Added to storie book.

**This pull request**

- [x] includes documentation
- [ ] includes tests
- [x] includes an interactive example
- [ ] includes screenshots/videos

PR #773, opened 2018-01-17
Add Alert API

JulianMayorga:
alert

# Motivation

React router 4 on a React Native Web app fails because [it needs `Alert` to work](https://github.com/ReactTraining/react-router/blob/master/packages/react-router-native/NativeRouter.js#L4).

@necolas please let me know if there's anything you'd like me to improve.

Thanks to @vaukalak for making #305, which really helped me.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -27,6 +27,10 @@
     "window": false,
     // Flow global types
     "HTMLInputElement": false,
+    "HTMLElement": false,
+    "HTMLHeadingElement": false,
+    "HTMLButtonElement": false,
+    "Text": false,
     "ReactClass": false,
     "ReactComponent": false,
     "ReactElement": false,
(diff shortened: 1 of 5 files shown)


necolas:
i don't think you're meant to be using the native router in a web app

necolas:
I'm not going to merge this patch as it is not authored in the way I believe the Alert API needs to be implemented (e.g., using Portals). There's also another PR for this that is closer to what I would like to see. Thanks

JulianMayorga:
@necolas you're right that the native router is not meant to be used in a web app. I do think it would make it easier for beginners to copy-paste the examples in react router native docs (like [this one](https://reacttraining.com/react-router/native/example/Basic)) into react native web apps.

JulianMayorga:
I understand. I agree that portals would be a better way to implement Alert. Thanks for such a great framework! I appreciate all the work put into it.

PR #779, opened 2018-01-18
Add react-fela to Benchmarks site

javivelasco:
Thanks for the project @necolas !

This PR adds to that amazing Benchmarks site [React Fela](https://github.com/rofrischmann/fela) by @rofrischmann. It is not that well known but when it comes to performance and API it is pretty amazing. I've followed the same pattern as in the rest of the libraries and it works like a charm. Here is an screenshots of some results I've got:

results-final

Let me know any changes you want me to make!
Thanks!


necolas:
Thanks, merged. I recorded the results for `react-fela` that I get on my machine, which are not as good as what you saw.

necolas:
FYI @rofrischmann and @rtsao, I noticed that class names generated by your libraries collide with each other. Try moving between the test previews for both libraries and they will start to break each other: https://necolas.github.io/react-native-web/benchmarks

rofrischmann:
Yeah, that's true. They're not meant to be used together and use a simple counting mechanism to determine the next class name => a, b, c, ..., aa, ab, and so on.

I could add a PR with a selector prefix to avoid that.

kompot:
Out of curiosity - why was version downgraded to 5.x from 6.x, @javivelasco? I guess latest stable version might be more relevant.

javivelasco:
@kompot I think @rofrischmann can tell you better but in the latest version there was a change that affected performance so it wouldn't be the "real benchmark" unless we jump into a version where that piece of code is missing.

rofrischmann:
Yeah exactly. I will bump the version as soon as the rehydration technique is optimised for cross use cases like this bench. Right now it wouldn't yield realistic results.

necolas:
I already updated and didn't see any difference in results

rofrischmann:
Huuuh? Strange. Which browser do you use by the way? I get quite different perf distribution (better for atomic approaches in general) on Safari.

necolas:
If there's a problem with rehydration it's not going to show up here.

Issue #775, opened 2018-01-18
babel-plugin-react-native-web badcase


hushicai:
It seems that `babel-plugin-react-native-web` can not cover this case:

```javascript
var RNDeviceInfo = require('react-native').NativeModules.RNDeviceInfo;
```

Environment:
  OS: macOS Sierra 10.12.6
  Node: 8.9.0
  Yarn: 1.3.2
  npm: 5.5.1
  Watchman: 4.7.0
  Xcode: Xcode 9.2 Build version 9C40b
  Android Studio: 3.0 AI-171.4443003

Packages: (wanted => installed)
  react: ^16.2.0 => 16.2.0
  react-native: ^0.52.0 => 0.52.0
  react-native-web:^0.3.1 => 0.3.1

webpack-dev-server errors:

> ERROR in ./node_modules/react-native/Libraries/react-native/react-native-implementation.js
Module not found: Error: Can't resolve 'takeSnapshot' in '/Users/hushicai/data/work/rnmanager/node_modules/react-native/Libraries/react-native'
 @ ./node_modules/react-native/Libraries/react-native/react-native-implementation.js 115:30-53
 @ ./node_modules/react-native-device-info/deviceinfo.js
 @ ./app/pages/app.js
 @ ./app/root.js
 @ ./index.js
 @ multi (webpack)-dev-server/client?http://localhost:8080 webpack/hot/dev-server ./index.js

necolas:
That's right, it can't convert those kind of require statements

hushicai:
so can I use webpack `resolve.alias` to fix this case?

necolas:
You should really move the use of `RNDeviceInfo` out of the file that is being loaded for web. Native modules aren't there on web.

hushicai:
Is there a recommended way to patch package that is not supported by official `react-native-web`?

necolas:
Not at the moment

hushicai:
How about this case:

```javascript
var ReactNative = require('react-native');
```

`babel-plugin-react-native-web` also  seems not handle it.

necolas:
What do you mean? There's a unit test for that https://github.com/necolas/react-native-web/blob/master/packages/babel-plugin-react-native-web/src/__tests__/index-test.js

hushicai:
This is an `VariableDeclaration`:

```javascript
var ReactNative = require('react-native');
```

but `babel-plugin-react-native-web` can not convert it to `react-native-web/dist/index`.

AST:

[...]

PR #781, opened 2018-01-19
fixed babel-plugin badcase #775

hushicai:
convert 
`const ReactNative = require('react-native'); `
 to 
`const ReactNative = require('react-native-web/dist/index');`


hushicai:
In this way, `react-native-web` will also convert to `react-native-web/dist/index`.

necolas:
Yes converting to `react-native-web/dist/index` is correct. Thank you for this great fix

necolas:
This has been merged. I will make a patch release (0.3.3) with this fix tomorrow. Thanks again for finding and fixing this issue

PR #782, opened 2018-01-19
Benckmarks: add styled-jsx

giuseppeg:


packages/benchmarks/src/implementations/styled-jsx/Box.js
```diff
@@ -0,0 +1,50 @@
...
+    case 5:
+      return '#E0245E';
+    default:
+      return 'transparent';
+  }
+};
+
+const Box = props => {
+  const { className, children: styles } = (
+    <scope>
```
giuseppeg: styled-jsx doesn't allow to style components (`View`) or pass styles down, the only way to style descendants is this or `:global()`. This workaround might add overhead. Would you accept an implementation of the benchmark that doesn't make use of a `View` component?

packages/benchmarks/src/implementations/styled-jsx/Box.js
```diff
@@ -0,0 +1,46 @@
...
+    case 5:
+      return '#E0245E';
+    default:
+      return 'transparent';
+  }
+};
+
+const Box = props => {
+  const { className, children: styles } = (
+    <scope className={classnames('Box', props.fixed && 'fixed')}>
```
necolas: What is `scope`?

giuseppeg: It is a dummy element, it could be any lowercase tag name e.g. `div`. This is a hack-workaround that a user came up with.

Because of the way how styled-jsx rewrites JSX, `scope` will get the scoped `className`: something like`jsx-someHash`. At runtime I extract it from the React Element together with `children` which are the actual styles or, to be precise, the styled-jsx React Component which takes care of [un]mounting and updating the actual CSS.

```jsx
const { props } = (
  <scope className="foo">
      <style jsx>{`.foo { color: red }`}</style>
  </scope>
)
```

is transpiled to

```jsx
const { props } = (
  <scope className="foo jsx-123">
      <_JSXStyle styleId={'123'} css={`.foo.jsx-123 { color: red }`} />
  </scope>
)
```

And finally to `React.createElement` calls.

giuseppeg: It is not bullet proof since passing down `foo` is not "safe", but imho it is ok when dealing with leaf elements.

necolas: Thanks for the explanation

necolas:
This has been merged. Interesting results

giuseppeg:
Thanks! Not too bad, I think that the benchmark code can be optimized to perform a bit better, but I will use the benchmark to work on improving the library speed.

Issue #791, opened 2018-02-01
Add `onKeyUp` and` onKeyDown` support?


alexandergunnarson:
React Native only supports `onKeyPress` in `TextInput`s (and by extension probably everywhere else, though I haven't looked thoroughly enough). This is fine for devices that don't have keyboards (e.g. phones), but not necessarily good for devices that do. For instance:

1) I can't `.preventDefault` on arrow keys (or other non-printable keys that cause default behavior to happen, though those are less important to me) with `onKeyPress`; for it to actually override default behavior, that needs to happen in an `onKeyDown` handler. See https://github.com/necolas/react-native-web/pull/540 for related use cases.
  - a) For instance, I'm trying to create a spreadsheet app in which the user should be able to navigate easily with arrow keys, but currently the up arrow key causes the selection to go to the beginning of the input and the down arrow key causes it to go to the end; only after this does the focus move to the cell above or below. Only a `.preventDefault` in `onKeyDown` can remedy this.
2) Sometimes a user should be able to hold down a key without releasing it, and still have it trigger an event. Use cases include: 
  - b) A modal or other UI element which appears only as long as a user holds down a particular key
  - c) When a user should be able to hold a printable key (e.g. a space) down in an input and have it continue to append

`onKeyPress` is fine for most use cases, but as you can see, it is insufficient for some. I propose two possible solutions:

- Solution 1) That `TextInput`s (and perhaps other components, but for now at least that one) allow web-specific custom `onKeyDown` and/or `onKeyUp` handlers and ensure that `onKeyPress` is as close to the React Native behavior as possible (as it currently is, i think). I know that `onKeyPress` interacts with both of these handlers to an extent and I'm not sure what the right behavior is, but it seems that the `onKeyDown` and `onKeyUp` handlers, if present, should always fire when the DOM fires their respective events.
- Solution 2) (Insufficient for case 2a) but sufficient for the others) Edit the following line to add support for other non-printable keys (such as at least the arrow keys) that, like Backspace and Tab, do not fire `keyPress` events:
https://github.com/necolas/react-native-web/blob/31d428a649ecf268628d6f6622a99f51e7344b1b/packages/react-native-web/src/exports/TextInput/index.js#L304

necolas:
I'm finding this a bit hard to follow.

> I can't `preventDefault` on "Enter" or "Tab" with `onKeyPress`; for it to actually override default behavior, that needs to happen in an `onKeyDown` handler

[...]

PR #792, opened 2018-02-01
Add `onKeyPress` support for arrow keys

alexandergunnarson:
Resolves https://github.com/necolas/react-native-web/issues/791.

### DIFF
--- a/packages/react-native-web/src/exports/TextInput/index.js
+++ b/packages/react-native-web/src/exports/TextInput/index.js
@@ -104,6 +104,7 @@ class TextInput extends Component<*> {
     onSelectionChange: func,
     onSubmitEditing: func,
     placeholder: string,
+    placeholderTextColor: ColorPropType,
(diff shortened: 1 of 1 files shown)


necolas:
How does this resolve that issue?

This PR would also need tests and documentation updates. The commits would be flattened to please

alexandergunnarson:
Yes, the commits can be flattened, though wouldn't that happen in a Squash and Merge? Let me know if you want them flattened prior to a potential merge.

As for tests and documentation updates, that's fine — I'll add those in. I was waiting on adding a bunch of additional metadata to this PR until we discussed #791. This PR implements Solution 2 of #791 but perhaps you'd want a different one because of reasons mentioned there. Could you take a look at it and we can discuss there?

necolas:
I'd like them flattened please. I don't use the GitHub UI to manage PRs

alexandergunnarson:
Okay, no problem! Will do

PR #800, opened 2018-02-07
Provided reasonable default props: statusbar is hidden

### LINKED ISSUE #799: Fails with react-native-dropdownalert due to missing default props in StatusBar
Let's consider this a feature request. I'm preparing a PR for this.

The package react-native-dropdown alert relies on some properties of the StatusBar that aren't implemented in the current stub.
This resulted in an error.

Should just work.

I am using 0.3.0-rc4.

raarts:
closes #799.

### DIFF
--- a/packages/react-native-web/src/exports/StatusBar/index.js
+++ b/packages/react-native-web/src/exports/StatusBar/index.js
@@ -17,6 +17,23 @@ export default class StatusBar extends Component<*> {
   static setHidden() {}
   static setNetworkActivityIndicatorVisible() {}
   static setTranslucent() {}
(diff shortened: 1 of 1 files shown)


raarts:
Just a question: I'm writing an app built on this package. How do I refer that app's package.json to this branch in my fork? I tried:  `"react-native-web": "raarts/react-native-web#799-missing-default-props-in-StatusBar",` but of course it can't find the `dist/` directory. Thanks.

necolas:
At this time I'd rather not include more unused implementation details in the stub just to work around a package unsafely accessing what is intended to be a private static on the class in RN

Issue #719, opened 2017-11-22
Touchable: right-click triggers onPress


MatthieuLemoine:
**Do you want to request a *feature* or report a *bug*?**

A bug.

**What is the current behavior?**

When I right-click on a Touchable, the onPress event is triggered.

Can be reproduced in the [Storybook](https://necolas.github.io/react-native-web/storybook/?selectedKind=Components&selectedStory=TouchableHighlight&full=0&down=0&left=1&panelRight=0).

**What is the expected behavior?**

When right-clicking in a browser, it should only open the contextual menu but not trigger a click on the underlying view.

While trying to reproduce, I saw that this issue is not present on [Twitter Lite](https://mobile.twitter.com). Either it doesn't use Touchable or there is a way to disable this behavior.

**Environment (include versions). Did this work in previous versions?**

* OS: macOS
* Device: Macbook Air
* Browser: Firefox 57 & Chrome 62
* React Native for Web (version): 0.1.10
* React (version): 16.0.0

vaukalak:
@necolas what is expected behavior here?

necolas:
As described in the OP

PR #804, opened 2018-02-08
Fix for issue #719 - Touchable: right-click triggers onPress

rxrossi:
For components based on Touchable, in the conditions:
  - Secondary mouse click alone
  - Secondary mouse click while primary is active
  - Dragging secondary mouse click on touchable area
  - Dragging secondary mouse click from touchable area to outside
The following events are no longer triggered:
  - onPress
  - onLongPress
  - onPressIn
  - onPressOut


necolas:
Thanks. I've made a simpler fix to the responder event system that I think fixes this issue, and prevents right-clicks from triggering any events in the responder system.

Issue #731, opened 2017-12-07
Touchable: onPress is triggered when scrolling at the top of a ScrollView


wjielim:
**Do you want to request a *feature* or report a *bug*?**
Bug

**What is the current behavior?**
TouchableOpacity inside a ScrollView will be triggered if one try to scroll up while the ScrollView is already at its top. See the below gif for better understanding:

![scrollview touchable issue](https://user-images.githubusercontent.com/34246445/33724353-2dc2e4de-dbaa-11e7-90d8-1f16ab487ed6.gif)

This also happens to horizontal ScrollView. 

**If the current behavior is a bug, please provide the steps to reproduce and
if a minimal demo of the problem via Glitch or similar (template:
https://glitch.com/edit/#!/react-native-web-playground).**

copy paste this code:

```
<ScrollView style={{backgroundColor: '#f1f2f3', flex: 1, height: Dimensions.get('window').height, width: Dimensions.get('window').width}}>
{
          (() => {for (let i=0; i < 10; i++){
            buttons.push(
              <TouchableOpacity onPress={() => {alert(`button ${i} Pressed!`)}}
                                style={{width: 300, height: 150, backgroundColor: 'blue', color: 'white', margin: 10}}>
                <Text>Button {i}</Text>
              </TouchableOpacity>
            )
          }})()
}

{buttons}
</ScrollView>
```

**What is the expected behavior?**

TouchableOpacity should not trigger onPress when it's inside a ScrollView (with height set to screen/window height) when scrolling under any circumstances.

**Environment (include versions). Did this work in previous versions?**

* OS: Windows
* Device: Microsoft Surface Pro 3
* Browser: Google Chrome v62.0.3202.94 (Official Build) (64-bit)
* React Native for Web (version): 0.1.16
* React (version): 16.2.0

aizigao:
same problem, mark

jondkoon:
I made a glitch to try and repro this and it seems to be working as expected when using a `ScrollView`:

View: https://giant-planet.glitch.me/
Edit: https://glitch.com/edit/#!/giant-planet?path=src/App/index.js:1:0

If you replace the `ScrollView` with a `View` then it fails as described

necolas:
I can't reproduce this issue either. Perhaps it is Windows-only, but that seems unlikely given the browser is Chrome.

aizigao:
[...]

PR #808, opened 2018-02-09
Undesirable onPress on mobile browsers scrolling fix

rxrossi:
Currently canceling a event by moving outside from the Touchable area would not prevent the onPress event of being triggered
This would affect users clicking on a Touchable by accident while scrolling up or down on mobile browsers

It is probably a fix for the issue #731

### DIFF
--- a/packages/react-native-web/src/exports/Touchable/index.js
+++ b/packages/react-native-web/src/exports/Touchable/index.js
@@ -386,6 +386,7 @@ const TouchableMixin = {
    * Place as callback for a DOM element's `onResponderGrant` event.
    */
   touchableHandleResponderGrant: function(e: Event) {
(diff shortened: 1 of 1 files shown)


necolas:
> Canceling a event by moving outside from the Touchable area would not prevent the onPress event of being triggered

Where are you seeing that?

rxrossi:
Hi, the current behavior is the the same using Chrome for desktop simulating a device or on a actual device: https://imgur.com/a/wgyr9
This code changes the behavior to this: https://imgur.com/a/hgik2

necolas:
I see, this only occurs when the Touchables are not contained within a `ScrollView` (i.e., you're using browser body scrolling).

necolas:
I don't think this PR works

PR #811, opened 2018-02-12
fix of image not loaded when previous source URI is not defined

JakubRimal:
I think that this was caused by copy-pasting from the constructor. This mistake led to image not showing when the source URI in the previous state was not defined.


necolas:
Well spotted. I added a unit test and merged. Thanks!

PR #812, opened 2018-02-12
[add] StyleSheet and I18nManager support for start/end properties and RTL without left/right flipping

necolas:
Add support for direction-independent properties and values to enable better internationalized UI. The start/end variants are automatically resolved to match the global writing direction, as defined by `I18nManager.isRTL`. Start/End take precedence over Left/Right.

Adds support for the following properties:

+ `borderTop{End,Start}Radius`
+ `borderBottom{End,Start}Radius`
+ `border{End,Start}Color`
+ `border{End,Start}Style`
+ `border{End,Start}Width`
+ `end`
+ `margin{End,Start}`
+ `padding{End,Start}`
+  `start`

And values:

+ `clear: "end" | "start"`
+ `float: "end" | "start"`
+ `textAlign: "end" | "start"`

I18nManager also supports `doLeftAndRightSwapInRTL` and `swapLeftAndRightInRTL` to query and control the BiDi-flipping of left/right properties and values. For example, you may choose to use `end`/`start` for positioning that flips with writing direction, and then disable `left`/`right` swapping in RTL so that `left` will always be `left`.

### DIFF
--- a/packages/react-native-web/src/exports/I18nManager/__tests__/index-test.js
+++ b/packages/react-native-web/src/exports/I18nManager/__tests__/index-test.js
@@ -2,44 +2,100 @@
 
 import I18nManager from '..';
 
+const getDocumentDir = () => document.documentElement.getAttribute('dir');
+
 describe('apis/I18nManager', () => {
-  describe('when RTL not enabled', () => {
+  describe('preferred language is LTR', () => {
     beforeEach(() => {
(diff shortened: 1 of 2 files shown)


necolas:
Merged to master: b75477637385e46037014cbec5b364f085821c3b, 92794cdc9f967f84ba6108568a6d60f17640a14a, f1fc2a9e37f194c991dd8e4c7a0f2185aefb7ecc

PR #820, opened 2018-02-14
[fix] Update the ImageBackground source link

dcalhoun:
Update the `ImageBackground` "view source" link URL to match the
location within repository. The previous URL returned a 404.


necolas:
Thanks, I just moved that example into its own directory instead.

Issue #830, opened 2018-02-18
SectionList support


raarts:
I'd like to implement SectionList support, and will open a PR for this.

necolas:
Duplicates https://github.com/necolas/react-native-web/issues/91

udfalkso:
@raarts Are you actively working on this?  I also have the same need and may work on it myself if not.  Thanks!

raarts:
@udfalkso I was sidetracked, but my need is still there. Be my guest to work on this. Buy the way, @necolas mentioned elsewhere already having done some groundwork for this component, you might want to look into that. The VirtualizedList component is rather complex.

udfalkso:
Ok.  I might be able to live with a ScrollView on web for the moment, but if I end up needing to head down this route I'll drop a note here that I'm proceeding.  Thanks

Sixzero:
As of current date, 'SectionList' shows me 'UnimplementedView' on web, while on mobile devices it is working well.
Are there a workaround for 'SectionList' elements, or should we make our own implementation of this inherited from FlatList?

necolas:
SectionList is available (as you can see in the examples linked to in the readme), just don't use the legacy implementation

PR #659, opened 2017-09-27
Port over VirtualizedList from react-native (Updated to master)

rmevans9:
**This patch solves the following problem**

This is #474 just updated to master for easy merging. I also updated the story to match the new storybook setup.

**Test Plan**
Run it in storybook and ensure it works correctly

**This pull request**

- [ ] includes documentation
- [ ] includes tests
- [ ] includes benchmark reports
- [x] includes an interactive example
- [ ] includes screenshots/videos

PR #831, opened 2018-02-18
Initial SectionList support, works with the Expo tabs template

raarts:
Addresses #830 .

I'd really like some input on this. since some issues are remaining:

* I get a few of `Warning: React does not recognize the 'xxxxx' prop on a DOM element.`, and I could use some pointers on how to fix this. My guess is it has to do with setNativeProps()
* There's no doc yet, but why not point to the ReactNative page?
* Although this PR now works with the Expo tabs template, I'm deeply worried that it's fundamentally broken, since I [removed a call](https://github.com/necolas/react-native-web/pull/831/files#diff-5d3b6006279b241304e519d4460980fbR369) to ReactNative() which points to a huge body of code which even seems to tie into the bus between JS and Native code. I spent some time already implementing this PR, and I don't want to spend more only to be told it's going to be a huge undertaking.
* I haven't use flow, and CI marks a lot of tests are failing, but before I embark on fixing everything, I'd really like some feedback.


necolas:
Hi this patch is currently too big and makes too many changes for me to effectively review

raarts:
Alright. So it needs to be split up. The best route for that I guess would be to build on #659, I reached out to @rmevans9 and he said he's in no hurry to do anything with it, so I am willing to assume maintenance of that part if that helps. Please advise.
This issue is a blocker in my startup so I am motivated to make this work.

necolas:
You can unblock yourself by using webpack to alias the import path to your local hack, or use a fork

raarts:
Sure I'm doing that but I don't like temporary solutions. Any input on the rest of my comment?

necolas:
I've been working my way through underlying functionality that's needed to get the lists working without significant modifications. None of the past list PRs have worked or been debugged by the author, which is why they weren't merged.

PR #838, opened 2018-02-27
[fix] Import react-art only client side to avoid crash

anonkey:
Little fix to avoid crash on server side rendering

### DIFF
--- a/packages/react-native-web/src/exports/ART/index.js
+++ b/packages/react-native-web/src/exports/ART/index.js
@@ -1,11 +1,4 @@
 /**
- * @Author: Theo Seguier <key>
- * @Date:   2018-02-27T10:16:16-08:00
- * @Last modified by:   key
- * @Last modified time: 2018-02-27T10:30:12-08:00
- */
-
-/**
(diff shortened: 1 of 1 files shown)


necolas:
This isn't a reliable approach. There's a fix upstream that I'm hoping will make it into the 16.3 react releases

koulmomo:
@anonkey you can avoid the crash by using https://www.npmjs.com/package/babel-plugin-react-native-web along with making sure you never import all of React Native and only using named imports instead.

```js
// never do this
import ReactNative from 'react-native';

// do this instead
import {View, Text} from 'react-native';
```

The downside is that you will have to run transpiled code on the server.

necolas:
I don't know when the next react release will fix this. So actually I'll reopen this and consider a temporary fix. May also make 'react-art' a peer dependency like the other react packages

anonkey:
@koulmomo Thanks a lot.
@necolas  Yes i didn't think about it, it seem a good idea.

necolas:
Art has been updated and this issue should be fixed

PR #850, opened 2018-03-04
Status bar

piranna:
Implement `StatusBar` component instead of being a noop. This allow to set backgroundColor as navigation bar color in Chrome for Android, and all the other props and static functions with their iOS equivalents by injecting and updating their `meta` tags.

### DIFF
--- a/packages/react-native-web/src/exports/StatusBar/index.js
+++ b/packages/react-native-web/src/exports/StatusBar/index.js
@@ -46,7 +46,11 @@ function setAppleMobileWebAppStatusBarStyle() {
 
 
 export default class StatusBar extends Component<*> {
-  static currentHeight = 0
+  static get currentHeight() {
+    const {availHeight, height} = screen
+
+    return height - availHeight
+  }
 
 
   static setBackgroundColor(color, animated) {
--- a/packages/react-native-web/src/exports/StatusBar/index.js
+++ b/packages/react-native-web/src/exports/StatusBar/index.js
@@ -79,6 +79,16 @@ export default class StatusBar extends Component<*> {
 
 
   render() {
+    const {animated, backgroundColor, barStyle, hidden,
+      networkActivityIndicatorVisible, showHideTransition,
+      translucent} = this.props
+
+    if(barStyle) StatusBar.setBarStyle(barStyle, animated)
+    if(backgroundColor) StatusBar.setBackgroundColor(backgroundColor, animated)
+    if(hidden) StatusBar.setHidden(hidden, animated)
+    if(networkActivityIndicatorVisible) StatusBar.setNetworkActivityIndicatorVisible(networkActivityIndicatorVisible)
+    if(translucent) StatusBar.setTranslucent(translucent)
+
     return null;
   }
 }
--- a/packages/react-native-web/src/exports/StatusBar/index.js
+++ b/packages/react-native-web/src/exports/StatusBar/index.js
@@ -7,20 +7,21 @@
  * @providesModule StatusBar
  * @flow
  */
+import { canUseDOM } from 'fbjs/lib/ExecutionEnvironment';
 import { Component } from 'react';
 
 
-const {head} = document
+const {head} = canUseDOM && document
 
 let _barStyle = 'default'
 let _hidden = false
 let _translucent = false
 
 
-function getMetaTag(attrName) {
-  const htmlCollection = head.getElementsByTagName('meta')
+function setMetaTag(attrName, content) {
+  if(!canUseDOM) return
 
-  let tag = Array.from(htmlCollection).filter(({name}) => name === attrName)[0]
+  let tag = head.querySelector(`meta[name=${attrName}]`)
 
   if(!tag) {
     tag = document.createElement('meta')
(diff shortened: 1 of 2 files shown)


packages/react-native-web/src/exports/StatusBar/index.js
```diff
@@ -7,16 +7,92 @@
  * @providesModule StatusBar
  * @flow
  */
-
 import { Component } from 'react';
 
+
+const {head} = document
```
necolas: This needs to be able to run in a server-side context too. Use `canUseDOM` guard as done elsewhere

packages/react-native-web/src/exports/StatusBar/index.js
```diff
@@ -7,16 +7,92 @@
...
+
+let _barStyle = 'default'
+let _hidden = false
+let _translucent = false
+
+
+function getMetaTag(attrName) {
+  const htmlCollection = head.getElementsByTagName('meta')
```
necolas: No need to collect all the meta tags when all you need is something like 

```
document.querySelector(`meta[name=${attrName}]`)
```

packages/react-native-web/src/exports/StatusBar/index.js
```diff
@@ -7,16 +7,92 @@
...
+    tag.name = attrName
+
+    head.appendChild(tag)
+  }
+
+  return tag
+}
+
+function setAppleMobileWebAppCapable() {
+  getMetaTag('apple-mobile-web-app-capable').content =
```
necolas: Setting `apple-mobile-web-app-capable` is problematic (it's basically a broken feature), which is why most web apps do not set it.

packages/react-native-web/src/exports/StatusBar/index.js
```diff
@@ -7,16 +7,92 @@
...
+    _translucent ? 'black-translucent' : _barStyle
+}
+
+
 export default class StatusBar extends Component<*> {
-  static setBackgroundColor() {}
-  static setBarStyle() {}
-  static setHidden() {}
+  static get currentHeight() {
+    const {availHeight, height} = screen
```
necolas: `window.screen`

packages/react-native-web/src/exports/StatusBar/index.js
```diff
@@ -7,16 +7,92 @@
...
+
+
+  static setBackgroundColor(color, animated) {
+    getMetaTag('theme-color').content = color
+  }
+
+  static setBarStyle(style, animated) {
+    if(style === 'light-content') style = 'black'
+
+    if(!['black', 'black-translucent', 'default'].includes(style)) return
```
necolas: `includes` is not polyfilled and it's not necessary to create an array to perform this check

packages/react-native-web/src/exports/StatusBar/index.js
piranna: Didn't know of that, fixed.

packages/react-native-web/src/exports/StatusBar/index.js
piranna: Good trick, thank you! :-) Fixed.

packages/react-native-web/src/exports/StatusBar/index.js
piranna: I've used it because seems [Apple docs say it's needed](https://developer.apple.com/library/content/documentation/AppleApplications/Reference/SafariHTMLRef/Articles/MetaTags.html) to enable the `apple-mobile-web-app-status-bar-style` meta tag and make it works. Do you know of any other alternative?

packages/react-native-web/src/exports/StatusBar/index.js
piranna: Done.

packages/react-native-web/src/exports/StatusBar/index.js
piranna: I'm used to use `includes()`, but it's ok for me to change that. Done.

necolas:
You need to run `yarn test` and fix the errors

piranna:
> You need to run yarn test and fix the errors

Fixed `Flow` errors. There are some remaining ones, but are related to other components.

piranna:
> I'm not sure that all of this is particularly useful. Perhaps only keep the theme color for apps not using a service worker / manifest or wanting to customize the tab bar color for mobile chrome.

I only needed the color theme just for aesthetics, since I knew that Chrome supported the change of the navigation bar it I though it could be a good replacement to the Android and iOS support to change status bar color. I tried to implement the other things just to don't have the API incomplete.


> The Apple web app stuff is broken in Safari

Lol...


> And the translucent setting in native is Android only, but here it's iOS Safari only. I think that can be removed.

Platform dependent options are noop for the ones that doesn't have an alternative, but if we can have some way to implement some behaviour available in just some of the platforms instead of all of them, I don't see why we shouldn't add it as far as the API is the same... In fact, `Picker` don't support `dialog` mode on iOS and [I have implemented it](https://github.com/QuantumBA/PickerDialog) in a module with the same API (sort of polyfill, if you want to see it that way), and I'm thinking to do a pull-request on ReactNative code to add that...

necolas:
Thanks for this, but at this stage I'm going to close it for the reasons previously mentioned.

piranna:
It's your project and your decission, but it's not fair. I've investigated about how to implement this and filled all the possible behaviour, and also fixed the problems that you showed up. Maybe it's not too much useful, but improve apps eyecandy and more important implement a missing API, I think this is enough reason to accept it.

necolas:
You didn't respond to requests for further changes, and unfortunately Safari's web app features currently have bugs that you can't do anything about. These settings are meant to be defined in an app manifest file on web now, that's the expectation for PWAs.

shamilovtim:
You've mentioned using a PWA manifest -- I assume that's for mobile safari and mobile chrome. What I'm not seeing is anything about an identical looking purely front end polyfill for desktop browsers. Philosophically, if we can make a button that looks like a button, or other widgets that look on web like they look on mobile, I don't see why we can't have a status bar that looks like a mobile statusbar, with the handful of customization that the mobile one has (mostly color). I don't know if this user's solution is correct, I just feel like this polyfill is missing.

necolas:
There's no value in creating a fake status bar for desktop. No apps want or need that. Apps should feel native to their host environment

PR #858, opened 2018-03-07
Add ability to provide render callback

slorber:
This render callback is needed for Gatsby integration. (https://github.com/gatsbyjs/gatsby/blob/021a460184be6b8f8e1d2b3d072baabbbe45544c/packages/gatsby/cache-dir/production-app.js#L185)

Not exactly sure what I'm doing so tell me if I did something wrong :)

### DIFF
--- a/packages/react-native-web/src/exports/AppRegistry/index.js
+++ b/packages/react-native-web/src/exports/AppRegistry/index.js
@@ -48,8 +48,8 @@ export default class AppRegistry {
     runnables[appKey] = {
       getApplication: ({ initialProps } = emptyObject) =>
         getApplication(getComponentFunc(), initialProps),
-      run: ({ initialProps = emptyObject, rootTag }) =>
-        renderApplication(getComponentFunc(), initialProps, rootTag)
+      run: ({ initialProps = emptyObject, rootTag, callback }) =>
+        renderApplication(getComponentFunc(), initialProps, rootTag, callback)
     };
     return appKey;
   }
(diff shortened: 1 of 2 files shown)


slorber:
Sure, will do that next week when I come back from holiday.

Thanks for the review

slorber:
Hi @necolas 

I tried to write a test for this callback but I'm not exactly sure how to do this.

All the other tests are currently using Enzyme which does not use any callback, and the following code is currently not tested as far as I've seen:

```js
const renderFn = process.env.NODE_ENV !== 'production' ? render : hydrate;

export default function renderApplication<Props: Object>(
  RootComponent: ComponentType<Props>,
  initialProps: Props,
  rootTag: any,
  callback?: () => void
) {
  invariant(rootTag, 'Expect to have a valid rootTag, instead got ', rootTag);

  renderFn(
    <AppContainer rootTag={rootTag}>
      <RootComponent {...initialProps} />
    </AppContainer>,
    rootTag,
    callback
  );
}
```

To test the callback addition, this `renderApplication` function should indeed be tested in the first place. I am ok to write such a test if you want. 

Do we agree that the test should not call the real ReactDOM render/hydrate functions and mock them? So basically the test will only verify that the arguments are dispatched correctly to real render/hydrate methods? This means I should then probably pass an extra argument `renderFn` to `renderApplication` with a default value to current `renderFn` so that I can mock it in the test and verify it is called with the callback as 3rd arg. This seems a bit overkill :)

I'll update the doc

slorber:
Hi @necolas 

I've updated the doc and resolved the conflicts of the PR related to the new `WrapperComponent`

Waiting for your feedback related to the `renderApplication` test.

Note that I wasn't able to run the storybook website. Do you have a contribution guide explaining that? Should I use Lerna or Yarn should be enough?

necolas:
Thanks, I'll take care of adding the test

slorber:
Thanks,  I'll upgrade the Gatsby plugin as soon as there's a release (ping me)

I tried `yarn website` but got errors, it was because I didn't know I had to compile first

PR #860, opened 2018-03-08
[fix] Picker default font styles

paularmstrong:
**Problem:** Picker was using `Times` as the font-style, because it was using `inherit`.

**Solution:** Same as applied for `TextInput` in b28cbbb37ece3476a40370d359e982b4ba563bbb, set `fontFamily` to `System`

### DIFF
--- a/packages/react-native-web/src/exports/Picker/__tests__/__snapshots__/index-test.js.snap
+++ b/packages/react-native-web/src/exports/Picker/__tests__/__snapshots__/index-test.js.snap
@@ -10,7 +10,7 @@ exports[`components/Picker prop "children" items 1`] = `
 
 exports[`components/Picker prop "children" renders items 1`] = `
 <select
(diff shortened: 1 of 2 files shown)


PeterKottas:
Hi Paul, just a quick question as it seems you understand what you did there ;) . What exactly will system mean in this context? I am changing the fontFamily for all text inputs via overriding default props which works fine for me. However, picker is not picking this up and I have the feeling that this PR might fix it. I am just not totally sure if I am right. Will this somehow enable me to change font family for picker? Ty

PeterKottas:
Ah sorry, I just figured it out. My problem was typescript typings are incorrectly forbidding one from passing fontFamily, color, or fontSize. That's why I assumed it's unsupported but when I forced the values in, it all started working. Also in the light of this, the PR makes much more sense now ;)

PeterKottas:
Weirdly I get an error when doing this:
```
   Warning: Failed prop type: Invalid props.style key `fontFamily` supplied to `Picker`.
```
But this rly has no longer anything in common with this PR so I'll open separate issue. I just feel there's definitely some weirdness around picker.

PR #867, opened 2018-03-16
ignore tvParallaxProperties prop

kgoggin:
Based off of #753, this ignores `tvParallaxProperties`, which only applies to AppleTV [per the RN docs](https://facebook.github.io/react-native/docs/touchablehighlight.html#tvparallaxproperties).

### DIFF
--- a/packages/react-native-web/src/exports/View/ViewPropTypes.js
+++ b/packages/react-native-web/src/exports/View/ViewPropTypes.js
@@ -12,7 +12,7 @@
 import EdgeInsetsPropType, { type EdgeInsetsProp } from '../EdgeInsetsPropType';
 import StyleSheetPropType from '../../modules/StyleSheetPropType';
 import ViewStylePropTypes from './ViewStylePropTypes';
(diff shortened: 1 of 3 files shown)


packages/react-native-web/src/exports/View/ViewPropTypes.js
```diff
@@ -74,7 +74,8 @@ export type ViewProps = {
   onMagicTap?: Function,
   removeClippedSubviews?: boolean,
   renderToHardwareTextureAndroid?: boolean,
-  shouldRasterizeIOS?: boolean
+  shouldRasterizeIOS?: boolean,
+  tvParallaxProperties?: any
```
necolas: This type is different to the one on L127

kgoggin: Sorry - I've never used flow before and thought that was the appropriate type in this case as we'd never actually use this prop in RNW so weren't too concerned what was in it? If you let me know what it should be instead I'd be happy to fix it and reopen. Thanks!

necolas:
No worries, I made the edit and merged, which is why this is closed. Thanks!

PR #868, opened 2018-03-18
Add Gatsby integration

slorber:


necolas:
This can be added after the following happens:

1) PR for `AppRegisty` is merged.
2) Your plugin doesn't use `getApplication` in the browser.
3) There is an official Gatsby example listed [here](https://github.com/gatsbyjs/gatsby/tree/master/examples), which is what RNW will link to.

slorber:
I've made a Gatsby example PR: https://github.com/gatsbyjs/gatsby/pull/4885

Waiting for 1) PR feedback or merge to tackle 2)

PR #870, opened 2018-03-19
Add webpack-cli to the required set of packages to get started

jsdario:
Small change to avoid webpack requesting the webpack-cli installed before trying to read webpack.config.js.


necolas:
Added in 8fb9a88ee60c98e8370797b9ae5d8f18f170747d

PR #891, opened 2018-04-12
Update Phenomic integration link to official example

MoOx:
I just pushed a minimal example https://github.com/phenomic/phenomic/tree/master/examples/react-native-web-app

[See full (but small) commit](https://github.com/phenomic/phenomic/commit/dd9d19f6e3a5dba60e7262311c8eb060d8661070)

This example use 0.5.0 and is tested (like all our examples, to ensure styles are correctly pre-rendered) to avoid regressions (in case of upgrade etc).

No plugin needed since Phenomic api is flexible enough :)

(I will add a README in this folder soon, but code is light enough to be understand without further explanations ^^)


necolas:
Thanks! Nice and simple :)

necolas:
e022d166dd9dbe7944ba4c7a64037c71c9fe7de4

PR #901, opened 2018-04-17
Fix onPress event accidentally triggered while scrolling the page

huangzuizui:
The touchable components are likely to be miss triggered while scrolling the page. We can fix this issues by judging by touch velocity to make a distinction between press or scroll;


necolas:
You mean scrolling the body? There's an issue open for that. I think if this were to be supported it would need to intersect with the responder system rather than a hack like this. So I'm going to close for now. Thanks

PR #904, opened 2018-04-19
Added LocationApi

adnan1naeem:
Added Location Api feature

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,7 @@
 {
   "private": true,
-  "version": "0.6.0",
-  "name": "react-native-web-monorepo",
+  "version": "0.1.0",
+  "name": "@blueeast/react-native-web-monorepo",
   "scripts": {
     "clean": "del ./packages/*/dist",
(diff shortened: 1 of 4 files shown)


necolas:
Hi. I don't know what this API. PRs should explain to the maintainer what their purpose is. This kind of looks like something taken from ReactXP and most of the patch is type interfaces and unwanted changes to package.json files etc. It's not something I will merge in this form

adnan1naeem:
Hi,
I want something like  location api in react_native_web.can u please
provide me that functionality for getting current location

necolas:
React Native implements the same geolocation API as the web https://developer.mozilla.org/en-US/docs/Web/API/Geolocation/getCurrentPosition

adnan1naeem:
I want Location Api support for mobile + web (both platforms)

MoOx:
Necolas just said that the RN api is ripped from the browser API. So you are not really supposed to need something more for the web, that what the browser already gives you.
If you have more detailed arguments, please share those by sharing the motivations.

PR #908, opened 2018-04-21
Prepare RNW for DOM specific ResponderEventPlugin

philipp-spiess:
This PR prepares RNW for #12629. To keep changes at a minimum, the added dependencies were upstreamed to `react-dom`. 

##  State of this PR

This PR is currently blocked by #12629 and will need a new release of `react-dom` after #12629 is merged to function properly. I've verified that everything still works by linking against a local build of `react` and `react-dom` and running the examples via `yarn website`.


necolas:
While working on this, did you notice that upgrading to React(DOM) 16.3 breaks the storybook? After I upgrade the project to use React 16.3 instead of 16.2, none of the touchables work and I see these warnings in the console:

```
Warning: Unknown event handler property `onResponderGrant`. It will be ignored.
```

philipp-spiess:
I'm super confused now. I was just testing my branch again (which is based on a commit that was created after the 1.3.2 commit) by placing the build artifacts into `node_modules` and it was working fine. 

I then updated to 1.3.2 stable (via yarn) and saw the issue you described. Strangely I can't get my version to run anymore now. 🤔

necolas:
OK it looks like maybe something unexpected changed in React or React DOM between 16.2 and 16.3

philipp-spiess:
Ok I think I had different versions for the `website` package. Now I wonder why it worked in the first place 🤔🤔I guess I need a bit more time to figure that out...

philipp-spiess:
Just brainstorming here but can it be that after the upgrade, the website (= storybook) runs on a different react-dom than the one we inject the responderplugin in?

necolas:
Possibly. I think the benchmark's "Run" button was also broken after the upgrade though

vladp:
I am running into this issue now too.   What was strange is that in my package.json I had react dependencies set to **"^16.3.2"**  and it was working, working, working.  Untill, yesterday when I needed to refresh my node_modules.  
I was chasing a change for a few hours, that  I, thought I made, that broke this side of my app, but then stumbled onto this bug report.
 I checked the node_modules/react/package.json.  And sure enough, 16.4 crawled into my "^16.3.2"...
I am thinking when published 16.4 there was some convention was broken for semver, so that 16.4 was made available even for "^16.3.2".
@philipp-spiess , may be that's why you did not observe a problem, and then after you reinstalled your node_modules - it showed up.
  
I cannot go back to 16.2  (and I cannot go forward to RNW 7.2+ (as my vector icons do not work in 7.2), so will stay at  RN "16.3.2"  RNW 0.6.1 for the time being.

necolas:
I’m sorry this broke things for you. React doesn't follow semver when making changes to unofficial APIs, like the ones exposed for React Native for Web.

vladp:
@necolas , thank you.  Understood.

necolas:
I was working on 0.8.0 when React 16.4 came out, but will pause that to make a patch release for React 16.4 and 0.7.x today.

necolas:
Thanks a lot for the help here. This fix is in 0.7.3

PR #909, opened 2018-04-21
Fix 2048 on desktop browsers

philipp-spiess:
One thing I found out while working on #908 this is that the 2048 game demo does not work on browsers since it listens for touch events only.

This can be solved by listening to both mouse and touch events - or perhaps there is a more RNW idiomatic way of solving this?


necolas:
Yeah this was just copy pasted from RN. Probably worth fixing

Issue #918, opened 2018-04-28
Hydrate method for development


codejamninja:
Can we switch the render method to hydrate in development mode?

Refer to issue #733

necolas:
See #745

necolas:
What's going to be deprecated?

English3000:
"Because React is backwards compatible, `render()` will continue to work when rendering on top of server-generated markup in React 16, but you should change those calls to `hydrate()` to quiet the warning and prepare your code for React 17."

See here: https://hackernoon.com/whats-new-with-server-side-rendering-in-react-16-9b0d78585d67

necolas:
`render` isn't being deprecated though. You still use it when not hydrating server-rendered markup. 

This issue doesn't explain what the problem is. Just that there's a warning in development if you're using SSR? I don't think there's much I can do about that, as either way is going to produce a warning for someone and dev-time SSR is the less common case.

codejamninja:
Ok, cool. Thanks for the explanation.

Issue #733, opened 2017-12-09
[React 16] hydrate method


jeans11:
**Do you want to request a *feature* or report a *bug*?**
_Feature_

**What is the current behavior?**
`AppRegistry.runApplication` call `ReactDom.render` method.

**What is the expected behavior?**
Call `ReactDom.hydrate` method when React 16 is used.

necolas:
Are there any side-effects from calling `hydrate` if the app wasn't SSR-ed?

MoOx:
Not from my experience. hydrate just tells React to "try to reuse existing markup as much as possible".

codejamninja:
I'm still getting the hydrate error, and I'm using the newest version of react-native-web.

MoOx:
hydrate is used in NODE_ENV production only https://github.com/necolas/react-native-web/blob/7a3a9a5c3ffdbf096150cc940ef542d96af0589e/packages/react-native-web/src/exports/AppRegistry/renderApplication.js#L18

codejamninja:
Ah thanks, why is that?

PR #919, opened 2018-04-28
Always hydrate

codejamninja:


codejamninja:
Fixes #918 
Also refer to #733

necolas:
I would like clear explanations for PRs. IIRC, hydrate will throw a warning in development if there's no SSR content to hydrate, which is why things are done this way.

bevacqua:
@necolas could `AppRegistry.runApplication` at least take a flag that indicate we always want to hydrate, e.g `ssr: true`?

> `react-dom.development.js:5585` Warning: render(): Calling ReactDOM.render() to hydrate server-rendered markup will stop working in React v17. Replace the ReactDOM.render() call with ReactDOM.hydrate() if you want React to attach to the server HTML.

bevacqua:
Also when you SSR, you get a console statement like `Running application "App" with appParams: …` which can be _huuuge_. Feels like this statement belongs on dev mode only

necolas:
> could AppRegistry.runApplication at least take a flag to indicate we always want to hydrate, e.g ssr: true

Why?

> Feels like this statement belongs on dev mode only

React Native always prints it. It's not related to SSR either as it's part of the client-side code. I think it's probably ok to make it dev only for web though

MrLoh:
@necolas React now throws the following Warning: `render(): Calling ReactDOM.render() to hydrate server-rendered markup will stop working in React v17. Replace the ReactDOM.render() call with ReactDOM.hydrate() if you want React to attach to the server HTML.`

So maybe adding a flag would be the best option after all?

codejamninja:
I've just been temporarily ignoring it with the `ignore-warnings` module.

https://www.npmjs.com/package/ignore-warnings

PR #935, opened 2018-05-07
Set Switch thumb position to absolute

maggo:
Thumb positioning breaks in RTL layout because `right: 0%` with relative positioning does nothing. Setting the position to absolute fixes it

### DIFF
--- a/packages/react-native-web/src/exports/Switch/index.js
+++ b/packages/react-native-web/src/exports/Switch/index.js
@@ -188,7 +188,7 @@ const styles = StyleSheet.create({
     backgroundColor: '#D5D5D5'
   },
   thumb: {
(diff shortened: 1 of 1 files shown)


necolas:
Looks fine to me here: https://harmonious-dart.glitch.me. The positioning changes to 100% when the switch is active, and that does get applied with position relative

maggo:
Ah didn't know that there is a test suite, thank you. Okay, the issue is that in my app the `align-self: flex-start` is not turned around automatically. I'll try to extract a test case.

maggo:
For anyone interested: The issue was that I had a wrapping <Text> that reset the direction of everything from RTL to LTR. Replaced it with a <View> and everything works fine.

Issue #802, opened 2018-02-07
PanResponder: onPanResponderGrant fired twice on touch devices


gut4:
**Do you want to request a *feature* or report a *bug*?**
bug
**What is the current behavior?**
On touch devices onPanResponderGrant fired twice. It happens on real devices and in chrome devTools emulator. 
https://dog-voyage.glitch.me/  (touch header -> chrome devTools device toolbar  -> see console)

**What is the expected behavior?**
fire once

**Environment (include versions). Did this work in previous versions?**

* OS: Android & iOS
* Device:
* Browser: Chrome
* React Native for Web (version): 0.4.0
* React (version): 16.2.0

PR #938, opened 2018-05-08
[change] ResponderEventPlugin filters browser emulated mouse events

### LINKED ISSUE #932: onValueChange not triggered on input based elements for touch events
bug
Elements like Switch, Input... doesn't fire `onValueChange` on mobile browsers and Chrome when switched to emulate touch events.

Not sure if this is needed but issue can be easily reproduced on simple Switch component.

Switch `onValueChange` on mobile browsers and Chrome is not called.

How to reproduce:
1. Create Switch component with onValueChange event handler
2. In Chrome switch to emulate touch events
3. Click on Switch

Switch `onValueChange` will be called

Yes it did. Not sure about exact version number, but version used following [createElement](https://raw.githubusercontent.com/mauron85/react-native-web/extended/src/modules/createElement/index.js) works.
I think problem is related to line: https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/createElement/index.js#L56

* OS: Ubuntu 16.04
* Device:
* Browser: Chrome 66
* React Native for Web (version):0.6.1
* React (version):16.3.2

necolas:
Browsers dispatch mouse events after touch events:
https://developer.mozilla.org/en-US/docs/Web/API/Touch_events/Supporting_both_TouchEvent_and_MouseEvent

There have been several attempts to avoid this behaviour affecting the
ResponderEvent system. The previous approach of cancelling the event in
the `onResponderRelease` event handler can end up cancelling other
events that are expected, e.g., `focus`.

Instead, this patch changes the `ResponderEventPlugin.extractEvents`
function to filter the mouse events that occur a short time after a
touch event. (It's assumed that people will not be clicking a mouse
within a few hundred ms of performing a touch.) This allows the
ResponderEvent system to function as expected and leaves other callbacks
to fire as they would be expected to in React DOM, i.e., both
'onTouchStart' and 'onMouseDown' will be called following a touch start.

Fix #932
Ref #802

### DIFF
--- a/packages/react-native-web/src/modules/injectResponderEventPlugin/index.js
+++ b/packages/react-native-web/src/modules/injectResponderEventPlugin/index.js
@@ -39,14 +39,29 @@ ResponderEventPlugin.eventTypes.selectionChangeShouldSetResponder.dependencies =
 ResponderEventPlugin.eventTypes.scrollShouldSetResponder.dependencies = [topScroll];
 ResponderEventPlugin.eventTypes.startShouldSetResponder.dependencies = startDependencies;
 
+let lastActiveTouchTimestamp = null;
+
 const originalExtractEvents = ResponderEventPlugin.extractEvents;
 ResponderEventPlugin.extractEvents = (topLevelType, targetInst, nativeEvent, nativeEventTarget) => {
   const hasActiveTouches = ResponderTouchHistoryStore.touchHistory.numberActiveTouches > 0;
+
+  let shouldSkipMouseAfterTouch = false;
+  if (nativeEvent.type.indexOf('touch') > -1) {
+    lastActiveTouchTimestamp = Date.now();
+  } else if (lastActiveTouchTimestamp && nativeEvent.type.indexOf('mouse') > -1) {
+    const now = Date.now();
+    shouldSkipMouseAfterTouch = lastActiveTouchTimestamp - now < 250;
+  }
+
   if (
     // Filter out mousemove and mouseup events when a touch hasn't started yet
     ((topLevelType === topMouseMove || topLevelType === topMouseUp) && !hasActiveTouches) ||
     // Filter out events from wheel/middle and right click.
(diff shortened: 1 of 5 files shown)


necolas:
Looking for people to test this out in their apps to make sure there are no regressions or unexpected issues.

packages/react-native-web/src/modules/injectResponderEventPlugin/index.js
```diff
@@ -39,14 +39,29 @@ ResponderEventPlugin.eventTypes.selectionChangeShouldSetResponder.dependencies =
...
 const originalExtractEvents = ResponderEventPlugin.extractEvents;
 ResponderEventPlugin.extractEvents = (topLevelType, targetInst, nativeEvent, nativeEventTarget) => {
   const hasActiveTouches = ResponderTouchHistoryStore.touchHistory.numberActiveTouches > 0;
+
+  let shouldSkipMouseAfterTouch = false;
+  if (nativeEvent.type.indexOf('touch') > -1) {
+    lastActiveTouchTimestamp = Date.now();
+  } else if (lastActiveTouchTimestamp && nativeEvent.type.indexOf('mouse') > -1) {
+    const now = Date.now();
+    shouldSkipMouseAfterTouch = lastActiveTouchTimestamp - now < 250;
```
doxiaodong: This  doesn't work when there is a `window.alert`  call

necolas: What does that mean?

doxiaodong: it means if you call window.alert in onPress , and then shouldSkipMouseAfterTouch is not work right

necolas: Ok. Maybe don't call alert

doxiaodong: Right, But people maybe can't understand this. Is there a way to record timestamp after onPress called tick

PR #939, opened 2018-05-08
[fix] onLayout timing without ResizeObserver 

### LINKED ISSUE #911: onLayout Callback called before ref Callback
It might be considered a bug

The OnLayout Callback  is called before the ref callback on component mount.

**If the current behavior is a bug, please provide the steps to reproduce and**

```
onLayout() {
   this.someRefAccess() // Does not work
   setTimeout(() => { 
      this.someRefAccess() //Works
  })
}
render() {
  return <View onLayout={this.onLayout} ref={(theref) => { this.someRef = theref; } }> 
}
```

On RN native (IOS & Android )
Sequence is
 1 ref callback
 2 onlayout
It used to be like this on RNWEB

Confirmed the change affects
-  Firefox 59 
-  internet explorer 11

it does not seem to affect chrome

Could it be related to a67bf0f490d92cd53711d55590b7d3119369f960
If the change is intented and we need to handle this manually could you please let us know.

Thank you for the amazing work.

### LINKED ISSUE #941: View with onLayout method cause 'Unable to find node on an unmounted component'
**The problem**
After unmounting component with ```<View onLayout={this.onLayout} />``` we steel have some event listeners for this View that fires ```Invariant Violation: Unable to find node on an unmounted component.```

**How to reproduce**
Test case: https://codesandbox.io/s/xp9780v4jp

Steps to reproduce:
1. Open https://codesandbox.io/s/xp9780v4jp in safari on ios ~~or chrome on adroid~~
2. Click "goto Page B"
3. Click "back to Page A"
4. Try to scroll

```Invariant Violation: Unable to find node on an unmounted component.```

[img_e8407dc8ae19-1](https://user-images.githubusercontent.com/1162501/39873578-a0bb773c-546b-11e8-886b-92746a7d15c0.jpeg)

**Expected behavior**
Should work with no errors

Not tested

* OS: iOS/~~Android~~
* Device: Mobile
* Browser: Safari/~~Chrome~~
* React Native for Web (version): 0.6.1
* React (version): 16.3.2

### PR DESCRIPTION
**Pull request by necolas**
onLayout is called after the component is mounted to the DOM. This makes
both the fallback and ResizeObserver code path behave the same as React
Native.

Fix #911
Fix #941
Close #939

### DIFF
--- a/packages/react-native-web/src/modules/applyLayout/index.js
+++ b/packages/react-native-web/src/modules/applyLayout/index.js
@@ -55,9 +55,10 @@ const observe = instance => {
     node._onLayoutId = id;
     resizeObserver.observe(node);
   } else {
-    const id = guid();
     instance._onLayoutId = id;
-    instance._handleLayout();
+    setTimeout(() => {
+      instance._handleLayout();
+    }, 0);
   }
 };
 
@@ -105,8 +106,6 @@ const applyLayout = Component => {
         observe(this);
       } else if (!this.props.onLayout && prevProps.onLayout) {
         unobserve(this);
-      } else if (!resizeObserver) {
-        this._handleLayout();
       }
     }
   );
@@ -123,10 +122,8 @@ const applyLayout = Component => {
     const layout = this._layoutState;
     const { onLayout } = this.props;
 
-    if (onLayout) {
+    if (this._isMounted && onLayout) {
       this.measure((x, y, width, height) => {
-        if (!this._isMounted) return;
-
         if (
           layout.x !== x ||
           layout.y !== y ||
--- a/packages/react-native-web/src/exports/AppRegistry/__tests__/index-test.js
+++ b/packages/react-native-web/src/exports/AppRegistry/__tests__/index-test.js
@@ -39,6 +39,14 @@ describe('AppRegistry', () => {
       expect(styleElement).toMatchSnapshot();
     });
 
+    test('"getStyleElement" adds props to <style>', () => {
+      const nonce = '2Bz9RM/UHvBbmo3jK/PbYZ==';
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Comment by necolas** [comment]
cc @giuseppeg

File: packages/react-native-web/src/modules/applyLayout/index.js [hunk 1]
```diff
@@ -55,10 +55,10 @@ const observe = instance => {
     node._onLayoutId = id;
     resizeObserver.observe(node);
   } else {
-    const id = guid();
```
necolas: I think this also fixes a bug. This `id` is never assigned in the registry (L50 is where the first `id` is created), and so the instance won't be removed in L65

giuseppeg: 🤦‍♂️ correct

necolas:
React Native makes the `ref` available in `onLayout` handlers on mount, so the ResizeObserver path looks like the expected behaviour https://snack.expo.io/r1vvfjlAz

giuseppeg:
so `onLayout` is fired after the component has mounted? (the ref is defined) might make sense if you think of `View` as a primitive (eg a `div`)

necolas:
I don't know what you mean. Isn't `onLayout` already being called after the component has mounted?

packages/react-native-web/src/modules/applyLayout/index.js
```diff
@@ -123,10 +122,8 @@ const applyLayout = Component => {
     const layout = this._layoutState;
     const { onLayout } = this.props;
 
-    if (onLayout) {
+    if (this._isMounted && onLayout) {
```
giuseppeg: 👍

PR #942, opened 2018-05-11
Upgrade React and all transitive dependencies

philipp-spiess:
I've made my attempt at upgrading RNW to React 16.3.2 and it seems like
everything is still working including the website and the benchmarks.

I'm pretty sure that some obscure dependency graph caused Yarn to not
de-dupe React in my first try as part of #908. To fix this, I forced an
update of all transitive dependencies as well so I suggest you also
look over the updated `yarn.lock` in more detail since there where quite
a few packages with minor updates.

`yarn check` still complains about an unmet dependency:

```
β react-native-web git:(react-16.3.2) ✗ yarn check
yarn check v1.3.2
warning "chokidar#fsevents#node-pre-gyp@^0.9.0" could be deduped from "0.9.1" to "node-pre-gyp@0.9.1"
error "benchmarks#reactxp#react@16.2.0" doesn't satisfy found match of "react@16.3.2"
error "benchmarks#reactxp#react-dom@16.2.0" doesn't satisfy found match of "react-dom@16.3.2"
warning "styled-jsx#stylis-rule-sheet#stylis@3.x" could be deduped from "3.5.0" to "stylis@3.5.0"
info Found 2 warnings.
error Found 2 errors.
info Visit https://yarnpkg.com/en/docs/cli/check for documentation about this command.
```

I tried upgrading `reactxp` as well but that broke the benchmark script.
I think this error can be ignored for now as the current version of
`reactxp` still works fine with React 16.3.2.

I've collected a couple of benchmarks before upgrading all dependencies
and compared those after upgrading. I can't spot differences that are
out of standard deviation but I'm pasting them here anyhow so you can
verify.


packages/react-native-web/src/exports/AppRegistry/__tests__/__snapshots__/index-test.js.snap
```diff
@@ -44,7 +44,7 @@ input::-webkit-inner-spin-button,input::-webkit-outer-spin-button,input::-webkit
 .rn-paddingLeft-gy4na3{padding-left:0px}
 .rn-position-bnwqim{position:relative}
 .rn-zIndex-1lgpqti{z-index:0}
-.rn-flexGrow-16y2uox{-ms-flex-positive:1;-webkit-flex-grow:1;flex-grow:1}
+.rn-flexGrow-16y2uox{-ms-flex-positive:1;-webkit-box-flex:1;-webkit-flex-grow:1;flex-grow:1}
```
philipp-spiess: I've no idea why this snapshot changed. I guess that there's some dependency to generate the prefixes and the minor upgrade added a new alias?

philipp-spiess: Never-mind! This change seems to be very much in your interest 😊  https://github.com/rofrischmann/inline-style-prefixer/issues/153

Issue #914, opened 2018-04-25
Support Layout Properties as View and Image props


brunolemos:
**Do you want to request a *feature* or report a *bug*?**

Feature / React Native Parity

**What is the current behavior?**

This is supported by `react-native`: `<View flex={1} />`, `<View width={200} />`, etc.
But not by `react-native-web`.

**What is the expected behavior?**

All props from the [LayoutPropTypes](https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/modules/LayoutPropTypes/index.js) file  should be supported as props for the `View` and `Image` components.

**Environment (include versions). Did this work in previous versions?** No

* OS: macOS 10.3.4
* Device: iOS 11.3 Simulator / iPhone X
* Browser: Chrome
* React Native for Web (version): 0.6.0
* React (version): 16.3.2

necolas:
Apparently this was an early accident in React Native. I don't think it needs to be repeated here

brunolemos:
This's been working for years and there is no intention to be dropped any time soon.

Anyways,
Ok, I'll support this locally by using patch-package.

PR #943, opened 2018-05-12
Support View Style Props

brunolemos:
> I'm opening and closing this PR because it is for feedback purposes, as you don't seem to want to have this feature on this repo (#914). Although, I really think react-native-web should support this syntax, because react-native does and I believe both APIs should be compatible.

Is this the right way to implement it? How could I optimize it?

E.g. what could I use instead of adding lodash? But it worked without adding it so I believe it was already indirectly included anyway.

Do I really need to do this `StyleSheet.create` inside the render method or is there a better way?


brunolemos:
Dammit, GitHub doesn't update the code here because I closed. I'll open another one.

PR #944, opened 2018-05-12
Support View Style Props

brunolemos:
> This PR is for feedback purposes, as you currently don't seem to want this feature on this repo (#914). Although, I really believe react-native-web should support this syntax, because react-native does and both APIs should be compatible.

Is this the right way to implement it? What would you change? How could it be optimized?

E.g. what could I use instead of adding lodash/omit lodash/pick?
> PS: It worked without adding lodash, so I believe this dependency is already being indirectly included anyway.

Do I really need to do this `StyleSheet.create` inside the render method or is there a better way?

### DIFF
--- a/packages/react-native-web/package.json
+++ b/packages/react-native-web/package.json
@@ -17,6 +17,7 @@
     "fbjs": "^0.8.16",
     "hyphenate-style-name": "^1.0.2",
     "inline-style-prefixer": "^4.0.0",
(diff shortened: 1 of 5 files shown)


necolas:
This is not something I think should be added to the library. It's one of the mistakes in the RN API that they'd be better off not documenting to avoid people using it

PR #948, opened 2018-05-14
[add] support for CSP policy requiring 'nonce' on <style>

necolas:
CSP policy may prevent writing to `<style>` unless a `nonce` attribute
is set. This change makes that possible by moving the modality-related
styles into the main style sheet, and allowing additional props to be
provided to the `<style>` element when rendering on the server. For
example:

```
const { element, getStyleElement } = AppRegistry.getApplication('App');
const html = renderToString(element);
const css = renderToStaticMarkup(getStyleElement({ nonce }));
```


necolas:
8c198b1c98d2e383823e8ba222ac359808c2447f

dafeizizhu:
It seems to be a solution for server rendering. How to deal with client rendering, for example adding a `nonce` attribute to `<style>` element, thanks.

PR #949, opened 2018-05-15
babel-plugin supports custom web modules

Kinka:
In order to use my own web module implements instead of react-native-web's, such as Modal which is still UnimpementedView.
```
{
  "plugins": [
    ["react-native-web", {
      "Modal": "/your/path/to/Modal", "Text": "/your/path/to/Text"
    }]
  ]
}
```

### DIFF
--- a/packages/babel-plugin-react-native-web/README.md
+++ b/packages/babel-plugin-react-native-web/README.md
@@ -21,6 +21,17 @@ yarn add --dev babel-plugin-react-native-web
 }
 ```
 
+You can set your own web modules in babel options:
+```
+{
+  "plugins": [
+    ["react-native-web", {
+      "Modal": "/your/path/to/Modal", "Text": "/your/path/to/Text"
+    }]
+  ]
+}
+```
+
 ## Example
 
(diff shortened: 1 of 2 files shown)


necolas:
I don't think this is a good idea. It can't ensure that the APIs of substituted modules are compatible and it can add *any* arbitrary export to the package. There are existing ways to hack modules in a package without adding it to this plugin. I'm actually looking at removing the need for the plugin and marking the package for tree-shaking bundlers

Kinka:
Would you please guide me to the `existing ways`?
As for compatibility, I think that's developer's responsibility to keep it. In my case, I don't want to break my react-native codes such as `import {Modal} from 'react-native'` since react-native-web still not providing its implementation.
Still, looking forward to other solutions.

necolas:
You can use webpack or babel alias features. You can alias `react-native-web/dist/exports/Modal` to another module you want.

Kinka:
Thanks, I'll check it out then.

Kinka:
webpack alias does not work like this:
```
'react-native': 'react-native-web',
'react-native-web/dist/exports/Modal': '/path/to/my/modal'
```

PR #955, opened 2018-05-18
fix scrollTop issue on Chrome/Firefox

davepack:
MeasureInWindow returns an incorrect top and left value on Chrome and Firefox because body.scrollTop and scrollLeft return zero on Chrome and Firefox. The current method does work on Safari. This new method works on both Safari and Chrome/Firefox.

### DIFF
--- a/packages/react-native-web/src/exports/UIManager/index.js
+++ b/packages/react-native-web/src/exports/UIManager/index.js
@@ -17,14 +17,15 @@ const getRect = node => {
   node = node.offsetParent;
 
   while (node && node.nodeType === 1 /* Node.ELEMENT_NODE */) {
-    left += node.offsetLeft;
-    top += node.offsetTop;
+    const scrollTop = node.nodeName === 'BODY' ? window.pageYOffset : node.scrollTop;
+    const scrollLeft = node.nodeName === 'BODY' ? window.pageXOffset : node.scrollLeft;
+
+    top += node.clientTop + node.offsetTop - scrollTop;
+    left += node.clientLeft + node.offsetLeft - scrollLeft;
+
     node = node.offsetParent;
   }
 
-  top -= window.pageYOffset;
-  left -= window.pageXOffset;
-  
   return { height, left, top, width };
 };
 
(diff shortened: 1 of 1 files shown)


necolas:
This looks like it will break the other measurements though

davepack:
Shoot sorry I haven't gotten back to this. All this does is shift the returned top and left values if there is any scroll. Relative measurements will be the same. Quickly tested and appears to still work.

necolas:
Yes but elements can also be scrolled within a ScrollView, which is not the window

davepack:
Ah of course. I’ll work on that and get a solution up hopefully in the next few days.

necolas:
Ping

davepack:
Worked on this today, thought I had a solution, but browser differences are throwing a wrench in it. Done for today but tonight/tomorrow I’ll push up what I have and see if I can create a demo to illustrate the issues.

davepack:
@necolas Is there a reason getRect wasn't using getBoundingClientRect? So far it's the only way that has been absolutely accurate in latest Safari, Chrome, and Firefox on macOS and mobile Safari.

This repo demonstrates measureInWindow working with getBoundingClientRect:
https://github.com/davepack/rnw-scroll-measure-test
http://davepack.com/rnw-scroll-measure-test/

I don't have Windows so I can't test Edge, but it seems to have plenty of support in older browsers for at least the top, left, height and width properties: https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect#Browser_compatibility.

necolas:
`getBoundingClientRect` doesn't exclude the influence of transforms, which is how React Native does it

davepack:
Ok, tbh I don't know enough about transforms or all the edge cases this needs to take care of. Without some more guidance I'm probably not the right person to work on this. And I really don't have time to spend researching and understanding all this. I'm in the middle of moving and starting a new job in a couple weeks, after which I'll have less time. Sorry I couldn't get this done. 

I'll update the PR with the latest code and deploy it to the demo page. The main problem this last iteration was borders were screwing up the left and top measurements. They would be off by the sum of top and left border widths of parent nodes... but only in Safari and Chrome. Firefox didn't care about borders. Accounting for borders made it work, but broke Firefox. 🤷‍♂️ 

Really wish I could have solved this! Feel free to keep it open and maybe somebody else can pick it up, I'll at least create an issue.

necolas:
Just closing this because it doesn't resolve the issue and isn't being worked on

hushicai:
sometimes actually does need the values with css transforms.

I am using a inverted Flatlist, and the measure api returned wrong positions.

PR #993, opened 2018-06-13
[fix] cannot convert undefined or null to object if node already unmo…

johakr:
…unted in resizeObserver

### DIFF
--- a/packages/react-native-web/src/modules/applyLayout/index.js
+++ b/packages/react-native-web/src/modules/applyLayout/index.js
@@ -52,6 +52,7 @@ const observe = instance => {
 
   if (resizeObserver) {
     const node = findNodeHandle(instance);
+    if (!node) return;
     node._layoutId = id;
     resizeObserver.observe(node);
   } else {
@@ -64,6 +65,7 @@ const unobserve = instance => {
   delete registry[instance._layoutId];
   if (resizeObserver) {
     const node = findNodeHandle(instance);
+    if (!node) return;
     delete node._layoutId;
     resizeObserver.unobserve(node);
   } else {


necolas:
Have you got example code of the issue this is meant to fix?

johakr:
I will try to cutout an example repository, but this will take some time (probably weeks) since the project this issue occurs in is quite complex. The specific place where this issue occurs is when I use Flipkart/recyclerlistview and add new elements to the top of the list. I don't know if that's an issue with recyclelistview in the first place, but I think react-native-web should be robust enough to handle those kind of edge cases.

necolas:
> The specific place where this issue occurs is when I use Flipkart/recyclerlistview and add new elements to the top of the list

Is that something that can be reproduced with a reduced test case in codesandbox? Test cases are probably the most important thing when making changes. It gives everyone an example that can be used to check for regressions and understand where in the code issues originate from. Right now I can't see how calling `observe` in those lifestyle methods could happen while the component is unmounted, so I just want to understand what is going on. Thanks

necolas:
Did you make a test case yet?

johakr:
Sorry, really tried to come up with a reduced test case to reproduce this issue, but had no luck and really am out of ideas. Must be some kind of obscure race condition. Since we moved to another implementation in our main project, there is no pressure from my side to get this merged. So feel free to close this pull request, if you see no need to get this merged.

For what it's worth here's the error call stack:

`Uncaught TypeError: Cannot convert undefined or null to object
    at unobserve (index.js?20e4:67)
    at Text.componentWillUnmount (index.js?20e4:107)
    at callComponentWillUnmountWithTimer (react-dom.development.js?61bb:9703)
    at HTMLUnknownElement.callCallback (react-dom.development.js?61bb:100)
    at Object.invokeGuardedCallbackDev (react-dom.development.js?61bb:138)
    at invokeGuardedCallback (react-dom.development.js?61bb:187)
    at safelyCallComponentWillUnmount (react-dom.development.js?61bb:9710)
    at commitUnmount (react-dom.development.js?61bb:9956)
    at unmountHostComponents (react-dom.development.js?61bb:10299)
    at commitDeletion (react-dom.development.js?61bb:10329)`

necolas:
Another implementation of what?

johakr:
We dropped `Flipkart/recyclerlistview`

necolas:
Out of interest, what are you using instead?

PR #1047, opened 2018-07-23
Correct type of animationIterationCount for 2+ animations

### LINKED ISSUE #1044: animationIterationCount has wrong type for 2 or more animations
**Do you want to request a feature or report a bug?**
A bug.

if a minimal demo of the problem via Glitch or similar (template:
https://glitch.com/edit/#!/react-native-web-playground).**

_Glitch demo code:_ https://glitch.com/edit/#!/zesty-carpenter?path=src/App/index.js:46:2
_Live Glitch demo:_ https://zesty-carpenter.glitch.me/

_Reproduction steps:_
1. Create a component which uses StyleSheet for styling and apply `styles.animation`:
```js
var styles = StyleSheet.create({
  animation: {
    animationDuration: '1s, 2s',
    animationIterationCount: '1, infinite',
    animationName: [
      {
        from: { top: '2em' },
        to: { top: 0 }
      },
      {
        '15%': {
          transform: [{ scale: 2 }],
          backgroundColor: 'red'
        },
        '30%': {
          transform: [{ scale: 1 }]
        }
      }
    ]
  }
})
```
2. Run the project locally and open the Developer Tools console.
3. Realize there's an error: `Warning: Failed prop type: Invalid prop `animationIterationCount` supplied to `View`.`

`animationIterationCount` should accept string values for multiple animations, much in the same way `animationDuration` does. There should be no `Failed prop type` error.

_OS:_ macOS Sierra 10.13.6
_Browser:_ Chrome 67.0.3396.99
_React Native for Web (version):_ 0.8.8
_React (version):_ 16.3.1

jdecked:
Fixes #1044.

**Do you want to request a feature or report a bug?**
A bug.

if a minimal demo of the problem via Glitch or similar (template:
https://glitch.com/edit/#!/react-native-web-playground).**

_Glitch demo code:_ https://glitch.com/edit/#!/zesty-carpenter?path=src/App/index.js:46:2
_Live Glitch demo:_ https://zesty-carpenter.glitch.me/

_Reproduction steps:_
1. Create a component which uses StyleSheet for styling and apply `styles.animation`:
```js
var styles = StyleSheet.create({
  animation: {
    animationDuration: '1s, 2s',
    animationIterationCount: '1, infinite',
    animationName: [
      {
        from: { top: '2em' },
        to: { top: 0 }
      },
      {
        '15%': {
          transform: [{ scale: 2 }],
          backgroundColor: 'red'
        },
        '30%': {
          transform: [{ scale: 1 }]
        }
      }
    ]
  }
})
```
2. Run the project locally and open the Developer Tools console.
3. Realize there's an error: `Warning: Failed prop type: Invalid prop `animationIterationCount` supplied to `View`.`

`animationIterationCount` should accept string values for multiple animations, much in the same way `animationDuration` does. There should be no `Failed prop type` error.

_OS:_ macOS Sierra 10.13.6
_Browser:_ Chrome 67.0.3396.99
_React Native for Web (version):_ 0.8.8
_React (version):_ 16.3.1

### DIFF
--- a/packages/react-native-web/src/exports/ActivityIndicator/__tests__/__snapshots__/index-test.js.snap
+++ b/packages/react-native-web/src/exports/ActivityIndicator/__tests__/__snapshots__/index-test.js.snap
@@ -15,28 +15,11 @@ exports[`components/ActivityIndicator prop "animating" is "false" 1`] = `
   <View
     style={
       Object {
-        "animationDuration": "0.75s",
-        "animationIterationCount": "infinite",
-        "animationName": Array [
+        "animation": Array [
           Object {
-            "0%": Object {
-              "transform": Array [
-                Object {
-                  "rotate": "0deg",
-                },
-              ],
-            },
-            "100%": Object {
-              "transform": Array [
-                Object {
-                  "rotate": "360deg",
-                },
-              ],
-            },
+            "animationPlayState": "paused",
           },
         ],
-        "animationPlayState": "paused",
-        "animationTimingFunction": "linear",
         "height": 20,
         "visibility": "hidden",
         "width": 20,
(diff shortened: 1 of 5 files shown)


necolas:
I think it might be better to redefine how animations are specified to use an array of objects a bit like how `transform` works

necolas:
Oh fantastic, you're making change, thanks! Please could you drop the 'animation' prefix from the object keys and have a go at updating the View documentation in the "website" package

necolas:
Actually, forget about dropping "animation" from the keys. I'll consider that separately and don't want to waste your time if it introduces complexities

necolas:
I asked that you update the view docs so the new property and type is included

jdecked:
My bad! All done now, thanks for your patience.

necolas:
Closing this because:

1. The PR doesn't work as described. The logic that generates the CSS hasn't been updated so all the existing animations don't run in the storybook. This is something I could fix but…

2. There's a problem with housing all the animation properties in an array under `animation`, which is that you can't modify individual characteristics of the animation (e.g., play state) without including all the rest of the animation properties, as the style flattener will replace any prior `animation` definition. This is a design problem.

I'll work a separate fix for #1044 that avoids these problems.

paularmstrong:
> I'll work a separate fix for #1044 that avoids these problems.

@necolas did you get anywhere on that? Otherwise we can take another stab.

necolas:
Please do. The animation properties need to accept arrays of values

PR #1048, opened 2018-07-23
Solve #1042 when scroll is disabled

yairopro:
Tested on chrome, firefox and safari.

### DIFF
--- a/packages/react-native-web/src/exports/ScrollView/ScrollViewBase.js
+++ b/packages/react-native-web/src/exports/ScrollView/ScrollViewBase.js
@@ -146,9 +146,6 @@ export default class ScrollViewBase extends Component<*> {
         if (handler) {
           handler(e);
         }
-      } else {
-        // To disable scrolling in all browsers except Chrome
-        e.preventDefault();
       }
     };
   };
@@ -206,7 +203,8 @@ export default class ScrollViewBase extends Component<*> {
 // https://developers.google.com/web/updates/2017/01/scrolling-intervention
 const styles = StyleSheet.create({
   scrollDisabled: {
-    touchAction: 'none'
+		overflowY : 'hidden',
+		overflowX : 'hidden',
   },
   hideScrollbar: {
     scrollbarWidth: 'none'
--- a/packages/react-native-web/src/exports/ScrollView/ScrollViewBase.js
+++ b/packages/react-native-web/src/exports/ScrollView/ScrollViewBase.js
@@ -203,8 +203,8 @@ export default class ScrollViewBase extends Component<*> {
 // https://developers.google.com/web/updates/2017/01/scrolling-intervention
 const styles = StyleSheet.create({
   scrollDisabled: {
-		overflowY : 'hidden',
-		overflowX : 'hidden',
+    overflowY : 'hidden',
+    overflowX : 'hidden',
   },
   hideScrollbar: {
     scrollbarWidth: 'none'
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,6 @@
 coverage
 node_modules
 dist
+
+# webstorm
+.idea


packages/react-native-web/src/exports/ScrollView/ScrollViewBase.js
```diff
@@ -137,12 +137,8 @@ export default class ScrollViewBase extends Component<*> {
   _createPreventableScrollHandler = (handler: Function) => {
     return (e: Object) => {
       if (this.props.scrollEnabled) {
-        if (handler) {
+        if (handler)
           handler(e);
-        }
-      } else {
-        // To disable scrolling in all browsers except Chrome
-        e.preventDefault();
```
necolas: Why is this code removed?

packages/react-native-web/src/exports/ScrollView/ScrollViewBase.js
```diff
@@ -200,6 +196,7 @@ export default class ScrollViewBase extends Component<*> {
 // https://developers.google.com/web/updates/2017/01/scrolling-intervention
 const styles = StyleSheet.create({
   scrollDisabled: {
-    touchAction: 'none'
-  }
+		overflowY : 'hidden',
```
necolas: You can use 'overflow:hidden'. Also please run the formatter

packages/react-native-web/src/exports/ScrollView/ScrollViewBase.js
yairopro: To be clearer, when the scrolling is disabled:
1. `touchAction: 'none'` doesn't prevent scrolling
2. `e.preventDefault()` was the real code which was preventing the scrolling. The problem is that this function does too much, *it catches the event, and stop it*. This prevent *all scrolling*, even for scrolling with outside views of the `ScrollView` (parents view) because the event is catched.

necolas: As the comments explain, different browsers required different ways to disable scrolling without removing the scroll bars or changing the scroll position. Propagation of the scroll event up to ancestors is a problem

yairopro: The comment on `preventDefault` says "To disable scrolling in all browsers except Chrome".
But it does disable scroll in chrome too. And, as you said, it stops the propagation of the scroll event up to ancestors. Then that really doesn't seem a good way to prevent scroll.

The new way to disable scroll using style's `overflowX` & `overflowY` properties works for all browser. You can test it in this sandbox with the browser of your choice.
Test on browser: https://w6yo77mrp5.codesandbox.io
Code: https://w6yo77mrp5.codesandbox.io/s/w6yo77mrp5

necolas: This patch doesn't seem to fix the problem either. Using your test case + this patch, when a touch-scroll starts over the disabled horizontal `FlatList` it doesn't scroll the parent `ScrollView`.

yairopro: Yes you're right about the **touch** -scroll.
But the ` e.preventDefault()` won't help you anyway. ^^

yairopro:
What should be a better solution ?

hushicai:
@yairopro your pr.

yairopro:
I hope too. But I still don't know why the touch scroll doesn't work.

packages/react-native-web/src/exports/ScrollView/ScrollViewBase.js
hushicai: > This patch doesn't seem to fix the problem either. Using your test case + this patch, when a touch-scroll starts over the disabled horizontal `FlatList` it doesn't scroll the parent `ScrollView`.

Is this the problem related to #1160 ?

necolas:
It's worth rebasing master to pick up a related fix for touch

artalat:
Hi, any updates on this?

MoOx:
@yairopro could you rebase?

yairopro:
@MoOx  Done 😁

PR #1053, opened 2018-07-27
Fix text-decoration issue on Safari

anton-md:
Styling text decoration on Safari requires you to write each css property with a `-webkit`-prefix. Currently if you try to style a link with for example:
```
const style = StyleSheet.create({
  link {
    textDecorationColor: '#f0f',
    textDecorationLine: 'underline',
    textDecorationStyle: 'solid',
  }
});
```
It won't work on Safari since `resolveTextDecoration` returns `textDecoration: "underline solid #f0f"`

This PR separates the text decoration styles so that `inline-style-prefixer` can prepend `-webkit` on each css property on Safari.

### DIFF
--- a/packages/react-native-web/src/exports/StyleSheet/__tests__/createReactDOMStyle-test.js
+++ b/packages/react-native-web/src/exports/StyleSheet/__tests__/createReactDOMStyle-test.js
@@ -227,7 +227,7 @@ describe('StyleSheet/createReactDOMStyle', () => {
           textDecorationLine: 'underline'
         })
       ).toEqual({
-        textDecoration: 'underline'
+        textDecorationLine: 'underline'
       });
     });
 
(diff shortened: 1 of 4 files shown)


necolas:
Do all other browsers continue to work as expected?

anton-md:
I tested in Safari, Chrome and Firefox. 
IE and Edge doesn't support [text-decoration styling ](
https://caniuse.com/#search=text-decoration)at all. 

One quirk is that I can't seem to get `Dashed line-through with custom color` and `Dotted line-through with custom color` to work in the examples. Maybe it's not fully supported by Safari?
Here is a fiddle: http://jsfiddle.net/31hf0se2/1/

**Chrome:**
screen shot 2018-07-29 at 11 48 31

**Firefox:**
screen shot 2018-07-29 at 11 45 17

**Safari:***
screen shot 2018-07-29 at 11 47 11

necolas:
Edge does support CSS2 text decoration, so this change may be a regression there

anton-md:
Added a check for Edge and IE that uses the `textDecorationLine` as `textDecoration`. 

*Edge:*
screen shot 2018-08-01 at 11 03 02

*IE 11:*
screen shot 2018-08-01 at 11 10 14

necolas:
I don't think it's a good idea to add an extra [6 KB](https://github.com/faisalman/ua-parser-js#uaparserjs) to the gzip bundle size to detect if the browser is Edge or IE.

Another approach might be to always include `textDecoration` as a fallback. I don't think there will be any CSS order issues.

PR #1064, opened 2018-08-08
Adds accessibility label to the picker for 508 Accessibility 

DaKaZ:
When using RNW for accessible web sites, it is necessary at times to pass an accessibilityLabel to the `<select />` component.  This small PR simply makes that possible.

### DIFF
--- a/packages/react-native-web/src/exports/Picker/index.js
+++ b/packages/react-native-web/src/exports/Picker/index.js
@@ -22,6 +22,7 @@ import { arrayOf, bool, func, number, oneOfType, string } from 'prop-types';
 const pickerStyleType = StyleSheetPropType(PickerStylePropTypes);
 
 type Props = {
+  accessibilityLabel?: string,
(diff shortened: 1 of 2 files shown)


necolas:
Looking at the current RN docs it seems Picker now accepts all the View props. Want to give that a shot?

DaKaZ:
@necolas sure thing... let me take a look at how you handle View's currently and try to make Picker mirror it

DaKaZ:
Hmmm, we could do that but I don't think we want to.  You are definitely more qualified to make this call but I think that props like `onResponderMove` don't translate into web's select element.  Seems to me we want to be a bit more selective.

DaKaZ:
I guess since the other elements do the same thing, there is consistency in doing it here.  Take a peek 👍

DaKaZ:
@necolas what do you think of the approach?

DaKaZ:
Glad it made in there one way or another 👍

Issue #1068, opened 2018-08-10
Allow accessibilityRole menuitem to route enter to onClick


comp615:
**Is your feature request related to a problem? Please describe.**
Previously in our dropdowns and lists we used `role=button` for items within the list. Good aria practice suggests that the container should be `role=menu`, with each item as a `role=menuitem`. Currently, RNW routes enter presses on `role=button` to `onClick`, but not `role=menuitem`. This means the change breaks enter handling.

More generally, interested to hear how much of the best practices should live in RNW vs in the app code. (e.g. Would RNW ever handle that up/down navigation within menu roles?)

**Describe a solution you'd like**
One solution would be to extend the handling of enter -> onClick not just for button, but a set of possible roles. See:
https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/createElement/index.js#L51

**Describe alternatives you've considered**
The alternative is that in the same code where the app handles up/down within the menu, could additionally handle enter and use that to trigger onClick when RNW does not.

**Additional context**
Demo: https://codesandbox.io/s/zxomy6lr7l
https://www.w3.org/TR/wai-aria-1.1/#menuitem
https://www.w3.org/WAI/GL/wiki/Using_ARIA_menus#Keyboard_Actions_if_focus_is_on_a_menu_item

jpstrikesback:
Was just about to post a query as to whether RNW had an intent to support keyboard nav on role=menu / menuitem (and enter, maybe esc to dismiss focus is out of scope?)

necolas:
Looks like a good idea.

Arrow key / spatial navigation is modeled globally and can be bolted on using existing libraries and hooking into an attribute exposed by RNW https://github.com/necolas/react-native-web/blob/master/packages/website/guides/accessibility.md#spatial-navigation

PR #1069, opened 2018-08-10
Handle Enter/Space for additional a11y roles

comp615:
Addresses #1068, there may be more roles to handle as well, but they can be added as they come up?

### DIFF
--- a/packages/react-native-web/src/exports/createElement/index.js
+++ b/packages/react-native-web/src/exports/createElement/index.js
@@ -23,8 +23,6 @@ EventPluginHub.injection.injectEventPluginsByName({
 const isModifiedEvent = event =>
   !!(event.metaKey || event.altKey || event.ctrlKey || event.shiftKey);
 
-const buttonLikeRoles = ['button', 'menuitem'];
-
 /**
  * Ensure event handlers receive an event of the expected shape. The 'button'
  * role – for accessibility reasons and functional equivalence to the native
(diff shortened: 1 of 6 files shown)


packages/react-native-web/src/exports/createElement/index.js
```diff
@@ -23,6 +23,8 @@ EventPluginHub.injection.injectEventPluginsByName({
 const isModifiedEvent = event =>
   !!(event.metaKey || event.altKey || event.ctrlKey || event.shiftKey);
 
+const buttonLikeRoles = ['button', 'menuitem'];
```
necolas: Let's use an object rather than an array

necolas:
tabindex/focus stuff to check over is here: https://github.com/necolas/react-native-web/blob/e8f2c987864562bda11bec6611db98b7239688d6/packages/react-native-web/src/modules/createDOMProps/index.js#L116

comp615:
Thank for that pointer. Sorry for the delay, we had to upgrade to the new accessibility stuff which was its own task. 

Updated list to accessibility util and moved into its own file for easy reuse. Not sure if that's overkill or not, but just let me know what your preference is!

One question on styles, I saw in createDOMProps `role===button` also applies some reset styles. Should that be changed to include this as well? Or are those intended specifically for role button.

https://github.com/necolas/react-native-web/blob/0e81c6ef2758d4ca9b2099b1d04a4f8c417f0f43/packages/react-native-web/src/modules/createDOMProps/index.js#L153

comp615:
@necolas any thoughts on the update? Not sure how to update it from changes requested status :( maybe it just stays like that forever.

PR #1076, opened 2018-08-17
Add inputAccessoryViewID to TextInput props

jlongster:
`InputAccessoryView` is a relatively new component for react native: https://facebook.github.io/react-native/blog/2018/03/22/building-input-accessory-view-for-react-native

React Native Web actually does already recognized this, even if it just implements it as an `UnimplementedView`. But it does not recognize the prop on `TextInput` that you use to specify which view to use for an input when it's focused: `inputAccessoryViewID`. It forwards it to the `input` DOM element and you see this error:

```
Warning: React does not recognize the `inputAccessoryViewID` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `inputaccessoryviewid` instead. If you accidentally passed it from a parent component, remove it from the DOM element.
```

This PR just pulls the prop off and ignores it.

### DIFF
--- a/packages/react-native-web/src/exports/TextInput/index.js
+++ b/packages/react-native-web/src/exports/TextInput/index.js
@@ -81,6 +81,7 @@ class TextInput extends Component<*> {
     clearTextOnFocus: bool,
     defaultValue: string,
     editable: bool,
+    inputAccessoryViewID: string,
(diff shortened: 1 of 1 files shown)


jlongster:
So many things I wish mobile browsers supported :)

PR #1081, opened 2018-08-23
allow user to override the tabIndex value when explicitly needed

DaKaZ:
There are times when a user needs to explicitly set a tabIndex on an accessible component.  This small patch allows it like:
```
        <TouchableOpacity
          onPress={() => {
            const callback = onPress();
            if (callback) callback();
            else navigation.navigate('Scope');
          }}
          style={Styles.container}
          accessibilityLabel="Select Scope"
          accessibilityRole="button"
          accessibile
          testID="selectScope"
          tabIndex="1"
        >
          <View style={Styles.navBarTitleContainer}>
            <Icon style={Styles.headerIcon} size={20} name="map-marker" />
            <Text
              accessibilityRole="heading"
              aria-level="1"
              style={Styles.navBarTitle}
            >
              {scopeTitle}
            </Text>
          </View>
        </TouchableOpacity>
```

### DIFF
--- a/packages/react-native-web/src/exports/View/filterSupportedProps.js
+++ b/packages/react-native-web/src/exports/View/filterSupportedProps.js
@@ -59,6 +59,7 @@ const whitelist = {
   onClick: true,
   onClickCapture: true,
   rel: true,
+  tabIndex: true,
   target: true
 };
 
--- a/packages/react-native-web/src/modules/createDOMProps/index.js
+++ b/packages/react-native-web/src/modules/createDOMProps/index.js
@@ -134,12 +134,12 @@ const createDOMProps = (component, props, styleResolver) => {
   } else if (role === 'button' || role === 'textbox') {
     if (accessible !== false && focusable) {
       domProps['data-focusable'] = true;
-      domProps.tabIndex = '0';
+      domProps.tabIndex = domProps.tabIndex || '0';
     }
   } else {
     if (accessible === true && focusable) {
       domProps['data-focusable'] = true;
-      domProps.tabIndex = '0';
+      domProps.tabIndex = domProps.tabIndex || '0';
     }
   }
 


DaKaZ:
@necolas I can give you TWO fantastic use cases.

1. If you use `react-navigation` (the leading react-native navigation tool), it renders the navbar AFTER the main content, then uses `flexDirection: 'column-reverse` to push it above the main content.   Many apps will put clickable elements in the navbar haeder.  Because of the render order on the DOM, this makes the tabOrder for the navbar LAST which breaks 508/Accessibility unless you explicitly set the tabIndex to 1.  Having these visually first on the page but with a tab order of last is a violation of 508 accessibility.  See: https://github.com/react-navigation/react-navigation-stack/blob/master/src/views/StackView/StackViewLayout.js#L625

2. If you have an element with `accessibilityRole='link'` that will render as an `<a />` element after RNW processes the file.  However, there are many times (at least in our apps) where a link does not actually have an `href` tag.  If an anchor tab does not have href, it is to keyboard accessible UNLESS it has a tab index set.  Since our behavior is that of a link (we are transitioning to another page) it is not appropriate to use the button or other roles, but we need to include tabIndex to make it accessible.

Please reconsider your denial of this PR.

necolas:
1 sounds like something for react navigation to fix in their project

2 can be done by using `accessible` to make an element focusable. If that doesn't work for links I can patch that. But I'm surprised your using links without href for navigation. AFAIK links without href are meant to be understood as placeholders, i.e., active page, rather than be expected to cause navigation https://stackoverflow.com/a/33046203

DaKaZ:
@necolas I have to respectfully disagree - there are times when `flexDirection` and other similar "tricks" are necessary as pointed out in the comment on react-navigations's code.  I merely provided that as a valid (and compelling) use case.  

That said, what is the harm with this PR to allow a user to explicitly pass in tabIndex if they need it?  Should RNW be so opinionated that it does not support valid HTML markup when the developer thinks it needed?  I don't see any downside to this PR and frankly stating that you don't agree with valid (and compelling) use case seems counter to the open source philosophy of this project.  Is there a technical justification that I am missing as to why this PR could introduce problems?  I can, of course, fork the project and use that but it seems like such a waste when all I want to do is help to improve yours and others experiences with this great library.

DaKaZ:
@necolas okay, thank you for taking the time to respond.  While I don't necessarily agree with your responses I do respect them and the work you have done here.  We have implements a work around that uses `setNativeProps` so we'll continue with that.  Thanks

Issue #13, opened 2015-09-08
Text: implement `numberOfLines` > 1


necolas:
As far as I know, this is non trivial cross-browser but should be possible using `document.createRange` and `getClientRects`: https://developer.mozilla.org/en-US/docs/Web/API/Range

basketofsoftkittens:
+1

nbclark:
Webkit only, but:

`    overflow: hidden;
    display: -webkit-box;
    -webkit-line-clamp: 3;
    -webkit-box-orient: vertical;
`

http://hackingui.com/front-end/a-pure-css-solution-for-multiline-text-truncation/

nbclark:
https://github.com/doubledutch/react-native-web/commit/c817f39ffbea4ae0e64e6538a7031b98aee8a5ea

Will submit a pull once I add a test+sample

necolas:
Unfortunately, `-webkit-line-clamp` has poor i18n support. It doesn't work when the layout is RTL, or when the language has no spaces, and it doesn't render a language-appropriate equivalent of `…`.

VinceBT:
+1

lstrrs:
I can try and take a shot at this but would like some comments in the approach I have in mind. 

The gist of it is:

- Use `getBoundingClientRect` to get a targetWidth
- Use `getComputedStyle` to get the text styles
- Use a `Canvas` and `measureText` method
- Break the targeted text into lines with the specified "ellipsis" component/text

Any comments? Of course there are some performance considerations.

steida:
I think we can use https://github.com/necolas/react-native-web/issues/795 and compute the current number of lines via `height / lineHeight`, then restrict it.

tafelito:
will any of this solutions work with rnw? [react-text-truncate](https://github.com/ShinyChang/React-Text-Truncate) or [react-lines-ellipsis](https://github.com/xiaody/react-lines-ellipsis)

brunolemos:
Isn't using `-webkit-line-clamp` better than nothing? 🤔

necolas:
That demo doesn't show ellipsis for LTR and clips the Arabic text vertically. Still looks broken to me against that criteria alone

brunolemos:
Okay... I'm glad we can just use our forks.

drumnation:
Was able to get react-lines-ellipsis working using Platform.OS === 'web'

greenais:
@drumnation, thank you for tip - I can confirm it works. 
[...]

PR #1090, opened 2018-09-01
[add] Text support for numberOfLines greater than 1

brunolemos:
Related: #13 

Today's support is much better than when #13 was created 3 years ago, maybe it's worth adding this? Better than no support at all IMO.

[CanIUse](https://caniuse.com/#feat=css-line-clamp)


hushicai:
Thanks for this.

I have merged it to [my forked](https://github.com/hushicai/react-native-web), and fix my case.

PR #1091, opened 2018-09-02
fix(ScrollView): support scroll props

gwuhaolin:
support props added:
- onScrollBeginDrag
- onMomentumScrollBegin
- onScrollEndDrag
- onMomentumScrollEnd

### DIFF
--- a/packages/react-native-web/src/exports/ScrollView/ScrollViewBase.js
+++ b/packages/react-native-web/src/exports/ScrollView/ScrollViewBase.js
@@ -129,7 +129,10 @@ export default class ScrollViewBase extends Component<*> {
         onTouchMove={this._createPreventableScrollHandler(this.props.onTouchMove)}
         onWheel={this._createPreventableScrollHandler(this.props.onWheel)}
         ref={this._setViewRef}
-        style={StyleSheet.compose(style, !scrollEnabled && styles.scrollDisabled)}
+        style={StyleSheet.compose(
+          style,
+          !scrollEnabled && styles.scrollDisabled
+        )}
       />
     );
   }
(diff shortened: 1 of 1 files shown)


packages/react-native-web/src/exports/ScrollView/ScrollViewBase.js
```diff
@@ -178,11 +188,17 @@ export default class ScrollViewBase extends Component<*> {
...
 
   _handleScrollEnd(e: Object) {
-    const { onScroll } = this.props;
+    const { onScroll, onScrollEndDrag, onMomentumScrollEnd } = this.props;
     this._state.isScrolling = false;
     if (onScroll) {
       onScroll(normalizeScrollEvent(e));
     }
+    if (onScrollEndDrag) {
+      onScrollEndDrag(normalizeScrollEvent(e));
```
necolas: Are these events really the same in React Native? I suspect 'drag' is meant to end when you lift your finger, but momentum continues until the scroll stops.

TannerAtVenu:
I would love to see this merged in as soon as possible. Working on a project that needs these props.

necolas:
I'm not merging something incomplete and with no response from the author

packages/react-native-web/src/exports/ScrollView/ScrollViewBase.js
geminiyellow: hi @necolas , `onscrollbegindrag` be listed in `rn 0.57`'s doc,

https://facebook.github.io/react-native/docs/scrollview#onscrollbegindrag

could you push this PR forward ?

@gwuhaolin looks like give up this PR, but 

> I'm not merging something incomplete and with no response from the author

so, could you give some update?

KingAmo:
at least `onscrollbegindrag` should be supported

Issue #1104, opened 2018-09-12
Warning: Unknown event handler property `onValueChange`. It will be ignored.


grit96:
**The problem**

The first render of a `Picker` with an `onValueChange` prop shows the following warning in the console:

```
Warning: Unknown event handler property `onValueChange`. It will be ignored.
```

The handler still gets fired on change despite the warning.

**How to reproduce**

https://codesandbox.io/s/x2wmj2moxq

**Expected behavior**

No warnings in the console.

**Environment**

* React Native for Web: `0.8.10`
* React: `16.4.2`
* Browser: `Chrome`

PR #1107, opened 2018-09-14
`onValueChange` should not be passed to select tag

tomoya55:
Closes https://github.com/necolas/react-native-web/issues/1104

### Background

Since https://github.com/necolas/react-native-web/pull/1064, Picker has started to pass `otherProps` to browser's select element. As the element doesn't have such a property, `onValueChange` should be excluded from `otherPropes`.


necolas:
Thanks

PR #1113, opened 2018-09-19
[add] support for accessibilityRole and accessibilityStates

### LINKED ISSUE #1112: New accessibility props

React Native 0.57 has added support for `accessibilityRole`, `accessibilityStates`, and `accessibilityHint`. https://facebook.github.io/react-native/blog/2018/08/13/react-native-accessibility-updates

Map `accessibilityStates` to ARIA states like `aria-disabled`, `aria-pressed`, and `aria-selected`. Investigate other states that may fit here.

Implement `accessibilityHint` as it is done on Android: append the hint text to the `accessibilityLabel`. Consider exposing an option (e.g., on `AccessibilityInfo`) to disable hints on web, as unlike iOS (and like Android) there is no system-level option to disable hints.

**Additional context**

The current RN implementation of these props crashes on Android when the value isn't recognized, and we should fix RN so that doesn't happen as web will have more roles and states than the shared native ones. I think it's important to allow the web to support more than the limited subset of roles RN currently supports.

necolas:
React Native 0.57 introduced 'accessibilityRole' and 'accessibilityStates' as cross-platform accessibility APIs to replace 'accessibilityComponentType' and 'accessibilityTraits' for Android and iOS.

React Native for Web has supported the 'accessibilityRole' for a while. This patch maps some of the values defined in React Native to web equivalents, and continues to allow a larger selection of roles for web apps. It also adds support for 'accessibilityStates', mapping values to ARIA states with boolean values and expanding support beyond 'disabled' and 'selected'.

Fix #1112


packages/react-native-web/src/modules/AccessibilityUtil/propsToAriaRole.js
```diff
@@ -23,6 +23,22 @@ const accessibilityTraitsToRole = {
   summary: 'region'
 };
 
+// 'false' value indicates role doesn't map to ARIA role
```
necolas: change to `null`

packages/react-native-web/src/modules/createDOMProps/index.js
```diff
@@ -103,7 +104,12 @@ const createDOMProps = (component, props, styleResolver) => {
   if (accessibilityLiveRegion && accessibilityLiveRegion.constructor === String) {
     domProps['aria-live'] = accessibilityLiveRegion === 'none' ? 'off' : accessibilityLiveRegion;
   }
-  if (role && role.constructor === String && role !== 'label') {
+  if (Array.isArray(accessibilityStates)) {
+    for (let i = 0; i < accessibilityStates.length; i += 1) {
+      domProps[`aria-${accessibilityStates[i]}`] = true;
```
necolas: This is only safe while all the values for RN accessibility states correspond exactly to aria state names

packages/react-native-web/src/modules/AccessibilityUtil/isDisabled.js
```diff
@@ -7,6 +7,10 @@
  * @flow
  */
 
-const isDisabled = (props: Object) => props.disabled || props['aria-disabled'];
+const isDisabled = (props: Object) =>
+  props.disabled ||
+  (Array.isArray(props.accessibilityStates) &&
+    props.accessibilityStates.indexOf('disabled') > -1) ||
+  props['aria-disabled'];
```
necolas: Drop this check. Add comment to release notes that any use of `aria-disabled` needs to move to `accessibilityStates`

necolas:
Merged into `react-native-0.57` 750790cee942d5fed49c3325bc7b5971e2d04e7f

PR #1122, opened 2018-09-30
Fix sticky header on virtualized lists

### LINKED ISSUE #1066: SectionList section heads are not sticky
**The problem**

Sticky headers not working in `SectionList`

**How to reproduce**

Simplified test case: https://codesandbox.io/s/v3x908zkw7?module=%2Fsrc%2FApp.js

Steps to reproduce:
1. Create a SectionList with section
2. add `stickySectionHeadersEnabled` (should this be default? it is on iOS)
3. Scroll SectionList

**Expected behavior**

Headers stick to the top of the list, are replaced by next header down.

* React Native for Web (version): 0.8.9
* React (version): 16.4
* Browser: Safari, Chrome

brunolemos:
Fix #1066 

I investigated why sticky headers wasn't working and found out the `style` prop was never used by the default `CellRendererComponent`.

_I don't know how your process of patching vendor code is, so I just changed it directly._

Another idea is passing a default `CellRendererComponent` prop to VirtualizedList that handles this style merging correctly, so vendor code won't need to be changed.

### Reproduction

- CodeSandbox (https://codesandbox.io/s/8yoolj322j)
- Snack (https://snack.expo.io/@brunolemos/sticky-headers)

### DIFF
--- a/packages/react-native-web/src/vendor/react-native/VirtualizedList/index.js
+++ b/packages/react-native-web/src/vendor/react-native/VirtualizedList/index.js
@@ -1562,6 +1562,7 @@ class CellRenderer extends React.Component<
       renderItem: renderItemType,
     },
     prevCellKey: ?string,
+    style: ?DangerouslyImpreciseStyleProp,
   },
   $FlowFixMeState,
 > {
@@ -1630,6 +1631,7 @@ class CellRenderer extends React.Component<
       index,
       inversionStyle,
       parentProps,
+      style,
     } = this.props;
     const {renderItem, getItemLayout} = parentProps;
     invariant(renderItem, 'no renderItem!');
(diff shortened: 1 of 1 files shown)


necolas:
Nice one. It might be that this was patched in RN by now. I'll see where this end up when I upgrade the vendor code to 0.57

brunolemos:
Not patched on RN because it is not a bug there, they have the prop `CellRendererComponent` to customize this, but rnw uses cloneElement

necolas:
Uses cloneElement where?

brunolemos:
On ScrollView, to add sticky header: https://github.com/necolas/react-native-web/blob/1f3a77dadad26c09dc3d854f07ffaf4f23f2417c/packages/react-native-web/src/exports/ScrollView/index.js#L167-L176

necolas:
I'm a bit confused. Why does the vendor code need to change? Are you saying CellRendererComponent does different things on different platforms?

brunolemos:
> Are you saying CellRendererComponent does different things on different platforms?

No

> Why does the vendor code need to change?

I see two ways of fixing the issue:

1. Patching `VirtualizedList` to fix the implementation of the default `CellRendererComponent` component, which is what I did on this PR
2. Patching `FlatList` and `SectionList` to edit the line `<VirtualizedList ... />` to `<<VirtualizedList CellRendererComponent={RNWCellRendererComponent] ... />` where `RNWCellRendererComponent` handles the `style` merging correctly

Let me know if you see other ways

necolas:
I see, thanks for explaining. Yeah maybe this patch can go into RN too. I'll look into it but this looks good for now :)

brunolemos:
Ok, thanks.

Issue #735, opened 2017-12-19
Interop between bs-react-native and react-native-web


jbaxleyiii:
This is a mix of a feature / bug, but I'd lean more towards it being a feature. When using RNW with bucklescript RN bindings, the bucklescript library passes down all props from its internal mapping. Would it be possible (happy to open a PR) to extract all unused props and not spread those into dom components?

Conversation started https://github.com/reasonml-community/bs-react-native/issues/122

MoOx:
Precision: 

> when using RNW with ~~Reason~~ _buklescript **RN**_ bindings

:)

I am also interested and will be more than happy to help (I am starting an reason/rn/rnw website/app this days and have tons of annoying warnings I would like to get ride of)

necolas:
Perhaps another option would be to stop passing unknown props to the underlying DOM node? At the moment, every time RN adds a new native-only prop RNW needs to strip it from the props being passed down.

I'm not sure how much people are relying on props being passed down, or if it would be confusing that nothing happens when you supply unknown props. Thinking out load, `className` could continue to be passed to the DOM node (if only to make migrating existing web apps easier), but any web-specific composite components built with `View` and relying on ARIA props for accessibility would need to use `setNativeProps`.

Iamsamwhoami:
It's actually the same case with `styled-components`, you pass a prop which you use to change styles and that is then passed down to the RNW View which causes an error. So not passing down unknown props to the DOM element would be nice!

<img width="1463" alt="screen shot 2017-12-19 at 23 35 35" src="https://user-images.githubusercontent.com/5459507/34182153-618e4884-e515-11e7-9444-5ad43a2909d4.png">

necolas:
That appears to be a design flaw in the styled-components API. I don't know if they fixed it yet

necolas:
Happy to review a PR to remove more props

PR #1123, opened 2018-09-30
[fix] ignore native-only TextInput props

rickiesmooth:
Hi this is my first PR. I've looked at a similar [PR](https://github.com/necolas/react-native-web/pull/753) and I hope everything is alright.
Ref #735

### DIFF
--- a/packages/react-native-web/src/exports/TextInput/index.js
+++ b/packages/react-native-web/src/exports/TextInput/index.js
@@ -210,6 +210,14 @@ class TextInput extends Component<*> {
       selectionState,
       textBreakStrategy,
       underlineColorAndroid,
+      hitSlop,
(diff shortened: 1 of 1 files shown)


necolas:
Is hitslop meant to work with text input?

rickiesmooth:
I'm not sure but it was passed down to the dom and I thought this would prevent that.

rickiesmooth:
It's getting hitslop from `...ViewPropTypes` but I couldn't find any reference to it in the text input.

necolas:
I checked and `hitSlop` didn't do anything for `TextInput` in React Native

necolas:
Thanks for this

rickiesmooth:
Thanks for checking and for RNW!

PR #1129, opened 2018-10-09
Don't ResizeObserve elements without onLayout props

### LINKED ISSUE #1128: Every RNW Node is being ResizeObserved
**The problem**
In `applyLayout`, the intent is to leverage ResizeObserver to power onLayout updating. Looking at the `componentDidUpdate` override, it's clear that the observation should be added or removed with `applyLayout`. However, looking at the `componentDidMount`, every single node is being observed, regardless of its props:
https://github.com/necolas/react-native-web/blob/af80b046fab34bcf49ed511b2d22e7289bdc1e01/packages/react-native-web/src/modules/applyLayout/index.js#L97

Coupled with large projects and the suggested `ResizeObserver` shim, this can tank performance in non-chrome browsers: the MutationObserver is attached at the window level and observers are accumulated. Any change, any where in the DOM will trigger a loop over all observers. In order to determine if a resize should be fired back to the consumer (RNW), it attempts to measure the size of each observed element.

TL;DR: In non-chrome browsers, every RNW element in the DOM is measured on every DOM change.

**How to reproduce**
https://codesandbox.io/s/mo9xn41y9x

Observe the console using the provided ResizeObserver mock

**Expected behavior**
Listener should only be attached when needed

* React Native for Web (version): Master
* React (version): 16.5
* Browser: Non-Chrome (Chrome supports ResizeObserver and isn't as bad since it's tightly scoped)

comp615:
Fixes #1128

### DIFF
--- a/packages/react-native-web/src/modules/applyLayout/index.js
+++ b/packages/react-native-web/src/modules/applyLayout/index.js
@@ -97,7 +97,7 @@ const applyLayout = Component => {
     function componentDidMount() {
       this._layoutState = emptyObject;
       this._isMounted = true;
-      observe(this);
+      this.props.onLayout && observe(this);
     }
   );
 


necolas:
Thanks

Issue #1116, opened 2018-09-20
Accessibility: working with ARIA labelledby and describedby properties


comp615:
Similar to #1006

**Is your feature request related to a problem? Please describe.**
Since RNW 0.8, many of our accessibility areas are broken as we relied on id references to throw certain labels and references across components.

**Describe a solution you'd like**
There are many instances when dealing with accessibility that we are required to reference dom ids. A few examples:

- aria-describedby has no alternative to using dom id (i.e. there is no aria-description)
- Typeahead or "combobox" elements need to be able to reference the id of the currently active object
- Often labels may need to span distances across components in certain layouts where there can be a complexity to trying to use aria-label. Consider a general Modal Routing component. It does not care about what's rendered within the modal, but it needs a label that's based on what is rendered inside itself. One option is upwards props (remote-control), but another is to simply have a static domId ("modal-label") that both components agree to use. 

Due to this, we'd like to bring back one of `id`, `domId` or `accessibilityId` as options for setting the dom id. These are a necessary side-effect of web development for accessibility. Naming it accessiblityId may help align it with testId and clarify that it's ONLY intended for use as an accessibility prop and not to ruin good component design. 

**Describe alternatives you've considered**
As mentioned above, there are sometimes workarounds. A good example is #1112 which aims to implement the aria describedby prop (dom id only), but to avoid using domIds (and elements) appends it to the label instead. Labels themselves can be specified directly via remote props or different organization of code. Many times though, this requires duplicating logic: whereas we could reference the various pieces of a tweet for a screen reader via describedby to reorder the spoken output, if only labels are employed, it becomes difficult to internationalize and properly mix all the output into a single field.

For many aria tasks, there is no viable alternative: a combobox (typeahead) cannot be created without domIds.

**Additional context**
Combobox example:
https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html

Thanks! Happy to chat more about this or author a PR.

necolas:
I guess relying on those props is something recent and you're having problems because of your long upgrade cycles. You can use `createElement` or `setNativeProps` for now

necolas:
[...]

PR #1130, opened 2018-10-10
[add] View prop nativeID

necolas:
Maps the View prop 'nativeID' to DOM 'id' as these are equivalent.
Enables declarative use of various 'aria-*' properties that require ID
references.

Ref #1116

cc @comp615 for review


comp615:
Ack lgtm, thanks!

PR #1134, opened 2018-10-11
Fix memory leak in applyLayout

### LINKED ISSUE #1133: Using onLayout causes memory leak
**The problem**
Running through memory traces of Twitter, one large source of held memory was Views being held by the resize observer. In poking around in this area, I see that there's a mapping of ids to component instances that's held by `applyLayout` so that when Resize observer calls back, it can notify the component instance. However there's a mismatch in the mapping logic.

When observing, we either assign the generated id to the node if it exists, or the instance if not:
https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/modules/applyLayout/index.js#L49-L63

When unobserving however, we always assume the id lays on the instance.
https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/modules/applyLayout/index.js#L65-L76

This means that when removing, we do not remove the component instance from the map and it's therefore retained by JS even after unmounting.

**How to reproduce**
I don't have a good test case / link here as determining this requires changing RNW itself to log or doing heap traces. One way to verify this is to alter RNW to log `registry[instance._layoutId]` in the `unobserve` function. You'll notice it's undefined.

**Expected behavior**
The applyLayout method should allow components to fully clean up after themselves.

* React Native for Web (version): > 0.9.0
* React (version): Any
* Browser: Any

**Fix**
I've attached a sample fix, but there's many ways of getting similar behavior.

comp615:
Fixes #1133.

There's other ways of addressing this too (always keep id on node or instance), but wasn't sure of the history behind the decision so here's a starter fix :)

### DIFF
--- a/packages/react-native-web/src/modules/applyLayout/index.js
+++ b/packages/react-native-web/src/modules/applyLayout/index.js
@@ -63,15 +63,15 @@ const observe = instance => {
 };
 
 const unobserve = instance => {
-  const node = findNodeHandle(instance);
-  const registryId = node ? node._layoutId : instance._layoutId;
-  delete registry[registryId];
   if (resizeObserver) {
+    const node = findNodeHandle(instance);
     if (node) {
+      delete registry[node._layoutId];
       delete node._layoutId;
       resizeObserver.unobserve(node);
     }
   } else {
+    delete registry[instance._layoutId];
     delete instance._layoutId;
   }
 };
--- a/packages/react-native-web/src/modules/applyLayout/index.js
+++ b/packages/react-native-web/src/modules/applyLayout/index.js
@@ -63,9 +63,10 @@ const observe = instance => {
 };
 
 const unobserve = instance => {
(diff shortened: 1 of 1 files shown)


necolas:
Good spot. Let's move the registry cleanup into each branch of the `if` statement to avoid the `findNodeHandle` call for the non-observer path.

I think the IDs are stored in 2 different ways because the observer returns DOM nodes that are then mapped to their backing instance, whereas the fallback only needs to tell instances to recalculate layout after the window resizes. It's certainly a bit confusing because the 2 models are quite different. Hopefully the other browsers implement ResizeObserver soon (😭) and we can clean this up

comp615:
done!

giuseppeg:
> I think the IDs are stored in 2 different ways because the observer returns DOM nodes that are then mapped to their backing instance

Correct, you could have added the id to the instance in both cases and keep `unobserve` as-is but this works too! ✌️

Issue #995, opened 2018-06-13
scrolling direction when using inverted prop


tafelito:
**The problem**
When using FlatList with inverted prop, the mouse scrolling direction is also inverted. I'm not sure if this is an issue per se, but I think only the content of the list should be inverted and not the direction of the scroll. I worked on implementation for the same feature for the [RLV](https://github.com/Flipkart/recyclerlistview) and I tried to handle this on the onScroll event reversing the scroll direction. This only happens on desktops, mobiles scroll is different so it works fine there

**How to reproduce**
Just using the FlatList with inverted={true}

**Expected behavior**
Invert list content but keep mouse scrolling direction unchanged

**Environment (include versions). Did this work in previous versions?**

* OS:
* Device:
* Browser: Chrome, Safari
* React Native for Web (version):^0.8.3
* React (version): 16.4

necolas:
> Is this something that could be done here as well?

Sounds like a good idea

necolas:
I'm open to a PR if you feel like it. No worries if not. Off the top of my head, I'm not sure where the best place for this to go is but can help figure that out if you're going to try

tafelito:
I'd be happy to try. I have to figure it out how to run RNW locally first and then any help would be great

necolas:
Cool. This setup guide will get you started https://github.com/necolas/react-native-web/blob/master/.github/CONTRIBUTING.md

tafelito:
So I saw that I can add the onWheel prop to the scrollProps in the VirtualizedList and I could call `this._scrollRef.scrollTo` setting y or x to the delta but it doesn't work.
I also tried a different approach but didn't work either. Instead of scaling the elements with -1, I used flexDirection to reverse the scrollview entirely but even though the items are reversed, they are still mounted in the same order, so you'll see how the items are pushed down while loading. So it's not a solution.

mglagola:
@tafelito did you ever find a solution?

tafelito:
I haven't had more time to spend on this unfortunately neither or the useWindow scroll feature

tafelito:
@necolas are you using FlatList at Twitter Lite?  Because I saw that the messages window uses an inverted scroll and the direction works fine there. 

Also I noticed that the home page uses some sort of window scroller but on the chat screen you can only scroll on the list and not outside

necolas:
No. Twitter PWA uses a custom scroller that isn't integrated with React Native at all, i.e., no support for the responder system etc

[...]

PR #1137, opened 2018-10-12
FlatList Inverted ScrollWheel Fix

tjbaron:
Fix for #995. Makes scroll wheel go in correct direction when FlatList is inverted. Note that desktop safari won't "bounce" with this implementation.

`wheelDeltaY` is for Safari and Chrome. `deltaY` is for Firefox.

### DIFF
--- a/packages/react-native-web/src/vendor/react-native/VirtualizedList/index.js
+++ b/packages/react-native-web/src/vendor/react-native/VirtualizedList/index.js
@@ -598,12 +598,7 @@ class VirtualizedList extends React.PureComponent<Props, State> {
       });
     }
     if (this.props.inverted) {
-      const scrollNode = this.getScrollableNode();
-      scrollNode.addEventListener("wheel", (e) => {
-        if (this.props.horizontal) scrollNode.scrollLeft += (e.wheelDeltaX || -e.deltaX);
-        else scrollNode.scrollTop += (e.wheelDeltaY || -e.deltaY);
-        e.preventDefault();
-      });
+      this._invertScrolling(true);
     }
   }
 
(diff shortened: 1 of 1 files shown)


packages/react-native-web/src/vendor/react-native/VirtualizedList/index.js
```diff
@@ -937,6 +956,11 @@ class VirtualizedList extends React.PureComponent<Props, State> {
       });
     }
     this._scheduleCellsToRenderUpdate();
+
+    if (prevProps.inverted !== inverted) {
```
necolas: Looks like this should be run on unmount too

packages/react-native-web/src/vendor/react-native/VirtualizedList/index.js
```diff
@@ -631,6 +634,22 @@ class VirtualizedList extends React.PureComponent<Props, State> {
...
+    const scrollNode = this.getScrollableNode();
+    if (invert) {
+      scrollNode.addEventListener("wheel", this._invertedWheelEvent);
+    } else {
+      scrollNode.removeEventListener("wheel", this._invertedWheelEvent);
+    }
+  }
+
+  _invertedWheelEvent = (e) => {
+    const scrollNode = this.getScrollableNode();
```
necolas: We could avoid another call to get the node by passing it from the first site that gets the node

packages/react-native-web/src/vendor/react-native/VirtualizedList/index.js
```diff
@@ -597,6 +597,9 @@ class VirtualizedList extends React.PureComponent<Props, State> {
         ref: this,
       });
     }
+    if (this.props.inverted) {
+      this._invertScrolling(true);
```
necolas: My thought is that this should probably be done in the platform ScrollView, rather than this component. I think this list passes all its props to ScrollView, even though ScrollView doesn't technical support that prop

raphaelrk:
Was running into this, came up with this hacky solution in my component with the flatlist until a fix is made:

```
  componentDidMount() {
    this._invertScrolling(true);
  }

  componentWillUnmount() {
    this._invertScrolling(false);
  }

  getScrollableNode = () => this._flatlistRef.getScrollableNode();

  _invertScrolling = (invert) => {
    this._scrollNode = this.getScrollableNode();
    if (invert) {
      this._scrollNode.addEventListener('wheel', this._invertedWheelEvent);
    } else {
      this._scrollNode.removeEventListener('wheel', this._invertedWheelEvent);
    }
  };

  _invertedWheelEvent = (e) => {
    if (this.props.horizontal) this._scrollNode.scrollLeft += (e.wheelDeltaX || -e.deltaX);
    else this._scrollNode.scrollTop += (e.wheelDeltaY || -e.deltaY);
    e.preventDefault();
  };
```

Unfortunately results in very janky scrolling (which I presume would also happen with this PR) which doesn't feel great on the chat screen I'm building.

onEndReached also isn't being called when I have inverted={true} and this does not fix it, but maybe that should be in a separate issue.

Wondering if anyone has any escape hatch solutions, e.g. dropping down to normal react + using something like react-window + using some css to anchor the messages to the bottom of the list.

danalloway:
Trying to move this forward in regards to the feedback and original attempt I took a stab at it over in #1241

rodrigorm:
The patch on this PR work in my app. Great job!

necolas:
Closing this PR due to inactivity and it not adequately solving the issue. Thanks for having a go

PR #1145, opened 2018-10-19
[bug] `process.env.NODE` -> `process.env.NODE_ENV`

koulmomo:
Build started failing for me with `Uncaught ReferenceError: process is not defined` when importing `LayoutAnimation`.

My build system only supports `process.env.NODE_ENV`. 

Was going to alias on my end, but checked and it seems everyone uses `process.env.NODE_ENV` (including the rest of this codebase).

Assuming it's a typo.


necolas:
Yes that's a typo, thanks for catching it!

PR #1147, opened 2018-10-22
Add NativeId to View supported props filter

comp615:
Follows up on #1116 and #1130 to let nativeId pass through on View


necolas:
Thank you. Sorry, I meant to get to this much sooner but got caught up with other things.

PR #1150, opened 2018-10-22
Cache Clipboard.isAvailable() value

### LINKED ISSUE #1149: Clipboard API value should be cached
Trying to dig into some performance, when we render a tweet action, we check if Clipboard.isAvailable() to know if we should show a button for that or not. While on the surface this seems fine, in tracing, it looks like these calls often force a style calculation in Chrome.

image

TBH, I don't fully understand why `queryCommandSupported` is causing that to happen. I tried to make a small repro, but it wasn't complex enough to trigger it. It may be because work is happening between calls to alter the styles in other components.

In any case, caching this value seems like it wouldn't hurt too much and if it helps prevent style recalcs, that might be the most minor of wins. Every little bit counts!

The user of RNW could also check and cache the value locally, but doing it in RNW makes it performant by default for all.

comp615:
Fixes #1149

### DIFF
--- a/packages/react-native-web/src/exports/Clipboard/index.js
+++ b/packages/react-native-web/src/exports/Clipboard/index.js
@@ -8,11 +8,15 @@
  * @flow
  */
 
+let clipboardAvailable;
+
 export default class Clipboard {
   static isAvailable() {
-    return (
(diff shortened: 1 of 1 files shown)


necolas:
Thanks, this will be part of 0.9.4 in a bit

PR #1157, opened 2018-10-31
perf: lazy call with locationX/locationY

hushicai:
`locationX/locationY` should be called lazy.

In most times, they are not used.

`locationX/locationY` was read in `Touchable/index.js`, but make no sense.

Reading `lcoationX/locationY` will cause `Layout` issue, when I touch on a Touchable element, and swipe to scroll in a FlatList, scroll performance is very bad:

image

As we can see, `getRect` cause `Force Synchronous Layout`, which costs up to **700ms+** in my demo!

we can call them lazy, and avoid unnecessary layout！

### DIFF
--- a/packages/react-native-web/src/exports/Touchable/index.js
+++ b/packages/react-native-web/src/exports/Touchable/index.js
@@ -696,9 +696,16 @@ const TouchableMixin = {
     const touch = TouchEventUtils.extractSingleTouch(e.nativeEvent);
     const pageX = touch && touch.pageX;
     const pageY = touch && touch.pageY;
-    const locationX = touch && touch.locationX;
-    const locationY = touch && touch.locationY;
-    this.pressInLocation = { pageX, pageY, locationX, locationY };
+    this.pressInLocation = {
+      pageX,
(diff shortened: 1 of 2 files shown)


necolas:
Thanks! Do you have a trace of your demo after you made this change (for comparison)?

hushicai:
After this change, `getRect` in `normalizeNativeEvent` will never be called! 

But there are another issue:

`_remeasureMetricsOnActivation` in `Touchable` will also cause `Layout`, which also make scroll slow.

image

I think this is another problem, and i am still looking for solution.

necolas:
Is `_remeasureMetricsOnActivation` being called on touchables when scrolling?

hushicai:
yes, it did.

image

image

hushicai:
just before scrolling, at the moment you touch on a touchable, and swipe to scroll.

necolas:
Hmm ok. We should find out if it is called on native too. 

The `getRect` code may benefit from switching to `getBoundingClientRect` anyway.

I'll merge this fix first

PR #1158, opened 2018-10-31
Components should be css-agnostic

Insomniak313:
In order to be more extensible, components should be as abstract as possible.
Writing component this way simplify the style overwriting.
I also prefer the use of { children } to make it less specific to the button content (so buttons need to be <Button><Text>Test</Text></Button>)


Insomniak313:
I see, it should match the same footprint as in React Native ? Why do you plan to create to replace it ? My I help ?

necolas:
I don't think there's a plan to replace it. You can make your own buttons from the other building blocks

PR #1159, opened 2018-10-31
Add style-sheet to benchmarks

giuseppeg:
This is an experimental CSS in JS lib I recently wrote. It supports extraction to file but it is still not as fast as I would like it to be :)

### DIFF
--- a/packages/benchmarks/index.html
+++ b/packages/benchmarks/index.html
@@ -8,6 +8,7 @@
     html, body { height: 100%; width: 100%; overflow: hidden; }
     .root { height: 100%; overflow: hidden; }
   </style>
+  <link id="__style_sheet_extracted__" rel="stylesheet" href="./style-sheet-bundle.css">
 </head>
 <body>
   <div class="root"></div>
--- a/packages/benchmarks/package.json
+++ b/packages/benchmarks/package.json
@@ -20,6 +20,7 @@
     "react-jss": "^8.6.1",
     "react-native-web": "0.9.3",
     "reactxp": "^1.3.0",
+    "style-sheet": "^3.0.1",
     "styled-components": "^3.3.3",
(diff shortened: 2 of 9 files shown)


necolas:
Feel free to develop against these tests if they're helpful but I don't think I want to start adding lots of style libraries or experiment to this repo. Thanks

giuseppeg:
No problem. Indeed I mostly use it to benchmark/perf tuning. 

Unrelated: you wrote that it is bad idea to support states, shallow selectors like `:hover > &` and media queries and I get why. From my experience when you sort them by specificity you can still resolve styles deterministically (at least with my library). Is there any other reason why states and shallow selectors might be a worst solution compared to explicitly toggling/controlling classes with JS? Fine tuning/resolution maybe?

necolas:
Yes :hover produces undesired UI changes on mobile web, ancester hover states get triggered by hovering children, and can't delay hover reactions which makes the CSS state useless for creating more complex and accessible hover interactions

necolas:
Re: https://mobile.twitter.com/giuseppegurgone/status/1076513890735927301

I'm curious where the runtime overhead lies here if you're using static extraction. Or is this screenshot not from an example using extraction?

giuseppeg:
Not sure I need to take a look (I haven't had the time to spend time on perf tuning or the lib in general). One reason could be that styles resolution is done at runtime. Basically `StyleSheet.create` converts objects to arrays of class names where the first part is the property name and at runtime I merge right to left using this information (similarly to how `Object.assign` works). I would like to precompile the static ones with the Babel plugin at some point.

The other reason could be the one mentioned in the tweet i.e. too many classes which have an impact on layout.

I am attaching a dumb of the recording: [style-sheet-profile.txt](https://github.com/necolas/react-native-web/files/2705568/style-sheet-profile.txt)

PR #1167, opened 2018-11-06
[add] scripts version for web support to work

ahmedu007:
`--with-web-support` flag does not work on its own unless the scripts version is also passed into it.

### DIFF
--- a/docs/guides/getting-started.md
+++ b/docs/guides/getting-started.md
@@ -35,7 +35,7 @@ Multi-platform: [create-react-native-app](https://github.com/react-community/cre
 includes experimental support for Web.
 
 ```
-create-react-native-app my-app --with-web-support
+create-react-native-app my-app --with-web-support --scripts-version 1.2.0-alpha.4
 ```
 
 ## Configuring a module bundler
@@ -56,7 +56,7 @@ module.exports = {
       'react-native$': 'react-native-web'
     }
   }
-}
+};
 ```
 
 Now you can create your components and applications with the React Native API.


necolas:
I see that expo archived CRNA and the expo-cli doesn't have any documentation for web support. So I think it's probably best to remove this from the RNW docs until it's clear

/cc @brentvatne

ahmedu007:
Its your call. Although I still find it to be the most effective way to get started with the RNW. I can update the Pull request to remove it from the docs if that's what is best.

brentvatne:
i'm going to be working on this again in the coming months but for now there is no way to create an app with expo-cli that has rnw supports built in. i don't think we want to suggest a deprecated approach so i would vote for removing this

comp615:
FWIW, also tried figuring out how to do this last night and wasn't able to :( Ended up trying to start from https://github.com/joefazz/react-native-web-starter, but it's surprisingly difficult to find how to make a project that does web in addition to native.

ahmedu007:
Yeah, that was essentially the reason for this pull request. RNW makes it awfully easy to start developing, and share code between platforms, and has awesome performance, but it took a long while trying to set it up and get running.

Do you think it would be a good idea to maintain a fork of the CRNA repo, atleast until expo officially starts supporting it? (i.e If they actually decide to do it)

brentvatne:
@ahmedu007 - it might be useful to maintain a react-native-scripts fork until i get around to working on this, yeah.

PR #1168, opened 2018-11-09
Fix modality performance issues

### LINKED ISSUE #1155: modality performance issue
**The problem**

Modality cause reflow/layout in chrome/safari.

And make the keyboard very slow to open in IOS.

**How to reproduce**

simple demo:

```
import React from 'react';
import {ScrollView, View, Text, StyleSheet, TextInput} from 'react-native';
import map from 'lodash/map';

const data = new Array(10000).fill(0);

class Test extends React.Component {
  render() {
    return (
      <View style={styles.container}>
        <ScrollView style={styles.scroller}>
          {map(data, (v, i) => {
            return (
              <View key={i + 1}>
                <Text>{i}</Text>
              </View>
            );
          })}
        </ScrollView>
        <View style={styles.footer}>
          <TextInput
            style={styles.input}
            placeholder={'enter...'}
            placeholderTextColor={'#ccc'}
          />
        </View>
      </View>
    );
  }
}

  container: {
    flex: 1
  },
  scroller: {
    flex: 1
  },
  footer: {
    height: 36
    // width: '100%',
    // overflow: 'hidden'
  },
  input: {
    flex: 1,
    height: 36
  }
});

export default Test;
```

before remove modality, chrome performance:

image

image

As we can see, `recalculate style + layout` costs up to 200ms, maybe more!!

after remove modality in `WebStyleSheet.js`, chrome performance:

image

Now just `recalculate style` happened, and it costs less than 1ms.

giuseppeg:
Fixes #1155

It seems that inserting and deleting the modality rule triggers styles recalculation which results in performance degradation on focus or, as the reporter suggests, makes the keyboard very slow to open in iOS.

This commit fixes the issue by using a more complex but permanent and performant rule, and toggling a scoped `rn-modality` class name on elements that get focus via keyboard.

Attached are two sample profiles:

* master 
[Modality-master.txt](https://github.com/necolas/react-native-web/files/2566035/Modality-master.txt)
* this branch 
[Modality-fix.txt](https://github.com/necolas/react-native-web/files/2566037/Modality-fix.txt)

### DIFF
--- a/packages/react-native-web/src/exports/StyleSheet/modality.js
+++ b/packages/react-native-web/src/exports/StyleSheet/modality.js
@@ -17,77 +17,46 @@
  */
 
 import { canUseDOM } from 'fbjs/lib/ExecutionEnvironment';
+import hash from '../../vendor/hash';
 
-const rule = ':focus { outline: none; }';
-let ruleExists = false;
+const modalityClassName =
+  'rn-' +
+  (process.env.NODE_ENV === 'production'
+    ? hash('rnw-modality')
+    : 'modality-' + hash('rnw-modality'));
+// These elements should always have a focus ring drawn, because they are
+// associated with switching to a keyboard modality.
+const keyboardModalityWhitelist = [
+  '[type="text"]',
+  '[type="search"]',
+  '[type="url"]',
+  '[type="tel"]',
+  '[type="email"]',
+  '[type="password"]',
+  '[type="number"]',
+  '[type="date"]',
+  '[type="month"]',
+  '[type="week"]',
+  '[type="time"]',
+  '[type="datetime"]',
+  '[type="datetime-local"]',
+  'textarea',
+  '[role="textbox"]'
+]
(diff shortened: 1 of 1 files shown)


packages/react-native-web/src/exports/StyleSheet/modality.js
```diff
@@ -17,78 +17,47 @@
...
+  '[type="time"]',
+  '[type="datetime"]',
+  '[type="datetime-local"]',
+  'textarea',
+  '[role="textbox"]'
+]
+  .map(s => `:not(${s})`)
+  .join('');
+
+const rule = `:focus:not(.${modalityClassName})${keyboardModalityWhitelist} { outline: none; }`;
```
giuseppeg: This is ugly af but it seems to perform better. It'd be great if you could test this because I changed the `keyboardModalityWhitelist` selectors a bit.

giuseppeg:
discarded in favor of #1169

PR #1169, opened 2018-11-09
Fix modality performance issues

### LINKED ISSUE #1155: modality performance issue

giuseppeg:
Fixes #1155

Inserting and deleting the modality rule triggers styles recalculation for the entire DOM tree which results in performance degradation on focus or, as the reporter suggests, makes the keyboard very slow to open in iOS.

This commit fixes the issue by integrating the [`:focus-visible` polyfill](https://github.com/WICG/focus-visible) which toggles a scoped `rn-focusVisible` class name on whitelisted elements or elements that get focus via keyboard. We insert a permanent rule to always disable the focus ring except for the elements that not match the class name i.e. `:focus:not(.rn-focusVisible)`. This performs much better since the browser needs to recalculate styles only for the focused element and its small subtree.

* master 
[Modality-master.txt](https://github.com/necolas/react-native-web/files/2569264/Modality-master2.txt)
* this branch 
[Modality-fixed-rev2.txt](https://github.com/necolas/react-native-web/files/2569265/Modality-fixed-rev2.txt)

### DIFF
--- a/packages/react-native-web/src/exports/StyleSheet/modality.js
+++ b/packages/react-native-web/src/exports/StyleSheet/modality.js
@@ -113,9 +113,6 @@ const modality = styleElement => {
    * originally added by the author.
    */
   function removeFocusVisibleClass(el) {
-    if (!el.classList.contains(focusVisibleClass)) {
-      return;
-    }
     el.classList.remove(focusVisibleClass);
   }
 
@@ -125,7 +122,7 @@ const modality = styleElement => {
    * of our keyboard modality state with `hadKeyboardEvent`.
    */
   function onKeyDown(e) {
-    if (e.metaKey || e.altKey || e.ctrlKey) {
+    if (e.metaKey || e.altKey || e.ctrlKey || e.shiftKey) {
       return;
     }
 
@@ -243,7 +240,7 @@ const modality = styleElement => {
   function onInitialPointerMove(e) {
     // Work around a Safari quirk that fires a mousemove on <html> whenever the
     // window blurs, even if you're tabbing out of the page. ¯\_(ツ)_/¯
-    if (e.target.nodeName.toLowerCase() === 'html') {
+    if (e.target.nodeName === 'HTML') {
       return;
     }
 
--- a/packages/react-native-web/src/exports/StyleSheet/modality.js
+++ b/packages/react-native-web/src/exports/StyleSheet/modality.js
@@ -125,6 +125,10 @@ const modality = styleElement => {
    * of our keyboard modality state with `hadKeyboardEvent`.
    */
   function onKeyDown(e) {
+    if (e.metaKey || e.altKey || e.ctrlKey) {
+      return;
+    }
+
     if (isValidFocusTarget(document.activeElement)) {
       addFocusVisibleClass(document.activeElement);
     }
--- a/packages/react-native-web/src/exports/StyleSheet/modality.js
+++ b/packages/react-native-web/src/exports/StyleSheet/modality.js
@@ -17,9 +17,15 @@
  */
 
 import { canUseDOM } from 'fbjs/lib/ExecutionEnvironment';
+import hash from '../../vendor/hash';
 
-const rule = ':focus { outline: none; }';
-let ruleExists = false;
+const modalityClassName =
+  'rn-' +
+  (process.env.NODE_ENV === 'production'
+    ? hash('rnw-modality')
+    : 'modality-' + hash('rnw-modality'));
+
+const rule = `:focus:not(.${modalityClassName}) { outline: none; }`;
 
 const modality = styleElement => {
   if (!canUseDOM) {
@@ -70,26 +76,6 @@ const modality = styleElement => {
   };
 
   /**
-   * Add the focus ring style
-   */
-  const addFocusRing = () => {
-    if (styleElement && ruleExists) {
-      styleElement.sheet.deleteRule(0);
-      ruleExists = false;
-    }
-  };
-
-  /**
-   * Remove the focus ring style
-   */
-  const removeFocusRing = () => {
-    if (styleElement && !ruleExists) {
-      styleElement.sheet.insertRule(rule, 0);
-      ruleExists = true;
-    }
-  };
-
-  /**
    * On `keydown`, set `hadKeyboardEvent`, to be removed 100ms later if there
    * are no further keyboard events. The 100ms throttle handles cases where
    * focus is redirected programmatically after a keyboard event, such as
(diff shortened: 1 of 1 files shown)


necolas:
Thanks for looking at this. This used feature to be implemented using a separate stylesheet that was enabled/disabled. Does that approach cause perf issues too?

Applying the class directly to the element is worth considering. And I think it's also worth looking into updating the implementation to pick up some of the improvements upstream - https://github.com/WICG/focus-visible - especially because this proposed approach creates the possibility that the class name isn't always removed from the element.

giuseppeg:
> Thanks for looking at this. This used feature to be implemented using a separate stylesheet that was enabled/disabled. Does that approach cause perf issues too?

No problem. I tried a few solutions including that one but I am afraid that it has similar performance as master. With a deep tree like the one from the test case in the issue description the browser has to recalculate styles for the entire path I guess. 

---

> And I think it's also worth looking into updating the implementation to pick up some of the improvements upstream - https://github.com/WICG/focus-visible 

Agreed! I didn't think of that proposal but it makes totally sense.

---

> especially because this proposed approach creates the possibility that the class name isn't always removed from the element.

I suppose you are referring to the following behavior (as described in the focus-visible repo):

> If there are elements which should always have a focus ring shown, authors may explicitly add the focus-visible class. If explicitly added, it will not be removed on blur.

We would add the modality class when `focusTriggersKeyboardModality(e.target) === true` and never remove it.

packages/react-native-web/src/exports/StyleSheet/modality.js
```diff
@@ -11,115 +11,252 @@
  * 1. a keydown event occurred immediately before a focus event;
  * 2. a focus event happened on an element which requires keyboard interaction (e.g., a text field);
  *
- * Based on https://github.com/WICG/focus-ring
+ * This software or document includes material copied from or derived from https://github.com/WICG/focus-visible.
```
giuseppeg: not sure how you deal with acknowledgment / copyright but this **might** be enough for this module.

packages/react-native-web/src/exports/StyleSheet/modality.js
```diff
@@ -11,115 +11,252 @@
...
+    if (tagName === 'INPUT' && inputTypesWhitelist[type] && !isReadOnly) {
+      return true;
     }
-  };
+
+    if (tagName === 'TEXTAREA' && !isReadOnly) {
+      return true;
+    }
+
+    if (el.isContentEditable) {
```
giuseppeg: On master you didn't whitelist `contenteditable`. I forgot to ask yesterday why / if there is any particular reason for that.

packages/react-native-web/src/exports/StyleSheet/modality.js
```diff
@@ -11,115 +11,252 @@
...
-  };
+    el.classList.add(focusVisibleClass);
+  }
+
+  /**
+   * Remove the `focus-visible` class from the given element if it was not
+   * originally added by the author.
+   */
+  function removeFocusVisibleClass(el) {
+    if (!el.classList.contains(focusVisibleClass)) {
```
giuseppeg: probably we could skip this check and call `classList.remove` right away (not sure which is faster)

necolas: Yeah let's just use `remove`, IIRC the native code does the `contains` check from `remove` anyway

packages/react-native-web/src/exports/StyleSheet/modality.js
necolas: No, just an omission

packages/react-native-web/src/exports/StyleSheet/modality.js
```diff
@@ -125,6 +125,10 @@ const modality = styleElement => {
    * of our keyboard modality state with `hadKeyboardEvent`.
    */
   function onKeyDown(e) {
+    if (e.metaKey || e.altKey || e.ctrlKey) {
```
necolas: Include `shiftKey` as well?

packages/react-native-web/src/exports/StyleSheet/modality.js
```diff
@@ -11,125 +11,256 @@
...
+  /**
+   * When the polfyill first loads, assume the user is in keyboard modality.
+   * If any event is received from a pointing device (e.g. mouse, pointer,
+   * touch), turn off keyboard modality.
+   * This accounts for situations where focus enters the page from the URL bar.
+   */
+  function onInitialPointerMove(e) {
+    // Work around a Safari quirk that fires a mousemove on <html> whenever the
+    // window blurs, even if you're tabbing out of the page. ¯\_(ツ)_/¯
+    if (e.target.nodeName.toLowerCase() === 'html') {
```
necolas: I think we can avoid the `toLowerCase` call and check for `HTML` as is done on L65

packages/react-native-web/src/exports/StyleSheet/modality.js
```diff
@@ -11,125 +11,253 @@
...
+    el.classList.remove(focusVisibleClass);
+  }
+
+  /**
+   * Treat `keydown` as a signal that the user is in keyboard modality.
+   * Apply `focus-visible` to any current active element and keep track
+   * of our keyboard modality state with `hadKeyboardEvent`.
+   */
+  function onKeyDown(e) {
+    if (e.metaKey || e.altKey || e.ctrlKey || e.shiftKey) {
```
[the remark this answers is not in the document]
giuseppeg: ouch right, thank you!

necolas:
I noticed a UI regression with this patch. When a component changes its styles in response to state changes, the className that is added here gets removed when React updates and the focus ring disappears. It looks like a similar issue came up in relation to usage with Vue https://github.com/WICG/focus-visible/issues/179

giuseppeg:
I think that React reconcile `className` but I probably didn't consider edge cases. I will make a patch to use a data attribute instead (I think that this won't work anyway when you have unstable/constant `key`)

PR #1175, opened 2018-11-14
fix: respect stickyHeaderIndices in ScrollView

jaulz:
See example https://codesandbox.io/s/zwx69r6nwl where `stickyHeaderIndices` do not work. The issue is that for rows in a `ListView` the `StaticRenderer` is used. The `StaticRenderer` does not pass on `styles` attribute.


jaulz:
See example https://codesandbox.io/s/zwx69r6nwl where `stickyHeaderIndices` do not work.

necolas:
Thanks!

PR #1182, opened 2018-11-19
Fix textShadow style

SKAhack:
This pull-req fixes the issue that textShadow does not work correctly when passing only blur.


necolas:
Please include descriptions in PRs.

This change to RN was relatively recent and I'll include it in 0.57 compatibility

SKAhack:
Thank you for your review. I updated this description.
I am not good at English, so you can edit it freely.

necolas:
Thanks, this has been applied to the react-native-0.57 branch

PR #1184, opened 2018-11-20
Add Scale3d as a proptype to transforms

comp615:
Much like translate3d, it's a valid prop too!


necolas:
If we're going to add this we should include `scaleZ` too.

(FWIW, neither `translateZ` nor `translate3d` are valid RN transform properties yet. And thinking about it, the `3d` values should probably be an array of 3 values but that doesn't need to be done here)

comp615:
Good idea on both counts. Agree it would be more clear as an array of numbers instead of the strings.

PR #1186, opened 2018-11-21
Fix regression in modality

giuseppeg:
Fixes https://github.com/necolas/react-native-web/pull/1169#issuecomment-440590544

### DIFF
--- a/packages/react-native-web/src/exports/StyleSheet/modality.js
+++ b/packages/react-native-web/src/exports/StyleSheet/modality.js
@@ -21,10 +21,12 @@
 import { canUseDOM } from 'fbjs/lib/ExecutionEnvironment';
 import hash from '../../vendor/hash';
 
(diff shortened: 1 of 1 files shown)


necolas:
Thanks, this works.

There's one more situation that the "polyfill" doesn't currently fix. You can keyboard-focus a focusable element that contains descendants and then click/touch a descendant (e.g., a button with text and icons). In this case, the focus ring will not be removed. Because the press happens on a descendant of the element that got the focus-visible attribute, checking whether `e.target` has the focus-visible attribute will always return `false` too.

One fix for this is to remove the attribute from any elements in the document if there's a pointer-down event after the keyboard has been used:

```js
  function onPointerDown(e) {
    if (hadKeyboardEvent === true) {
      removeAllFocusVisibleAttributes();
    }
    hadKeyboardEvent = false;
  }
```

PR #1191, opened 2018-11-28
Upgrade to Babel 7

### LINKED ISSUE #1170: Upgrade to Babel 7
Looking for someone to help upgrade the internal tools and the published babel plugin to rely on Babel 7.

brunolemos:
Some tests are failing:

```
Test Suites: 5 failed, 33 passed, 38 total
Tests:       32 failed, 316 passed, 348 total
Snapshots:   15 failed, 3 obsolete, 73 passed, 88 total
```

Might be related: https://github.com/facebook/jest/pull/7016, https://github.com/facebook/jest/issues/7118

Closes #1170


packages/babel-plugin-react-native-web/package.json
```diff
@@ -4,7 +4,8 @@
   "description": "Babel plugin for React Native for Web",
   "main": "index.js",
   "devDependencies": {
-    "babel-plugin-tester": "^5.0.0"
+    "@babel/core": "^7.1.6",
```
necolas: Why is `@babel/core` needed here?

package.json
```diff
@@ -49,8 +51,8 @@
     "flow-bin": "^0.63.1",
     "glob": "^7.1.2",
     "husky": "^0.14.3",
-    "jest": "^22.4.3",
-    "jest-canvas-mock": "^1.0.2",
+    "jest": "^23.6.0",
```
necolas: Jest needs to be upgraded to upgrade Babel?

package.json
```diff
@@ -25,18 +25,20 @@
...
-    "babel-preset-react": "^6.24.1",
-    "babel-preset-react-native": "^4.0.0",
+    "@babel/cli": "^7.1.5",
+    "@babel/core": "^7.1.6",
+    "@babel/plugin-proposal-class-properties": "^7.1.0",
+    "@babel/plugin-proposal-object-rest-spread": "^7.0.0",
+    "@babel/preset-env": "^7.1.6",
+    "@babel/preset-flow": "^7.0.0",
+    "@babel/preset-react": "^7.0.0",
+    "babel-core": "^7.0.0-bridge.0",
```
necolas: What is this for?

package.json
```diff
@@ -25,18 +25,20 @@
...
+    "babel-eslint": "^10.0.1",
+    "babel-jest": "^23.6.0",
```
necolas: Why is this needed now?

.flowconfig
```diff
@@ -6,7 +6,6 @@
 <PROJECT_ROOT>/packages/.*/dist/.*
 <PROJECT_ROOT>/packages/examples/.*
 <PROJECT_ROOT>/packages/website/.*
-.*/node_modules/babel-plugin-transform-react-remove-prop-types/*
```
necolas: Is this not needed anymore?

scripts/jest/config.js
```diff
@@ -15,4 +15,4 @@ module.exports = {
   snapshotSerializers: ['enzyme-to-json/serializer'],
   testEnvironment: 'jsdom',
   timers: 'fake'
-};
+};
```
necolas: Please revert this change

packages/babel-plugin-react-native-web/package.json
brunolemos: Not sure if it's required, it was automatically added by https://github.com/babel/babel-upgrade

package.json
brunolemos: No, I noticed jest was still installing the babel 6 dependencies so I upgraded, but it still was after the upgrade.

Reverted, but had to change a config var `testURL` to `'http://localhost/' because it was giving an error without this (ps: it's the default on newer jests).

package.json
brunolemos: This was also added automatically by https://github.com/babel/babel-upgrade 

It seems this causes babel to compile the code, but based on your comment that shouldn't happen.

But without this I get invalid syntax errors when it sees a `import`.

.flowconfig
brunolemos: I don't think so, but reverted

packages/babel-plugin-react-native-web/package.json
brunolemos: Removed

package.json
necolas: Might that be because you removed the babelrc file? Jest might have been using that before

brunolemos: The `.babelrc` was migrated to `babel.config.js`, this change was required to fix something.

package.json
colincclark: It is so that `babel-jest` can work

necolas:
Are you planning to fix the test failures?

packages/react-native-web/src/exports/TextInput/index.js
```diff
@@ -401,7 +401,7 @@ class TextInput extends Component<*> {
         onSubmitEditing(e);
       }
       if (shouldBlurOnSubmit) {
-        this.blur();
+        if (this.blur) this.blur();
```
brunolemos: The tests were failing with `_this.blur is not a function`, not sure what's the best way to fix is, so I changed this code.

brunolemos:
@necolas build is passing now.

Please make the remaining changes, if any, I'll not be available next few days.

necolas:
Thank you! Do you know which update fixed the AsyncStorage tests?

packages/babel-plugin-react-native-web/src/__tests__/__snapshots__/index-test.js.snap
```diff
@@ -7,11 +7,32 @@ export { ColorPropType, StyleSheet, Text, createElement } from 'react-native';
...
 
-export { default as View } from 'react-native-web/dist/exports/View';
-export { default as ColorPropType } from 'react-native-web/dist/exports/ColorPropType';
-export { default as StyleSheet } from 'react-native-web/dist/exports/StyleSheet';
-export { default as Text } from 'react-native-web/dist/exports/Text';
-export { default as createElement } from 'react-native-web/dist/exports/createElement';
+\\"use strict\\";
+
+exports.__esModule = true;
+exports.createElement = exports.Text = exports.StyleSheet = exports.ColorPropType = exports.View = void 0;
```
necolas: Ensure these snapshots are not changed. The commonjs flag is probably no longer needed here: https://github.com/necolas/react-native-web/blob/0.9.8/package.json#L18

packages/react-native-web/src/exports/TextInput/index.js
necolas: Get to the bottom of what causes this to fail

brunolemos:
> @necolas: Thank you! Do you know which update fixed the AsyncStorage tests?

Yes, it was caused by a new version of `jsdom`, `11.12.0` I think. It "freezes" the `window` object, so when you do `window.localStorage = mockXXX` in the test it doesn't have any effect.

package.json
```diff
@@ -25,17 +25,19 @@
...
+    "@babel/cli": "^7.0.0",
+    "@babel/core": "^7.0.0",
+    "@babel/plugin-proposal-class-properties": "^7.0.0",
+    "@babel/plugin-proposal-object-rest-spread": "^7.0.0",
+    "@babel/preset-env": "^7.0.0",
+    "@babel/preset-flow": "^7.0.0",
+    "@babel/preset-react": "^7.0.0",
+    "babel-core": "^7.0.0-bridge.0",
+    "babel-eslint": "^10.0.0",
+    "babel-jest": "^23.4.2",
```
SimenB: make sure to match jest major version here

hyochan:
When could this be merged? I am waiting for #1202 to test out. Maybe it isn't related but I think it is.

packages/babel-plugin-react-native-web/src/__tests__/__snapshots__/index-test.js.snap
necolas: This was being caused by the `babel-core` bridge dependency

necolas:
This is done in branch `react-native-0.57`

brunolemos:
Nice, would appreciate `git cherry-pick` being used next time

necolas:
Sorry in this case I didn't use this PR as the basis for the work as I had to change all the dependencies and fix the tests rather than commit incorrect an snapshot and leave the blur instance method broken

PR #1200, opened 2018-12-12
Add wordBreak style to Text Component

pruthvip:
Usage of wordBreak is throwing a warning. Hence added wordBreak style to Text Component
For reference: https://developer.mozilla.org/en-US/docs/Web/CSS/word-break


necolas:
Thanks. Merged into 0.10-wip

PR #1205, opened 2018-12-20
[change] Compile using Babel 7

necolas:
Problems with the unit tests when upgrading. Jest chokes on `import`.

`babel-plugin-tester` isn't compatible with Babel 7.


necolas:
This is done in branch `react-native-0.57`

PR #1206, opened 2018-12-20
Fix tests

matallo:


necolas:
Thanks but this doesn't fix the issue as it is simply using the DOM API directly

PR #1208, opened 2018-12-21
Use element.scroll method instead of scrollBehavior style property

### LINKED ISSUE #1203: Support polyfilling smooth scroll
Currently the `ScrollResponder` module is setting the `scrollBehavior` CSS property to `smooth` whenever doing an animated scroll: https://github.com/necolas/react-native-web/blob/c00360491b6f0658ec8f33c2f59349794646b27a/packages/react-native-web/src/modules/ScrollResponder/index.js#L378

The "scroll behavior" feature is not supported by some browsers (e.g. iOS Safari), but there is [a polyfill](https://github.com/iamdustan/smoothscroll). The polyfill does not have support for polyfilling the CSS property [because it seems hacky to add support for it](https://github.com/iamdustan/smoothscroll/issues/12#issuecomment-354547390), but works if the smooth scrolling is done with Javascript.

Would it be possible to detect & use the `element.scroll` method instead of a style inside `ScrollResponder`? If the method is not supported in some browsers there could be a fallback to the current implementation. Might be of course that there some other reasons why the style property is used currently.

I quickly tested the code below and it fixed the smooth scroll polyfilling, and smooth scrolling started working the same way in iOS Safari as on the native side:

`src/modules/ScrollResponder/index.js`

```diff
-UIManager.updateView(node, { style: { scrollBehavior: !animated ? 'auto' : 'smooth' } }, this);
-node.scrollLeft = x || 0;
-node.scrollTop = y || 0;
+node.scroll({ top: y || 0, left: x || 0, behavior: !animated ? "auto" : "smooth" });
```

@necolas

kristerkari:
Use `element.scroll` instead of `scrollBehavior` style property and `element.scrollLeft/element.scrollTop` to scroll.

This makes it possible to polyfill the smooth scroll behavior (https://github.com/iamdustan/smoothscroll).

I tried to find info about the browser support of `element.scroll`, but it seems to me that it would have the same support as `element.scrollIntoView` method, which means that all the browser that are supported by `react-native-web` support it (some without the support for `behavior: smooth`).

https://caniuse.com/#feat=scrollintoview

Closes #1203

### DIFF
--- a/packages/react-native-web/src/modules/ScrollResponder/index.js
+++ b/packages/react-native-web/src/modules/ScrollResponder/index.js
@@ -375,7 +375,12 @@ const ScrollResponderMixin = {
       ({ x, y, animated } = x || emptyObject);
     }
     const node = this.scrollResponderGetScrollableNode();
-    node.scroll({ top: y || 0, left: x || 0, behavior: !animated ? 'auto' : 'smooth' });
+    if (typeof node.scroll === 'function') {
+      node.scroll({ top: y || 0, left: x || 0, behavior: !animated ? 'auto' : 'smooth' });
+    } else {
+      node.scrollLeft = x || 0;
+      node.scrollTop = y || 0;
+    }
   },
 
   /**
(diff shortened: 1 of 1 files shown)


necolas:
It has [no support in Edge or Safari](https://developer.mozilla.org/en-US/docs/Web/API/Element/scroll), so we need to do a feature check before using `.scroll` otherwise fallback to the existing implementation (without the problematic smooth CSS)

kristerkari:
@necolas I added the feature check and fallback.

Issue #1057, opened 2018-07-31
ScrollView: support for pagingEnabled prop


nazarireza:
Hi, I'm using the version '0.8.9' of 'react-native-web' and 'pagingEnabled' prop in 'ScrollView' not work.
Any Idea?

Thank you

brunolemos:
For now you can do something like this:


```jsx
<ScrollView className="snap-container" horizontal pagingEnabled>
  <View className="snap-item-start">{/* ... */}</View>
  <View className="snap-item-start">{/* ... */}</View>
</ScrollView>
```

`index.css`
```
.snap-container {
  scroll-snap-type: x mandatory;
}

.snap-item-start {
  scroll-snap-align: start;
}
```

hushicai:
Hi, all, is there any new idea for `pagingEnabled`?

without `pagingEnabled`, swiper like `react-native-swiper` can not work in the browser。

kristerkari:
@necolas Would you be willing to accept a PR which adds the CSS properties from @brunolemos example when `pagingEnabled` is set to true?

At least it seems to me that it works exactly the same way as `pagingEnabled` on the native side. Of course it might be that there are some cases where it would work in a differently from the native implementation.

The browser support is only Safari, iOS Safari, Chrome and Android Chrome. Edge and Firefox only support an older version of the spec which does not have the `scroll-snap-align` property that is needed. There is also [polyfill](https://github.com/PureCarsLabs/css-scroll-snap-polyfill) for CSS scroll snap, but at least I could not get it to work.

kristerkari:
Here's the PR:
https://github.com/necolas/react-native-web/pull/1212

PR #1212, opened 2018-12-22
Add support for ScrollView pagingEnabled property

kristerkari:
Add support for ScrollView's pagingEnabled property using [CSS scroll snap](https://www.w3.org/TR/css-scroll-snap-1/)

Related issue #1057

Links:
- caniuse: https://caniuse.com/#feat=css-snappoints
- spec: https://www.w3.org/TR/css-scroll-snap-1/

### DIFF
--- a/packages/react-native-web/src/exports/ScrollView/index.js
+++ b/packages/react-native-web/src/exports/ScrollView/index.js
@@ -136,11 +136,10 @@ const ScrollView = createReactClass({
       horizontal,
       onContentSizeChange,
       refreshControl,
-      stickyHeaderIndices,
+      pagingEnabled,
       /* eslint-disable */
       keyboardDismissMode,
       onScroll,
-      pagingEnabled,
       /* eslint-enable */
       ...other
     } = this.props;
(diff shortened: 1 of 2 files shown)


necolas:
Cloning the children won't work properly for the same reasons it didn't for sticky headers. We should combine those two into a single function that adds the wrapping view, if possible

kristerkari:
> Cloning the children won't work properly for the same reasons it didn't for sticky headers

What was the issue? 

If I add a wrapping view, which styles should it have when `pagingEnabled` is `true`, so that it does not breaking the styles from each children?

necolas:
The RN list views do not all forward style to the children. So the element you're cloning may ignore the style prop. Look at the sticky headers function you touched. The paging stuff can probably go in there

kristerkari:
ok, I made changes to use the same map function for both `stickyHeaderIndices` and  `pagingEnabled`. I hope that I did not mess up the logic.

necolas:
Nice work! Thank you for doing this

PR #1213, opened 2018-12-22
Remove Array.join

sayrer:
I profiled the Script time in the benchmark by adding console.profile around the timings for Script time in the benchmark code. The only thing that leapt out was this call to Array.join in the resolve function. Since the code iterates over the array just before anyway, I got rid of it. Now I don't see this function in any profiles, and it tests faster here.

I won't make any claims on the performance, but I'll let you check, since I assume you have a more consistent test routine than I do.


sayrer:
Oh, you do have to live with a trailing comma on your cache keys...

necolas:
Thanks Rob! I do see a difference and every little bit helps

Before

image

After

image

PR #1226, opened 2019-01-11
Implement TextInput prop "onContentSizeChange"

### LINKED ISSUE #793: TextInput: onContentSizeChange
Looks like it's not implemented.

awinograd:
This PR adds support for `TextInput`'s `onContentSizeChange`.

A couple of questions / considerations:

1. ~I'm currently called onContentSizeChange when the ref callback fires. In practice, it seems to be nice to get information about the initial layout, but it feels a little weird to fire that on "mount" since it could be argued the initial mount isn't a size "change".~ https://github.com/necolas/react-native-web/pull/1226#issuecomment-453564745
1. enzyme doesn't support scrollHeight and scrollWidth (https://github.com/airbnb/enzyme/issues/1435#issuecomment-357130838) so it makes testing those values a little tricky. Please let me know if there's a better way to go about that. I also left some relevant comments in the corresponding test.

Fixes #793

### DIFF
--- a/packages/react-native-web/src/exports/TextInput/__tests__/index-test.js
+++ b/packages/react-native-web/src/exports/TextInput/__tests__/index-test.js
@@ -166,6 +166,32 @@ describe('components/TextInput', () => {
     expect(onChangeText).toBeCalledWith(newText);
   });
 
+  describe('prop "onContentSizeChange"', () => {
+    test('without "multiline"', () => {
+      const onContentSizeChange = jest.fn();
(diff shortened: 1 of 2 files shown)


necolas:
Thanks! What you'll want to do is use snack.expo.io to test the behaviour in React Native and make sure it matches up with your implementation

awinograd:
@necolas, I've confirmed that onContentSizeChange is called on mount.

https://snack.expo.io/@awinograd/textinput-oncontentsizechange

necolas:
Thanks!

PR #1229, opened 2019-01-15
Fix Image edge bug when using backgroundColor and borderRadius

### LINKED ISSUE #1228: Images with backgroundColor and borderRadius > 1 has a small bug at the edge
**The problem**

| Web | Mobile |
| --- | ------ |
| image | image |

Notice the white edge at the web version and not on the mobile.
It's a very subtle thing but we should aim for pixel perfect.

**How to reproduce**
Simplified test case: https://codesandbox.io/s/817nx1kzq8

```
import React from 'react'
import { Image, View } from 'react-native'

  return (
    <View
      style={{
        flex: 1,
        alignItems: 'center',
        justifyContent: 'center',
        backgroundColor: '#000000',
      }}
    >
      {/* with backgroundColor and borderRadius */}
      <Image
        source={{ uri: 'https://github.com/pedrottimark.png' }}
        style={{
          width: 200,
          height: 200,
          margin: 20,
          borderRadius: 100,
          backgroundColor: '#FFFFFF',
        }}
      />

      {/* without backgroundColor */}
      <Image
        source={{ uri: 'https://github.com/pedrottimark.png' }}
        style={{ width: 200, height: 200, margin: 20, borderRadius: 100 }}
      />
    </View>
  )
}
```

**Environment**

* React Native for Web (version): 0.9.9
* React (version): 16.7.0-alpha.2
* Browser: Chrome 71

**Additional context**

Based on your implementation, you created another div inside my `<Image />` and added `background-image` there.

One possible solution is to move both `background-color` and `border-radius` to the same div that has `background-image`.

brunolemos:
Fix #1228

### DIFF
--- a/packages/react-native-web/src/exports/Image/index.js
+++ b/packages/react-native-web/src/exports/Image/index.js
@@ -222,6 +222,8 @@ class Image extends Component<*, State> {
     const backgroundImage = displayImageUri ? `url("${displayImageUri}")` : null;
     const flatStyle = { ...StyleSheet.flatten(this.props.style) };
     const finalResizeMode = resizeMode || flatStyle.resizeMode || ImageResizeMode.cover;
+    const backgroundColor = flatStyle.backgroundColor;
+    const borderRadius = flatStyle.borderRadius;
 
     // CSS filters
     const filters = [];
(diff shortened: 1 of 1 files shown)


brunolemos:
I just noticed it fixes the issue on Chrome but has no effect on Safari and Firefox. Maybe you can come up with something that works on all of them, but until them I'll keep using this solution on my project.

brunolemos:
Tried some alternatives, moving the `background-image` and `background-size` from the inner div to the main div fixes it in all browsers. But I don't know if this would break something. cc @necolas

PR #1231, opened 2019-01-18
feat(ProgressBar): add animating property

jaulz:
Adds `animating` property similar to the one here: https://facebook.github.io/react-native/docs/progressbarandroid#animating


necolas:
Thanks. The UI for `ProgressBarAndroid` is largely equivalent to what is already available with `ActivityIndicator`. The `ProgressBar` in React Native for Web is closer to `ProgressViewIOS` in the UI it produces, but since it aims to be a cross-platform replacement for progress bar UIs, it cherry-picks some props props from `ProgressBarAndroid`. Hopefully we'll deprecate the platform-specific Progress components in React Native and add `ProgressBar` some day.

PR #1233, opened 2019-01-23
ViewPropTypes: Fix accessibilityStates propType to expect an array

ariabuckles:
Hi! Thank you for all your work on react-native-web, it has been fantastic.

I've been trying 0.10.0-alpha.2 to be able to try using #1113 's accessibility options,
and ran into a proptype warning when I pass `['selected']` to `accessibilityStates`, and
looking at the source, it looks like changing accessibilityStates's proptype
to arrayOf would match its flow type and react-native's documentation.

I hope it's all right to make a pull request against 0.10.0-wip, but if not feel
free to let me know if there's a better way I could help with this.

Currently, ViewPropTypes.accessibilityStates specifies `oneOf` several
string options, but requires an array of such strings. This commit changes
View's accessibilityStates proptype to accept an array of those options, using PropTypes.arrayOf(Proptypes.oneOf())

Test plan:

Ran `yarn compile` and took the resulting
dist/cjs/exports/View/ViewPropTypes.js file and used it in an internal
project, fixing the proptype warning `accessibilityStates={['selected']}`
was previously reporting.

Best wishes,
Aria


necolas:
Thanks for spotting this. I've already rebased the wip branch to remove the ongoing changes to style sheet, so I editing the original accessibility states commit to apply this fix. I published `alpha.3` with this fix, and with the internal changes to styles reverted.

ariabuckles:
Awesome; thank you!

Issue #1011, opened 2018-06-28
Text: onLongPress support


chrisdrackett:
**The problem**

When I add a `onLongPress` event to a text node I can't get the event to fire on a long press

**How to reproduce**

Simplified test case: https://codesandbox.io/s/6w34l63jw

Steps to reproduce:
1. Long press on text

**Expected behavior**

console will long "long press!"

**Environment (include versions). Did this work in previous versions?**

* React Native for Web (version): 0.7.3
* React (version): 16.4.0
* Browser: Safari / Chrome

necolas:
Yes, that's mentioned in the compatibility table https://github.com/necolas/react-native-web#compatibility-with-react-native

chrisdrackett:
Oops! Sorry about that!

necolas:
Decided to close this as "wontfix" because the primitive Text component should be simple and fast, which I don't think is the case when it include built-in touch handling, and it could cause regressions for apps like Twitter Lite https://github.com/necolas/react-native-web/pull/1272#issuecomment-539217703.

PR #1235, opened 2019-01-26
Text: onLongPress feature

DaKaZ:
In an effort to close issue #1011 I implemented an onLongPress handler for text elements

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "private": true,
-  "version": "0.9.12",
+  "version": "0.9.13",
   "name": "react-native-web-monorepo",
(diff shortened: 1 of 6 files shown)


necolas:
Thanks but the implementation should draw from React Native and integrate with the event system ([see here](https://github.com/facebook/react-native/blob/master/Libraries/Text/Text.js)), so this PR is not a suitable implementation

DaKaZ:
Thanks for the quick feedback. I’m sorry I don’t follow your direction though. Could you provide me a little more direction on what you’re looking for and I’ll do my best to update the PR

PR #1241, opened 2019-02-05
Re-implement onWheel to fix inverted scrolling

danalloway:
Beginning with this issue #995 and then this pull request #1137 with it's review, this is my attempt at a fix for inverted scrolling direction.

The original pull request #1137 did a few things differently and did not work properly on Firefox.

This PR:
- Addresses an issue with Firefox using `deltaMode` of `1` which gives you the number of lines, not pixels, in the `deltaY` value of the `onWheel` event. This results in the inverted scroll only scrolling a few pixels at a time.
- Uses the `onWheel` prop that is passed to the underlying ScrollView to listen to the wheel event instead of attaching an additional one. This eliminates the need to attach and detach listeners.
- Dispatches the `onWheel` event to any child lists as well as forwards the `onWheel` prop on just like how it does in `_onScroll` (do we want to do this, is this right?)

How I'm figuring out the number of pixels in a line is interesting, it only applies to events that have their `deltaMode` set to 1 though. I question if we even need this? is just setting it do the default enough for a proper scroll experience?

### DIFF
--- a/packages/react-native-web/src/vendor/react-native/VirtualizedList/index.js
+++ b/packages/react-native-web/src/vendor/react-native/VirtualizedList/index.js
@@ -1185,10 +1185,10 @@ class VirtualizedList extends React.PureComponent<Props, State> {
 
     // prepare direction and delta based on scroll orientation
     if (this.props.horizontal) {
-      delta = -e.wheelDeltaX || -e.deltaX
+      delta = -e.deltaX || e.wheelDeltaX
       direction = 'scrollLeft'
     } else {
-      delta = -e.wheelDeltaY || -e.deltaY
+      delta = -e.deltaY || e.wheelDeltaY
       direction = 'scrollTop'
 
       // if deltaMode is 1 (Firefox) then the deltaY is reported in lines, not pixels
@@ -1201,7 +1201,7 @@ class VirtualizedList extends React.PureComponent<Props, State> {
           lineHeight = styles.getPropertyValue('line-height')
         }
 
-        delta *= parseFloat(lineHeight)
+        delta *= (parseFloat(lineHeight) * 2)
       }
     }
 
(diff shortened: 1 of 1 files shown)


danalloway:
testing this a bit more and it appears to have issues on Windows and Android, will update if I find a solution

danalloway:
there are some cases where the corrected inverted scrolling jumps backwards as you scroll (can easily reproduce in the Brave browser)
wrapping the `scrollTop` assignment in a `setTimeout(() => {}, 0)` resolves the issue
can anyone shed some light on why that would happen?

packages/react-native-web/src/vendor/react-native/VirtualizedList/index.js
```diff
@@ -1248,6 +1278,24 @@ class VirtualizedList extends React.PureComponent<Props, State> {
...
+      this.props.onWheel(e);
+    }
+
+    if (this.props.inverted) {
+      const scrollNode = this.getScrollableNode()
+      const { delta, direction } = this._selectWheelDelta(e, scrollNode)
+
+      scrollNode[direction] += delta
+
+      e.preventDefault();
```
danalloway: do we need this `preventDefault` here?

danalloway: ok, it looks like we do, as it breaks the scrolling in Chrome without it

danalloway:
ok, everything works properly on Windows across Brave, Firefox, Edge and Chrome now.
just need to confirm that scrolling in Chrome on Android works properly

danalloway:
broken again on latest version of chrome
fixing the scrolling on the `scaleY -1` hack with another hack just doesn't smell right

going to close this PR for now, this isn't the way to solve it

necolas:
Thanks for looking into this. I'll give this some thought as it might be something we can solve at a lower level in the future

danalloway:
when that time comes let me know how I can help, even if it's just testing or vetting an idea

annie-elequin:
Hey y'all, attempting to build a chat web app and running into this same problem with the inverted FlatList. is there a recommended way to workaround the scrollwheel for now, or was it not really solved?

Issue #1172, opened 2018-11-12
React Native 0.60 tasks


necolas:
Tasks related to changes in React Native 0.60.

- [x] Migrate docs and RNTester examples to storybook
- [x] Update project dependencies
- [ ] ~Add touchable support to `Text` (#1011) (@DaKaZ)~
- [x] Add flow types for `ScrollView`
- [x] Add flow types for `View`
- [x] Add flow types for `Text`
- [x] Add flow types for styles
- [x] Update `Touchable` snapshot (@necolas)
- [x] Update `TouchableHighlight` snapshot (@necolas)
- [x] Update `TouchableOpacity` snapshot (@necolas)
- [x] Update `TouchableWithoutFeedback` snapshot (#1105) (@necolas)
- [x] Add `forwardedRef` prop to `Text`
- [x] Add `forwardedRef` prop to `View`
- [x] Add `accessibilityState`
- [x] Add `accessibilityRelationship`
- [x] Remove `accessibilityStates`
- [x] Remove `accessibilityTraits`
- [x] Remove `accessibilityComponentType`
- [x] Replace `context.isInAParentText` w/ `React.createContext` (@awinograd)
- [x] Update `ListView` snapshot (Requires types for ScrollView) (@necolas)
- [x] Update `PanResponder` snapshot
- [x] Update `SwipeableListView` snapshot
- [x] Add deprecation warnings for `className` (#1146) (@necolas)
- [x] Merge StyleSheet default layout optimisations (#1136, #1165) (@necolas)
- [x] Upgrade Flow (https://github.com/necolas/react-native-web/issues/1171) (@necolas)
- [x] Upgrade Babel (https://github.com/necolas/react-native-web/issues/1170) (@necolas)
- [x] Update `Animated` snapshot (@necolas)
- [x] Update `Batchinator` snapshot (@necolas)
- [x] Update `FillRateHelper` snapshot
- [x] Update `FlatList` snapshot (@necolas)
- [x] Update `NativeEventEmitter` snapshot
- [x] Update `SectionList` snapshot (@necolas)
- [x] Update `SwipeableFlatList` snapshot (@necolas)
- [x] Update `ViewabilityHelper` snapshot
- [x] Update `VirtualizeUtils` snapshot
- [x] Update `VirtualizedList` snapshot (@necolas)
- [x] Update `VirtualizedSectionList` snapshot (@necolas)
- [x] Change accessibility props (#1112) (@necolas)

Afterwards:

[...]

Issue #1171, opened 2018-11-12
Upgrade Flow


necolas:
Looking for help to upgrade Flow to the latest version.

We can `@noflow` most of the files and then migrate modules individually.

Also looking to start using Flow `strict-local`. You can look at the React Native codebase for examples of how to type components

necolas:
Thanks! Let's use the same version of flow as found in RN 0.57 release. Prioritise the exported modules. I'm not sure about prop types; they're being removed from RN but a lot of web people won't have flow setup

necolas:
This is done in branch `react-native-0.57`

PR #1248, opened 2019-02-20
Replace contextType with createContext for isInAParentText

awinograd:
This is a first pass at the createContext task in https://github.com/necolas/react-native-web/issues/1172

A few questions before I continue:

1. ~Do you like the context being created in `Text` and then being imported in `View`/`Image` or should it be in it's own export/module~
1. Do you like how render/renderInner are split, or did you envision a different way of adding the consumers to the tree. One consequence of this current approach is that now `shallow()` renders in the test suite fail because they aren't rendering past the `IsInAParentTextConsumer` from what I can tell.
1. ~I'm having an issue with flow that I'm not really sure how to solve. Looks like it doesn't think `createContext` exists even though the code runs fine. Sorry, I use Typescript in my project so don't have much troubleshooting skill with flow~

```
13: import { Component, createContext } from 'react';
                         ^^^^^^^^^^^^^ Named import from module `react`. This module has no named export called `createContext`.
```

### DIFF
--- a/.flowconfig
+++ b/.flowconfig
@@ -1,5 +1,5 @@
 [version]
-^0.63.0
+^0.78.0
 
 [ignore]
 <PROJECT_ROOT>/.*/__tests__/.*
--- a/package.json
+++ b/package.json
@@ -46,7 +46,7 @@
     "eslint-config-prettier": "^2.9.0",
     "eslint-plugin-promise": "^3.7.0",
     "eslint-plugin-react": "^7.7.0",
-    "flow-bin": "^0.63.1",
+    "flow-bin": "^0.78.0",
     "glob": "^7.1.2",
     "husky": "^0.14.3",
     "jest": "^22.4.3",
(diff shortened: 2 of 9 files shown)


SimenB:
For 3. you need to upgrade flow (or ignore the error?) - react types are built-in to flow: https://github.com/necolas/react-native-web/blob/8ce5750d3437d1caa3d74a74a3625c70c54b66ab/package.json#L49
Latest is 0.93. I'm not sure which version came with `createContext`

necolas:
Please take a look at how React Native implements the text context

awinograd:
@necolas, I've updated to follow the implementation of react-native.

I've also just bumped flow as you requested in https://github.com/necolas/react-native-web/issues/1171#issuecomment-449069940 since I needed it to properly add `createContext`. There were 2 introduced flow errors I've ignored for now. 

One thing I'm still not sure about is how to handle the testing errors described above

PR #1251, opened 2019-02-20
[DRAFT] Remove `react-art`

EvanBacon:
React ART is a massive library that has mostly been replaced by `react-native-svg`. Including it by default makes the lighthouse scores suffer by adding an additional ~110kb  to the bundle (~100ms+ FCP).

Currently there is no clear way use the library without react-art. The docs state that you can install it seperately but build errors are thrown when the library isn't present.


necolas:
If you use the babel plugin or tree shaking (ES modules) it won't be included. This is what Twitter does. You might be using the common js modules

necolas:
Removing ART from React Native has been requested here: https://github.com/facebook/react-native/issues/23313. But rather than wait on React Native I'm going to remove it from this project in 0.11.

I think ART could be moved to community maintenance without waiting on RN changes by starting with a package `react-native-art` that exports:

```
// index.js
import { ART } from 'react-native';
export default ART;
```

```
// index.web.js
import ART from 'react-art';
export default ART;
```

PR #1252, opened 2019-02-21
toolchain fixes for contributing in a Windows environments

danalloway:
Fixes various package.json tasks that lacked Windows support.


necolas:
Thank you!

PR #1261, opened 2019-02-28
View: Pass through drag event handlers to underlying DOM element

sstur:
This ensures that we pass the [web-only] event handlers for drag and drop to the underlying DOM element for a `View`.

 - onDragStart
 - onDragEnter
 - onDragOver
 - onDragLeave
 - onDrop

I know this feature diverges from react-native, but it appears there is no other way to get drag & drop file uploads with RNW (besides imperatively grabbing a ref and doing some [hacky addEventListener stuff](https://user-images.githubusercontent.com/369384/53592744-c5e1b400-3b4b-11e9-9e2d-f9ee7ca1bbbf.png)).


packages/react-native-web/src/exports/View/ViewPropTypes.js
```diff
@@ -84,17 +89,19 @@ const ViewPropTypes = {
...
+      'selected',
+      /* web-only */
+      'busy',
+      'checked',
+      'expanded',
+      'grabbed',
+      'invalid',
+      'pressed'
+    ])
+  ),
```
sstur: Evidently, this formatting was automatically done by the pre-commit hook.

gut4:
@necolas can this be merged?

sstur:
@gut4, just an update on what other workarounds I've explored since I made this PR..

I think the easiest way might be to use [createElement from react-native-web](https://github.com/necolas/react-native-web/blob/master/docs/guides/advanced.md#use-with-existing-react-dom-components).

Something like this should work:
```js
import { createElement } from 'react-native-web'; // or from 'react-native'

function ThingWithDragHandlers(props) {
  return <div {...props} onDragStart={...} />
}

function Draggable(props) {
  return createElement(ThingWithDragHandlers, props);
}

export default Draggable; // This can be used anywhere a `<View />` can be used.
```

PR #1269, opened 2019-03-03
Update flow bin to 0.94

zoontek:
Flow team has made great improvements on the checking speed and introduced the usage of a LSP since the version 0.75, now by default.

This update is quite a huge bump, but the type fixes are minor.

### DIFF
--- a/packages/react-native-web/src/exports/StyleSheet/index.js
+++ b/packages/react-native-web/src/exports/StyleSheet/index.js
@@ -1,5 +1,10 @@
+// @flow
+
 import { canUseDOM } from 'fbjs/lib/ExecutionEnvironment';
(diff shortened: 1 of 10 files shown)


packages/react-native-web/src/exports/StyleSheet/index.js
```diff
@@ -1,6 +1,11 @@
+// @flow
+
 import { canUseDOM } from 'fbjs/lib/ExecutionEnvironment';
 import StyleSheet from './StyleSheet';
 
+type Atom = number | boolean | Object | Array<?Atom>;
+export type StyleObj = Atom;
```
necolas: This creates a cyclic dependency between various files in `StyleSheet`. The types should live in a separate file

PR #1270, opened 2019-03-03
Fix cyclic dependency with PickerItem

th317erd:
This is to fix a cyclic dependency issue with the PickerItemPropTypes


necolas:
Thanks!

PR #1272, opened 2019-03-04
Fix issue #1011 - adds touchable event handlers to text

DaKaZ:
This is my second attempt to close issue #1011.  This borrows much from the react-native Text component (https://github.com/facebook/react-native/blob/master/Libraries/Text/Text.js) but modifies the approach to handle web specific events like `mousedown` and `mouseup` and `mouseout`.


DaKaZ:
@necolas did you want me to update this PR... just let me know

necolas:
Thanks for this PR, it looks good. However, I've been thinking about this feature for a while, and I think I'm going to keep it excluded from web by design. Apps like Twitter are using `Text` everywhere and adding the build and runtime overhead of `Touchable` to `Text` is (on balance) going to make the most common use of `Text` become more expensive. I think RN is doing too much in `Text` (it didn't use to overload it with press functionality either) vs `View`, and RN should probably redesign its exports to include something like `TouchableText`.

PR #1275, opened 2019-03-07
Made NativeEventEmitter work as expected.

EvanBacon:
# Why

We want to be able to use the same JS API across iOS, Android, and Web. This will enable that platform functionality. 
With turbo modules, the event system will more than likely change and we should revisit this. Regardless it seems like if you can implement NativeEventEmitter, and emit events, then either something should happen or a warning should be thrown. Otherwise this creates a very hard to track bug in the expected behavior.

I also removed the iOS Platform specific feature as we don't use that in web.

### DIFF
--- a/packages/react-native-web/src/vendor/react-native/NativeEventEmitter/RCTDeviceEventEmitter.js
+++ b/packages/react-native-web/src/vendor/react-native/NativeEventEmitter/RCTDeviceEventEmitter.js
@@ -1,0 +1,84 @@
+/**
+ * Copyright (c) 2015-present, Facebook, Inc.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @format
+ * @flow
+ */
+
+'use strict';
+
+import EventEmitter from '../emitter/EventEmitter';
+import EventSubscriptionVendor from '../emitter/EventSubscriptionVendor';
+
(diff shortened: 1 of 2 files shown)


packages/react-native-web/src/vendor/react-native/NativeEventEmitter/RCTDeviceEventEmitter.js
```diff
@@ -0,0 +1,84 @@
...
+
+import type EmitterSubscription from '../emitter/EmitterSubscription';
+
+function checkNativeEventModule(eventType: ?string) {
+  if (eventType) {
+    if (eventType.lastIndexOf('statusBar', 0) === 0) {
+      throw new Error(
+        '`' +
+          eventType +
+          '` event should be registered via the StatusBarIOS module',
```
necolas: This check doesn't seem necessary for web

EvanBacon: The AppState one may be useful as the web module actually registers a listener.

EvanBacon:
I'm going to run this through the Expo test suite to see if I get the expected behavior.

necolas:
Sounds good, thanks!

EvanBacon:
Is `__DEV__ ` not available in RNWeb?

EvanBacon:
@necolas will this be able to make the cut for 0.11?

necolas:
Yep!

PR #1277, opened 2019-03-08
[change] Remove default ES Module export

### LINKED ISSUE #1258: Remove default import

AFAICT you cannot shake the default export, even if it is not included in your project. 
Repro: [Basic Treeshaking](https://github.com/EvanBacon/treeshaking-test/blob/master/src/art.js)
Just removing `react-art` saves 143kb (non-gzip).

Also noting that in an RNWeb project, omitting the library `react-art` does not build: [Issue](https://github.com/necolas/react-native-web/issues/903#issuecomment-467268612)

@ide has opened a PR to remove the default import from React Native proper https://github.com/facebook/metro/pull/362

We could then delete: https://github.com/necolas/react-native-web/blob/8ce5750d3437d1caa3d74a74a3625c70c54b66ab/packages/react-native-web/src/index.js#L218-L325

 * Use a plugin like [deep-scope-analysis](https://github.com/vincentdchan/webpack-deep-scope-analysis-plugin) which will remove the default if there are absolutely no module.exports/requires. Unfortunately this means that react-art would still get bundled (amongst others). 

**Additional context**

N/A

necolas:
Remove the default export that was provided for compatibility with legacy
imports of React Native CommonJS modules. These patterns is no longer
supported:

```
import ReactNative from 'react-native';
const ReactNative = require('react-native');
```

Fix #1258


joan-saum:
Hi @necolas 

I'm not sure to understand why the default has been removed.
In my project, some dependencies try to access to the default export of react-native. It only works if i add the export default you removed here

See my issue for more information : https://github.com/necolas/react-native-web/issues/1366

Is there any other way to make it work without change the export ?

paularmstrong:
@joan-saum Please see the related issue: #1258

PR #1290, opened 2019-03-14
[jest-preset] export module to resolve in multi-project jest setups

paularmstrong:
## Problem

In a multi-project jest setup, using the `react-native-web` jest preset option will not work correctly, unless your sub-project sets its root to the repo root instead of the project root.

For example, given the following structure:

```
repo/
├── node_modules/
├── packages/
│   ├── with-react-native-web/
│   │   └── jest.config.js
│   └── other-package
│       └── jest.config.js
└── jest.config.js
```

In `packages/with-react-native-web/jest.config.js`, we add:

```
preset: 'react-native-web',
rootDir: './'
```

Since Yarn will install `react-native-web` at the root `node_modules` folder, the `<rootDir>` in the `jest-preset` will not resolve correctly:

```
$ jest
● Validation Error:

  Module <rootDir>/node_modules/react-native-web/jest/setup.js in the setupFiles option was not found.
         <rootDir> is: /Development/my-app/packages/with-react-native-web

  Configuration Documentation:
  https://jestjs.io/docs/configuration.html
```

## Solution

Convert to a JS module and use `path` and `__dirname` to resolve relative to the config file.

### DIFF
--- a/packages/react-native-web/jest-preset.js
+++ b/packages/react-native-web/jest-preset.js
@@ -1,0 +1,9 @@
+const path = require('path');
+
+module.exports = {
+  moduleNameMapper: {
+    '^react-native$': path.join(__dirname, 'dist', 'cjs')
+  },
+  setupFiles: [path.join(__dirname, 'jest', 'setup.js')],
+  testEnvironment: require.resolve('jest-environment-jsdom')
+};
--- a/packages/react-native-web/jest-preset.json
+++ b/packages/react-native-web/jest-preset.json
@@ -1,9 +1,0 @@
-{
-  "moduleNameMapper": {
-    "^react-native$": "<rootDir>/node_modules/react-native-web/dist/cjs"
-  },
-  "setupFiles": [
-    "<rootDir>/node_modules/react-native-web/jest/setup.js"
-  ],
-  "testEnvironment": "jsdom"
-}
--- a/packages/react-native-web/package.json
+++ b/packages/react-native-web/package.json
@@ -11,7 +11,7 @@
   "files": [
     "dist",
     "jest",
-    "jest-preset.json",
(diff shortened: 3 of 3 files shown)


SimenB:
I'd recommend to do `require.resolve` where possible (including on `jest-environment-jsdom`) instead of constructing paths using `path`. Both for early errors, but also for e.g. yarn PnP to work properly

paularmstrong:
@SimenB in what scenarios would the proposed solution not work? The file references are relative to the given file already. Doing a resolve seems like an extra step of indirection for the same result, so I'm not sure I understand the benefit that you're getting at.

SimenB:
Oh, I missed that the referenced files were in the same module - I misread the React one to reach into React's dist directory not RNW's. No need to change then!

Could still consider to change the `testEnvironment`, but `jest-config` has dependencies on both `jest-environment-jsdom` and `jest-environment-node` precisely to avoid issues with hoisting

---

Thanks for doing this btw - I made the same change to RN itself in https://github.com/facebook/react-native/pull/22972 and I'm hopeful Expo will do the same whenever they bump to a newer version of RN. Using RN in monorepos is pretty painful when modules hardcode assumptions about paths in `node_modules` and changes like this helps decouple them 🙂

paularmstrong:
> Oh, I missed that the referenced files were in the same module

Phew! I thought maybe there was something I'd completely glossed over as well 😄 

> Could still consider to change the testEnvironment

To `jsdom`?

SimenB:
To `require.resolve('jest-environment-jsdom')`. Again, not really needed as it's bundled with jest, though

paularmstrong:
@SimenB Makes sense! Just updated.

necolas:
Thank you both

PR #1302, opened 2019-03-21
[fix] CSS insertion in Edge browser

### LINKED ISSUE #1300: StyleSheet: Edge browser bug causes broken styles
**The problem**

Edge browser throws `DOMException: HierarchyRequestError` while inserting CSS rules into CSS Media Queries.

**How to reproduce**

Simplified test case: https://gist.github.com/necolas/aa0c37846ad6bd3b05b727b959e82674

Steps to reproduce:
1. Open HTML page in Edge
2. Observed: Screen is orange in color
3. Expected: Screen is violet in color

**Expected behavior**

Rendering in Edge should match other browsers.

* React Native for Web (version): 0.11
* React (version): N/A
* Browser: Edge

**Additional context**

React Native for Web 0.11 introduced ordered CSS rules, which is implemented using Media Query blocks. This appeared viable in an initial proof-of-concept, including in Edge, because the Edge bug doesn't seem to manifest until a few insertions have taken place.

To avoid the Edge bug the ordering will have to rely on a different mechanism.

necolas:
Edge browser throws `HierarchyRequestError` while inserting CSS rules into CSS
Media Queries. Therefore, a different mechanism is required to control CSS
order. This patch tracks the starting index of each group of CSS rules in the
DOM style sheet.

Fix #1300

### DIFF
--- a/packages/react-native-web/src/exports/StyleSheet/createOrderedCSSStyleSheet.js
+++ b/packages/react-native-web/src/exports/StyleSheet/createOrderedCSSStyleSheet.js
@@ -7,7 +7,7 @@
  * @flow strict-local
  */
 
-type Groups = { [key: number]: Array<string> };
+type Groups = { [key: number]: { start: ?number, rules: Array<string> } };
 type Selectors = { [key: string]: boolean };
 
 const slice = Array.prototype.slice;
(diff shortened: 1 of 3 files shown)


packages/react-native-web/src/exports/StyleSheet/createOrderedCSSStyleSheet.js
```diff
@@ -147,7 +161,12 @@ function insertRuleAt(root, cssText: string, position: number) {
   try {
     // $FlowFixMe: Flow is missing CSSOM types needed to type 'root'.
     root.insertRule(cssText, position);
+
+    return true;
   } catch (e) {
+    console.log(e);
+    console.log(cssText, position);
```
necolas: Remove logs

PR #1305, opened 2019-03-26
Added `onScroll(Begin|End)Drag` Events

EvanBacon:
* Added [`onScrollBeginDrag`](https://facebook.github.io/react-native/docs/scrollview#onscrollbegindrag)
* Added [`onScrollEndDrag`](https://facebook.github.io/react-native/docs/scrollview#onscrollenddrag)

### DIFF
--- a/packages/react-native-web/src/exports/ScrollView/ScrollViewBase.js
+++ b/packages/react-native-web/src/exports/ScrollView/ScrollViewBase.js
@@ -170,6 +170,10 @@ export default class ScrollViewBase extends Component<*> {
   _handleScrollStart(e: Object) {
     this._state.isScrolling = true;
     this._state.scrollLastTick = Date.now();
+    const { onScrollBeginDrag } = this.props;
+    if (onScrollBeginDrag) {
+      onScrollBeginDrag(normalizeScrollEvent(e));
+    }
   }
 
   _handleScrollTick(e: Object) {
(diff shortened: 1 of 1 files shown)


EvanBacon:
Upon further review it seems that perhaps the `onScrollEndDrag` is implemented closer to `onMomentumScrollEnd` since a debounce is involved. 
I can't find a very consistent way to observe if `Momentum` is enabled. Perhaps we should add that before implementing `onMomentumScrollBegin`, `onMomentumScrollEnd`.

EvanBacon:
hmm #1091 🙁 perhaps that PR should be fixed instead. I need this and a few other features to get added to `ScrollView` to fix a couple of complex interactions.

necolas:
Yeah this doesn't look quite right. I think these scroll events are going to be more complicated to implement as they will need to be aware of whether a touch is active in order to determine whether scrolling is momentum-based or not.

EvanBacon:
Perhaps we could create new props like `onScrollBegin`, `onScrollEnd`. It would be really helpful to access these values. The alternative of using `onScroll` with an animated value is a huge bottle neck on performance. Of course I would much rather have this done correctly, but the particular invocation this is currently surfacing is pretty valuable.

ex: [onScroll](https://github.com/expo/expo/blob/55c5b930b4a2730bc26b0a04dc28df5a983aabf9/apps/native-component-list/instagram/components/ViewPager.js#L53) & [comparing every frame](https://github.com/expo/expo/blob/55c5b930b4a2730bc26b0a04dc28df5a983aabf9/apps/native-component-list/instagram/components/Slider.js#L75-L82)

evie9z:
Hi, I'm working on a project that really needs this feature. May I ask any updates on this PR yet?

shljessie:
How is this going sir? I need this feature desperately!
@EvanBacon @necolas

KingAmo:
any update on this ? i just want do something when scroll begin

lucaashrq:
Any update?

EvanBacon:
I don't have time to investigate this further, please see @necolas's comment for reasons this cannot be merged in its current state.

PR #1307, opened 2019-03-26
[ScrollView] Added `showsHorizontalScrollIndicator` & `showsVerticalScrollIndicator`

EvanBacon:
Added pseudo support for `showsHorizontalScrollIndicator` and `showsVerticalScrollIndicator` by creating the style property `scrollIndicator` which disables the scroll bars.

Unfortunately this means if either `showsHorizontalScrollIndicator` or `showsVerticalScrollIndicator` are false, then both are hidden.

### DIFF
--- a/packages/react-native-web/src/exports/ScrollView/ScrollViewBase.js
+++ b/packages/react-native-web/src/exports/ScrollView/ScrollViewBase.js
@@ -132,11 +132,11 @@ export default class ScrollViewBase extends Component<*> {
         onTouchMove={this._createPreventableScrollHandler(this.props.onTouchMove)}
         onWheel={this._createPreventableScrollHandler(this.props.onWheel)}
         ref={this._setViewRef}
-        style={StyleSheet.flatten([
+        style={[
           style,
           !scrollEnabled && styles.scrollDisabled,
           !showScrollIndicator && style.disableScrollIndicator
-        ])}
+        ]}
       />
     );
   }
--- a/packages/react-native-web/src/exports/ScrollView/ScrollViewBase.js
+++ b/packages/react-native-web/src/exports/ScrollView/ScrollViewBase.js
@@ -68,7 +68,9 @@ export default class ScrollViewBase extends Component<*> {
 
   static defaultProps = {
     scrollEnabled: true,
-    scrollEventThrottle: 0
+    scrollEventThrottle: 0,
+    showsHorizontalScrollIndicator: true,
+    showsVerticalScrollIndicator: true
   };
 
   _debouncedOnScrollEnd = debounce(this._handleScrollEnd, 100);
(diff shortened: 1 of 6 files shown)


packages/react-native-web/src/exports/ScrollView/ScrollViewBase.js
```diff
@@ -131,7 +134,8 @@ export default class ScrollViewBase extends Component<*> {
         ref={this._setViewRef}
         style={StyleSheet.compose(
           style,
-          !scrollEnabled && styles.scrollDisabled
+          !scrollEnabled && styles.scrollDisabled,
+          !showScrollIndicator && style.disableScrollIndicator
```
necolas: `StyleSheet.compose` only accepts 2 arguments.

packages/react-native-web/src/exports/StyleSheet/compile.js
```diff
@@ -144,6 +144,16 @@ function createAtomicRules(identifier: string, property, value): Rules {
       break;
     }
 
+    case 'scrollIndicator': {
+      if (value === 'none') {
+        rules.push(
+          `${selector}::-webkit-scrollbar{width: 0 !important}`,
```
necolas: Why is this rule `width:0` and using `!important`? Would `display:none` work instead?

packages/react-native-web/src/exports/View/ViewStylePropTypes.js
```diff
@@ -55,6 +55,7 @@ const ViewStylePropTypes = {
   overscrollBehavior: overscrollBehaviorType,
   overscrollBehaviorX: overscrollBehaviorType,
   overscrollBehaviorY: overscrollBehaviorType,
+  scrollIndicator: string,
```
necolas: Since this is an invented property we should have a stricter type

packages/react-native-web/src/exports/ScrollView/ScrollViewBase.js
```diff
@@ -122,17 +124,19 @@ export default class ScrollViewBase extends Component<*> {
...
+    const showScrollIndicator = showsHorizontalScrollIndicator && showsVerticalScrollIndicator;
     return (
       <View
         {...other}
         onScroll={this._handleScroll}
         onTouchMove={this._createPreventableScrollHandler(this.props.onTouchMove)}
         onWheel={this._createPreventableScrollHandler(this.props.onWheel)}
         ref={this._setViewRef}
-        style={StyleSheet.compose(
+        style={StyleSheet.flatten([
```
necolas: `flatten` is a deopt.

```suggestion
style={[
  style,
  !scrollEnabled && styles.scrollDisabled,
  !showScrollIndicator && style.disableScrollIndicator
]}
```

EvanBacon:
@necolas is there something I can do to get this merged <33

PR #1316, opened 2019-03-31
Add mui styles

oliviertassinari:
I was curious to see how it would perform.

Capture d’écran 2019-03-31 à 15 30 43


necolas:
OK. Not going to be significantly different with the same underlying engine. Maybe it was useful for you to check you're happy with how your lib performs, but the benchmarks in this repo aren't meant to be a place where I collect and profile every styling library.

oliviertassinari:
Most of the value I got was from running the suite. It's historized here, even if closed. It's good enough for me :). Thanks.

Issue #1327, opened 2019-04-14
No `borderStyle` set to `<Text />` component


yairopro:
**The problem**
`<Text />` have no `borderStyle` set by default.

**How to reproduce**
Simplified test case: https://codesandbox.io/s/ry0ok78owp

Steps to reproduce:
1. Add a border to a text
2. No border will appear unless you clearly set the `borderStyle`.

**Expected behavior**
A solid border should appear like in react-native: https://snack.expo.io/By-F-zbc4 

**Environment (include versions). Did this work in previous versions?**

* React Native for Web (version): 0.11.2
* React (version): 16.8.5
* Browser: Chrome 73

necolas:
Fixed in master

PR #1335, opened 2019-05-01
Fix border style for Text component

ashbhir:
Fixes Issue #1327 
Currently when transpiled for web, Text Component is not rendering with any border when only `borderWidth` is given (renders correctly in Native). The simple fix is to set `border` property to `0 solid black` as default for `Text` component similar to what has been done for most of the other components like `View`, `TextInput` etc.


ashbhir:
@necolas

PR #1344, opened 2019-05-11
Update ImageURICache when prefetching; Implement Image.queryCache

comp615:
Image.queryCache is a React Native method that allows the user to see
if a given uri is in the cache. It specifies three return options:
disk, memory or both. Choosing both seemed most appropriate since
we don't really know and can't confirm.

The way Image is implemented, if RNW thinks the image might already
be loaded, it displays it immediately. Otherwise there can be a flash
of a frame. In some scenarios, if the user chooses to preload and then
make an Image element, it would still flash. By adding it to the cache,
we can prevent that.

Finally, there was some test pollution from the image uri cache
in the tests, so I cleared that out.

### DIFF
--- a/packages/react-native-web/src/exports/Image/__tests__/index-test.js
+++ b/packages/react-native-web/src/exports/Image/__tests__/index-test.js
@@ -14,6 +14,7 @@ const findImageSurfaceStyle = wrapper => StyleSheet.flatten(wrapper.childAt(0).p
 
 describe('components/Image', () => {
   beforeEach(() => {
+    ImageUriCache._entries = {};
     window.Image = jest.fn(() => ({}));
   });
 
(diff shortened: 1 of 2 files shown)


necolas:
Related question. Where are you performing the `HEAD` request to get data about the image's size (bytes) during data saver mode? Would it be helpful if you pass a `source` to `prefetch` and could do something like this `Image.prefetch({ uri, method: 'HEAD' }).then(data => {})`

comp615:
It's part of the image cache where we markAsViewed once data saver has been accepted. That seems to suggest it could align well as part of the ImageCache in RNW. The hitch is that it looks like DMs that are protected send the information on a custom header instead of content-length.

More generally, I think API-wise I'd expect a different method since calling it with head does not actually prefetch the image, so it wouldn't have the same behavior. In addition, there's methods for getSize and getSizeWithHeaders...so getFileSize might be consistent? Then the data it returns could be the size, or could be the headers and status code, or whatever else makes sense.

PR #1345, opened 2019-05-11
Image - Don't swap from defaultSource to source until fully loaded

comp615:
Image provides a mechanism (defaultSource) for showing a loading or
fallback image. Currently, as soon as the source starts loading,
which is immediately after the first render, it's swapped in. The
intention is to support progressive JPEGs.

However, in cases where the user has explicitly given a defaultSource,
I'd argue that should be a better experience than watching the PJPEG load.
In conjunction with my other PR, a dev can now easily manage multiple versions
of an image via RNWs cache. Imagine on Twitter that you view a low-res image
in the timeline. Clicking to the gallery, we can read RNW's cache and provide
the highest variant found in it as the defaultSource. For users offline, or on
slow connections, the image will immediate show up and stay until we are ready
to swap in the full image. It's a roll your own PJPEG.

In addition, Enzyme now runs lifecycle methods on shallow, so I cleaned up
a few other nearby tests.

### DIFF
--- a/packages/react-native-web/src/exports/Image/__tests__/index-test.js
+++ b/packages/react-native-web/src/exports/Image/__tests__/index-test.js
@@ -21,7 +21,7 @@ describe('components/Image', () => {
     window.Image = originalImage;
   });
 
-  test.only('prop "accessibilityLabel"', () => {
+  test('prop "accessibilityLabel"', () => {
     const defaultSource = { uri: 'https://google.com/favicon.ico' };
     const component = shallow(
       <Image accessibilityLabel="accessibilityLabel" defaultSource={defaultSource} />
--- a/packages/react-native-web/src/exports/Image/__tests__/index-test.js
+++ b/packages/react-native-web/src/exports/Image/__tests__/index-test.js
@@ -21,7 +21,7 @@ describe('components/Image', () => {
     window.Image = originalImage;
   });
 
-  test('prop "accessibilityLabel"', () => {
+  test.only('prop "accessibilityLabel"', () => {
     const defaultSource = { uri: 'https://google.com/favicon.ico' };
     const component = shallow(
       <Image accessibilityLabel="accessibilityLabel" defaultSource={defaultSource} />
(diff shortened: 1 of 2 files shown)


packages/react-native-web/src/exports/Image/index.js
```diff
@@ -166,7 +166,9 @@ class Image extends Component<*, State> {
       ImageUriCache.remove(prevUri);
       const isPreviouslyLoaded = ImageUriCache.has(uri);
       isPreviouslyLoaded && ImageUriCache.add(uri);
-      this._updateImageState(getImageState(uri, isPreviouslyLoaded));
+      this._updateImageState(getImageState(uri, isPreviouslyLoaded), !!this.props.defaultSource);
```
necolas: Please can we avoid coercion. And since this check is used 3 times we can lift up to a var: `const hasDefaultSource = this.props.defaultSource != null`

packages/react-native-web/src/exports/Image/__tests__/index-test.js
```diff
@@ -21,7 +21,7 @@ describe('components/Image', () => {
     window.Image = originalImage;
   });
 
-  test('prop "accessibilityLabel"', () => {
+  test.only('prop "accessibilityLabel"', () => {
```
necolas: `test.only`

necolas:
With regards to the example use case for Twitter, React Native's `Image` also supports multiple `source` objects which makes some of this more convenient and flexible to work with.

> [The `source`] prop can also contain several remote URLs, specified together with their `width` and `height` and potentially with `scale`/other URI arguments. The native side will then choose the best uri to display based on the measured size of the image container. A cache property can be added to control how networked request interacts with the local cache.

There's probably more we can do to make `Image` more sophisticated, especially once the loader is replaced with an `xhr` implementation that can report loading progress, include HTTP headers in the request, and abort requests. All the stuff related `Image` is also pretty old code that could be simpler if rewritten.

Issue #1219, opened 2019-01-03
Touchable issues


necolas:
**The problem**

The overall issue appears to be that a) events don't always occurs in the same order, which means the responder is not transferred correctly, b) mouse move events during a touch can cause the responder to be terminated, and c) scroll events are not always fired in situations where they are expected in React Native.

There are problems with the responder event system setup that are covered by these issues:

## #693 

https://glitch.com/edit/#!/statuesque-bedbug

> locationX is relative to the element on which the native event occurred but not the element which the responder event is attached.

## #731

https://glitch.com/edit/#!/cultured-bracket

> In React Native, ScrollView will trigger onScroll events if you're dragging down when at the top of the scroll view - this allows the responder system to transfer the responder grant from the touchable back to the scroll view. But the web stops dispatching scroll events when scrollTop === 0, so the responder system leaves the touchable as the responder. The reported issue goes away when ScrollView dispatches a DOM scroll event after it receives a touchmove that occurs when scrollTop === 0.

## #829 

> There are 2 significant issues with using `Touchable*` components within an app that relies on the `document.body` (rather than `ScrollView`) being the root scroller. I believe the root cause of both issues is that the body is not part of the responder system, which `Touchable` and `ScrollView` rely on to coordinate transfer of the responder grant.

## #1164

https://codesandbox.io/s/6njwzqojnr

> TouchableOpacity will trigger another component at the same location when changing views (if second component is an "<input>" or "<TextInput>")

## #1124 

> Across multiple browsers, if you hold the mouse very still onPress triggers, but if the mouse is in motion onPress does not work.

**Context**

Part of the solution may related to producing the correct events (as can be inferred from React Native). This might involve changes to the event filtering logic in the REP, and the scroll event logic in ScrollView.

* [Responder Event Plugin](https://github.com/necolas/react-native-web/blob/0.9.13/packages/react-native-web/src/modules/ResponderEventPlugin/index.js)
* [ScrollResponder](https://github.com/necolas/react-native-web/blob/0.9.13/packages/react-native-web/src/modules/ScrollResponder/index.js)
* [ScrollView](https://github.com/necolas/react-native-web/blob/0.9.13/packages/react-native-web/src/exports/ScrollView/index.js)

**Related incomplete patches**

Attempts to patch these problems.

* https://github.com/necolas/react-native-web/pull/997/files
* https://github.com/rxrossi/react-native-web/commit/1e3eb7fdf1ccb9e29abda83ae8482968e50d49b2

[...]

PR #1348, opened 2019-05-16
Preserve onClick/keyUp/keyDown event listener for Touchables

EyMaddis:
By default, React Native for Web handles some edge cases for Touchables which lead to unexpected behaviour.
Example:
```jsx
<TouchableOpacity
 onClick={() => console.log('never called')}
 accessibilityRole="link"
>
```
vs 

```jsx
<TouchableOpacity
 onClick={() => console.log('called')}
 accessibilityRole="button"
>
```

For key events its even more extreme and might block propagation, but only if ENTER was used.

A little bit of context: 
I develop a SmartTV App with React Native Web, in which I implemented spatial navigation (via arrow keys) in which I need to be able to globally handle ENTER events, this allows for a workaround. 
What would already work help would be the removal of `e.stopPropagation()`. I was unsure whether this solves a specific problem, so I chose the opt-out method instead.

This makes [my workaround](https://github.com/necolas/react-native-web/issues/1219#issuecomment-491017893) from #1219 usable.

### DIFF
--- a/packages/react-native-web/src/exports/Touchable/index.js
+++ b/packages/react-native-web/src/exports/Touchable/index.js
@@ -800,6 +800,15 @@ const TouchableMixin = {
     const ENTER = 13;
     const SPACE = 32;
     const { type, which } = e;
+    const { onKeyUp, onKeyDown } = this.props;
+
+    let eventHandlerFromProps: ?(e: Event) => any = null;
+    if (type === 'keyup') {
+      eventHandlerFromProps = onKeyUp;
+    } else if (type === 'keydown') {
+      eventHandlerFromProps = onKeyDown;
+    }
+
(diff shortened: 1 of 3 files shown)


necolas:
If you're listening to global events outside of React you can use the capture phase instead to work with events before they reach the bubble phase. No plans to add more events to the comments

EyMaddis:
I am not sure I understand this correctly. 
This contains two problems:
1. onClick is blocked
2. Propagation is blocked.

I understand your comment to only relate to 2. However, Issue 1 is more crucial as there is no way around it because the `createElement` will block the handler.

I could would be fine (under minor protest :) ) to leave out the propagation part, but blocking click for links is wrong in just wrong in my opinion.

PR #1350, opened 2019-05-18
[docs] heading => header for consistency with react-native

sstur:
This is a followup to https://github.com/necolas/react-native-web/commit/39d2e18ccfda6e6fd27f01f61198c27d2281107b.

Update the docs to match [the React Native docs](https://facebook.github.io/react-native/docs/accessibility#accessibilityrole-ios-android).

RN uses the term header instead of heading. RNW can accept either one, but we should probably encourage the use of the one that will work across all platforms.


necolas:
I think this is an example of where the RN API should be changed, as `header` is a different semantic tag on the web. So although RNW maps `header` to `heading`, I don't want to confuse web engineers. People running existing RN code on the web won't have issues.

PR #1351, opened 2019-05-22
StyleSheet: enhance hairlineWidth

MoOx:
This change avoids `hairlineWidth` to look fat on high density screen, which defeat the purpose. 

Note that this can look as a breaking change: with current (RN) implementation, a border-top of 0.33333333px (on a device with a device px ratio to 3) can be invisible (eg: safari ios).

If we want to avoid this, we could just add something like

```js
// some browser render as 0 something like 0.3333333333, instead of 0.34...)
const hairlineWidthMaxPrecision = 6;
const hairlineWidthMaxPrecisionTrick = Math.pow(10, hairlineWidthMaxPrecision)
hairlineWidth = Math.round(hairlineWidth * hairlineWidthMaxPrecisionTrick)/hairlineWidthMaxPrecisionTrick;
```

### DIFF
--- a/packages/react-native-web/src/exports/StyleSheet/StyleSheet.js
+++ b/packages/react-native-web/src/exports/StyleSheet/StyleSheet.js
@@ -7,6 +7,7 @@
  * @flow
  */
 
+import PixelRatio from '../PixelRatio';
 import ReactNativePropRegistry from '../../modules/ReactNativePropRegistry';
 import flattenStyle from './flattenStyle';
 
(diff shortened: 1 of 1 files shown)


necolas:
hairlineWidth was previously removed because it was becoming invisible too often - https://github.com/necolas/react-native-web/commit/3f8624e25f644c305e923a6776acabc2f99e70a4

MoOx:
Oh. I guess that why I am currently using the trick I mentioned :)
Feel free to close if you know the cause is desperate to get a proper hairlineWidth in various browsers.
I guess a comment on top of `hairlineWidth: 1` would help the next one trying to improve the situation to think (& test) twice.
I have very few phones around & don't want to spend time on testing this as it's not a big deal (I can currently live with the fact that "some phones" won't have a line where I am using this (on some personal projects)

MoOx:
I created #1488 to make this clear for futur people wondering about the implementation.

PR #1353, opened 2019-05-27
Fix handling of onClick event for accessibilityRole="link"

EyMaddis:
This is a separation from #1348 in order to separate the discussion of two separate issues.
By default `onClick` and other DOM based props are passed down to the element like `div`s, but not for links!

This is currently not in the documentation, but used in [this example](https://gist.github.com/necolas/f9034091723f1b279be86c7429eb0c96) and in the [official guides](https://github.com/necolas/react-native-web/blob/master/docs/guides/web-recipes.md) (here: `onMouseEnter/Leave`) from @necholas!

Using `onClick` is crucial when **avoiding** Touchables/ScrollViews in to not inherit issues from #1219. Mobile browsers already solve scroll-based issues if we use regular click events. 
This PR aims to allow just that and fix the link handling, as all other RNW elements support the `onClick`.

This results in problems like:
```jsx
<View
 onClick={() => console.log('called')}
 accessibilityRole="button"
>
```

```jsx
<View
 onClick={() => console.log('never called')}
 accessibilityRole="link" // want to use for accessibility and SEO reasons
>
```

### DIFF
--- a/packages/react-native-web/src/exports/createElement/__tests__/index-test.js
+++ b/packages/react-native-web/src/exports/createElement/__tests__/index-test.js
@@ -24,6 +24,45 @@ describe('modules/createElement', () => {
     });
   });
 
+  test('gracefully handles onClick for links', () => {
+    let wasCalled = false;
+    const component = shallow(
+      createElement('div', {
+        accessibilityRole: 'link',
+        onClick: () => {
+          wasCalled = true;
+        }
+      })
+    );
+    component.find('a').simulate('click', {
+      isDefaultPrevented() {},
+      nativeEvent: {},
+      preventDefault() {}
+    });
+
+    expect(wasCalled).toBe(true);
+  });
+
+  test('gracefully handles onClick for touchable links', () => {
+    let wasCalled = false;
+    const component = shallow(
+      createElement('div', {
+        accessibilityRole: 'link',
+        onResponderRelease: () => {}, // fake touchable
+        onClick: () => {
+          wasCalled = true;
+        }
+      })
+    );
+    component.find('a').simulate('click', {
+      isDefaultPrevented() {},
+      nativeEvent: {},
+      preventDefault() {}
+    });
+
+    expect(wasCalled).toBe(true);
+  });
+
   describe('prop "accessibilityRole"', () => {
     test('and string component type', () => {
       const component = shallow(createElement('span', { accessibilityRole: 'link' }));
--- a/packages/react-native-web/src/exports/createElement/index.js
+++ b/packages/react-native-web/src/exports/createElement/index.js
@@ -70,12 +70,16 @@ const adjustProps = domProps => {
   // element. Click events are not an expected part of the React Native API,
   // and browsers dispatch click events that cannot otherwise be cancelled from
   // preceding mouse events in the responder system.
+  // Handle overrides gracefully if the user provides an onClick method.
+  // If this is provided, the user must make sure to handle the edge cases.
   if (isLinkRole && onResponderRelease) {
-    domProps.onClick = function(e) {
-      if (!e.isDefaultPrevented() && !isModifiedEvent(e.nativeEvent) && !domProps.target) {
-        e.preventDefault();
-      }
-    };
+    domProps.onClick =
+      onClick ||
+      function(e) {
+        if (!e.isDefaultPrevented() && !isModifiedEvent(e.nativeEvent) && !domProps.target) {
+          e.preventDefault();
+        }
+      };
   }
 
   // Button-like roles should trigger 'onClick' if SPACE or ENTER keys are pressed.


loudwinston:
This is a great fix!  Thanks @EyMaddis !

Any chance that this can get some attention from project maintainers @necolas ?  We're having to fork `react-native-web` to apply this patch for our project.  It'd be great to include this in the official version.

tafelito:
one problem I see with that solution is that you manually need to handle propagation. If you have a Touchable inside another touchable, the onPress only gets called once. Using onClick you have to stop propagation depending on your use case it might just work adding in to the onClick event

necolas:
Why are you using `onClick` at all?

loudwinston:
@neclos we are using `onClick` to get around an issue with `onPress` where scrolling in a mobile brower sometimes triggers the `onPress` event.

I believe there's an issue for this specifically but I'm on mobile at the moment so having trouble finding it

EyMaddis:
Hi @necolas, thanks for taking a look! :) 

Like @loudwinston, I avoid ScrollViews and Touchable. I just had [too many issues](https://github.com/necolas/react-native-web/issues/1219), and want an app that feels native to the web. By that I do not mean an iOS/Android app that also runs on the browser, but a fully fledged "regular" web app.
This also includes body scrolling and how Chrome/Safari handle touch on clickable elements.

In addition to that, the app is also a Smart TV app which triggers click events when using the "OK" button on the remote control.
The main issue is with responder system, but I don't see that getting fixed in the few weeks before release. 

I did try to do it properly, e.g. I actually have a body scroller prototype locally, but it does not solve the broader issues.

Generally my opinion is that RNW enhances the web platform and should still allow the web apis to be usable. So while this might seem like a hack, I sincerely do not feel that way for the click event since it should be exposed anyway. I find it extremely irritating that the click handling depends on what element is used. If `onClick` is disallowed, it should not work at all (which I hope never happens!) and not for a specific case with a "random" `e.preventDefault()`.

Thanks

loudwinston:
@eymaddis @necolaa to clarify, the application I am working on does use Touchables and ScrollViews.  We had to start using `onClick` to avoid inadvertently triggering touch events on Touchables in ScrollViews when scrolling on the web.  We had to fork RNW to remove this `e.preventDefault` behavior because we use `accessibilityRole="link"` extensively to ensure screen readers (iOS Voiceover and Android TalkBack) announce our Touchables as links for users that navigate the web version of our app.

There's some refactoring we want to do to make these Touchables into actual links on the web, but the app started as a native app and was modified to also work on the web at a later time, so a lot of the code is very native focused at this time.

It would be preferable if we didn't have to fork RNW to make this behavior work for us.  I don't know the details of the changes that are being made to the responder system or how this `onClick` change would impact them, but perhaps this is a small change that  could be introduced in the current version to while the responder system changes are still in progress?

EyMaddis:
@necolas any chance you can take another look at this? :) 
I need to deploy a fork of RNW right now for a rather trivial fix.

necolas:
None of this is stable, so relying on `onClick` is not a good idea. For example, React Native recently added an `onClick` prop to `View` for keyboards, which has different semantics to `onClick` on the web.

EyMaddis:
Ok, interesting and understandable.
However right now, there is no other way to fix the issues. It seems that regular views are always part of the responder system so it seems to be impossible to use their functionality.

I do not see the responder system getting fixed properly in the next few weeks. I tried, but quickly noticed that this is a deep dive to reacts internals.
All that in order to let react native web reimplement the click handling that mobile browsers include natively anyway. 

On the other hand, the lines right after that fix officially support and use `onClick` for button elements, however links are blocked:
https://github.com/necolas/react-native-web/blob/fb9301959079c706fcca5c092fb273f23731c884/packages/react-native-web/src/exports/createElement/index.js#L86-L96


I agree with the problem that react native introduces via its own `onClick`, but for me this does not invalidate this approach at all.

EyMaddis:
@loudwinston I now use https://www.npmjs.com/package/patch-package with the following:

You might use another version of RNW, if follow the instructions from patch-package and manually fix it in your project

```diff
diff --git a/node_modules/react-native-web/dist/exports/createElement/index.js b/node_modules/react-native-web/dist/exports/createElement/index.js
index 903f855..8152350 100644
--- a/node_modules/react-native-web/dist/exports/createElement/index.js
+++ b/node_modules/react-native-web/dist/exports/createElement/index.js
@@ -72,7 +72,7 @@ var adjustProps = function adjustProps(domProps) {
   // preceding mouse events in the responder system.
 
   if (isLinkRole && onResponderRelease) {
-    domProps.onClick = function (e) {
+    domProps.onClick = onClick || function (e) {
       if (!e.isDefaultPrevented() && !isModifiedEvent(e.nativeEvent) && !domProps.target) {
         e.preventDefault();
       }

```

PR #1356, opened 2019-06-01
Fix Groups Sorting in StyleSheet

giuseppeg:
While I was playing with this I found that groups are sorted by key (strings). 
If you have `["10", "1", "2"]` the result is `[1, 10, 2]`. Fixing this with the PR.


giuseppeg:
by the way are rules supposed to be deduplicated when they are inserted in different groups? 

```js
sheet.insert('.test {}', 2)
sheet.insert('.test {}', 4)
```

Right now it seems that the library deduplicates without showing a warning (failing silently). Should it throw an error or print a warning?

Issue #1026, opened 2018-07-07
Alert: implementation


necolas:
https://facebook.github.io/react-native/docs/alert

piranna:
https://github.com/piranna/react-native-web-extended/blob/master/src/apis/Alert/index.js

RichardLindhout:
For now I do something like this to workaround this issue.

```
const alertTitle = 'Foutmelding'
const alertText = 'Geen toegang tot locatiedata'
if (Platform.OS === 'web') {
   alert(alertText)
} else {
    Alert.alert(alertTitle, alertText)
}
```

piranna:
Browser `alert()` is blocking, so it's not recomended to use. Not sure why it's not deprecated yet...

vlimag:
Does react-native-web supports Alert already?

steida:
@vlimag Check readme https://github.com/necolas/react-native-web#modules

arqex:
What do you think about implementing it using react portals? We can inject the node in the dom and use a portal to show/hide.

We can ship simple styling for it and create a way to let the developer customize the appearance and the appearing/hiding transition.

vlimag:
What I've done so far.
one file for default react native alert, and one file for react web alert.
Alert. js
```
import { Alert } from 'react-native'

export const StdAlert = (title, desc, onPress = () => { }) => {
  Alert.alert(
    title,
    desc,
    [
      { text: 'OK', onPress: () => onPress() }
    ],
    { cancelable: false }
  )
}

export const BinaryAlert = (title, desc, onPositivePress = () => { }, onNegativePress = () => { }) => {
  Alert.alert(
    title,
    desc,
    [
      { text: 'Sim', onPress: () => onPositivePress() },
      { text: 'Não', onPress: () => onNegativePress() }
    ],
    { cancelable: false }
  )
}
```

and for web, Alert.web.js


```
export const StdAlert = (title, desc, onPress = () => { }) => {
  alert(`${title}\n${desc}`)
  if (onPress) onPress()
}

export const BinaryAlert = (title, desc, onPositivePress = () => { }, onNegativePress = () => { }) => {
  const res = window.confirm(`${title}\n${desc}`)
  if (res) onPositivePress()
  else onNegativePress()
}
```
Then you use it as following:

```


```

joan-saum:
@vlimag , how about dependencies which use the Alert from react native ? (#1366)

vlimag:
[...]

PR #1364, opened 2019-06-10
Alert API

arqex:
Hey! I needed the alert API working for a project and I saw that it wasn't implemented (https://github.com/necolas/react-native-web/issues/1026) so I gave it a try. 

It follows the original API, tries to be simple, to play well with accessibility and to be customizable enough to adapt to every web project.

It comes with automated tests, flow types, and I have added some examples to the website. You can see it working here:

https://rnw-alert.surge.sh/?selectedKind=APIs&selectedStory=Alert

## Implementation
At first, I thought that the right thing would be using react portals. But I quickly discarded the idea since the API contains just a static method, and trying to mount alerts in our components makes no much sense.

The selected approach appends a new node to the `document.body` whenever we call `Alert.alert()`. In the append node we mount a component called `<AlertOverlay>`, that will be responsible to show/hide the alert gently to the users. We also apply some other tricks to make the alert content the only accessible one in the page (more about it in the accessibility section below).

The alert content is shown by a different component. In the implementation the default alert component is called `<AlertDefaultComponent>`, and it tries to be really simple, using react-native's `<Button>` component to render the different options the alert provides. The default styles, the alert component and the buttons are customizable to adapt to any web app design (see more at the customization section below).

There is a small default animation when the alert is getting open/closed. The implementation uses RN's `Animated` library to make it work.

In Android, alerts are dismissable by tapping outside of the box, on the overlay background. This implementation inherits that behavior by default, but we can disable it by using `{cancelable: false}` as the options argument of the `Alert.alert()` method. See https://facebook.github.io/react-native/docs/alert for more info about the arguments.

In iOS, alert can have more than 3 buttons and this implementation allows the same. On the other hand, in Android the order of the buttons has a meaning - 'neutral', 'negative', 'positive' - but this implementation really doesn't care about what are the purposes of the given buttons.

## Accessibility
When we call `Alert.alert()`, this implementation ensures that the alert's content is the only one accessible in the screen, even if the current page's content is visible behind. When the alert is open:

* The background content is hidden to screen readers thanks to the `aria-hidden=true` attribute.
* The background content is not scrollable anymore.
* The background content is not selectable by dragging the mouse click, using `userSelect=none`.
* The focus is moved into the alert. When the alert get closed, the focus is moved back to the previous selected element.
* The content of the alert is the only one focusable by pressing the TAB key.
* Pressing the ESC key, dismiss the alert if `cancelable` is not set to `false`.

All the accessibility changes above are reverted when the alert is closed, leaving the screen as it was before opening the alert.

## The back button
Even if we set the option `{cancelable: false}` for our alerts, users will be always able to cancel the alert in web apps by pressing browser's back button. This implementation of the alert API is aware of changes in the URL:
* When the alert is open and the URL changes, the alert is dismissed since it might not be related to the context anymore, and the `onDismiss` callback is called.
* When we call `Alert.alert()`, it tries to add a parameter to the URL so, when we press the back button, the browser will just remove the parameter and we dismiss the alert without changing the main content of the page.
* When the alert is closed because the user pressed a button (or dismissed by clicking the overlay), we go back in the browser's history removing the added URL parameter. This approach has the downside of rendering the browser with an URL in the browser history's future, but it was the best one I could imagine. If you find out anything better, I'd be glad to implement it.

## Customizing the styles
In mobile, the alerts are native and follow the styles of the platform so they always look ok. In web apps, there are no predefined styles for them, and user expects their styles to match the application ones.

There isn't a standard way of applying styles to native components in RNW (as far as I know), so I decided to let the developer define some static methods or attributes in the `Alert` object to override the styling defaults and internal components.

* **`Alert.getCustomStyles`** - If defined, the implementation tries to call this method when the alert is getting open, expecting an object with the styles to override as the result. Eg:
```js
// Before calling `Alert.alert()`
Alert.getCustomStyles = function( animatedValue ){
	return {
		alertWrapper: {
			backgroundColor: 'green'
		}
	};
}
```
We can override the styles for the following elements (from outside in):
`{wrapper, overlay, alertWrapper, titleWrapper, title, messageWrapper, message, buttonsWrapper, buttonWrapper}`

As we can see in the example above, the method will receive an `Animated.Value` as the only parameter. It's the one used for the appearing/hiding animations, as its value changes from `0` to `1`. We can interpolate this value to create animations in the `overlay` style (the only one animatable). Please, animate only `opacity` and `transform` attributes to keep the animation performant.

* **`Alert.Button`** - By default, this implementation uses the shipped component `<AlertDefaultButton>` to render the different clickable options of the alert. This component uses RNW's `<Button>` internally. If we have customized buttons in our app, we might want to use them also in our alerts. We can tell Alert to use custom components for its buttons by setting `Alert.Button`:
```js
// CustomButton is our button component
// Before calling `Alert.alert()`
Alert.Button = CustomButton

// Our component will receive the props:
CustomButton.propTypes = {
	// The object returned by Alert.getCustomStyles
	customStyles: PropTypes.object,
	// The index of the button
	index: PropTypes.number,
	// The callback to be called when the button is pressed
	onPress: PropTypes.func,
	// The text of the button
	text: PropTypes.string,
	// The "style" of the button as seen the alert docs
	// the `style` prop has a special meaning, so it is renamed to `type`
	type: PropTypes.oneOf(['default', 'cancel', 'destructive'])
};
```
If we passed our component to `Alert.Button`, the `buttonWrapper` style returned by `getDefaultStyles` method won't be used unless we apply it in our component's code.

**`Alert.Component`** - If overriding the styles and the button component is not enough for our App, we can override the whole alert box. By default, the shipped component `AlertDefaultComponent` is used, but we can override it:
```js
// CustomAlert is our Alert component
// Before calling `Alert.alert()`
Alert.Component = CustomAlert

// The props received
CustomAlert.propTypes = {
	// The button component as set in `Alert.Button`
	Button: PropTypes.oneOfType([PropTypes.instanceOf(Component), PropTypes.func]),
	// The button definitions from the options
	buttons: PropTypes.array,
	// The styles as returned by `Alert.getCustomStyles`
	customStyles: PropTypes.object,
	// The message from the options
	message: PropTypes.string,
	// The title from the options
  title: PropTypes.string
};
```
If `Alert.Component` is used, we will receive the `customStyles` and the `Button` components as props, but it's up to use them.


hodgef:
@necolas Please merge this. `Alert` support would be amazing to have!

necolas:
Since your implementation doesn't rely on any internals you should maintain it in "userland". This patch is too big and complicated for me to review, especially without any prior discussion about the implementation (i.e., relying on Animated in a perf problem).

At the moment we're rethinking what code belongs in a core UI library and I'm unlikely to add new modules while so many are being removed from RN and reassessed more generally.

joan-saum:
@necolas What about the implementation in #305 ? I thought this PR was close to be merge but maybe got abandonned ?

arqex:
Sorry for the big PR, I had the alert API implemented and I just adapted it to work within RNW so I didn't give the chance of discussing. 🙏 

Knowing that, I tried to add a good explanation of it in the PR so we can discuss now all the opinionated parts. I have no problems in changing whatever we think that can be done better, if we still want to have the alert API in the RNW.

RN alerts are a strange mix, they are used as a static method but it also involves the UI components. Maybe RNW should include all the logic that ensures that any component will be shown in an acessible way and the UI itself should be provided by the developer.

ziaulrehman40:
@arqex Is there a possibility to get this out as a separate npm package, like we have a few for react-native-web-modal implementation. Would be great.
Thanks.

PR #1365, opened 2019-06-11
Do not stringify the app params

edelventhal:
We have a circular structure in our app params. This trivial console.log therefore causes our app to crash. We have no reason not to want a circular structure, and we also have no reason to log out all the startup params.

IMO, this seems like a console log you don't need.


necolas:
This patch would just print `[Object object]` instead. And this code is part of React Native - https://github.com/facebook/react-native/blob/ac7ec4602fd4d8ec033a084df6a92437ea2d3a8a/Libraries/ReactNative/AppRegistry.js#L180

edelventhal:
Yeah, you can inspect the value of it in the browser though. Or just don't print it at all.

Huh wow I didn't realize that React Native does it too. That's weak.

PR #1395, opened 2019-07-09
[fix] Hiding scroll support on Firefox ver. 64+

bcontrol:
Since Firefox ver. 63 we should use scrollbar-width property


necolas:
The `scrollbar-width` property should probably appear last in the list of rules

yungkittty:
@necolas Sorry, I wasn't aware there were already a pull-request opened about this issue. Do you know when this will be released? Thanks.

PR #1396, opened 2019-07-09
Add Animated.FlatList and Animated.substract

### LINKED ISSUE #1378: Add Animated.FlatList and Animated.substract
Since react native propose a Animated.FlatList component and a Animated.substract function, we should integrate them

I just manually added the Animated.FlatList component in my local react-native-web, like it is done with View and other component, and it works very well.
I didn't try for the substract function, but since there is already the add function, i assume this one will be really easy to do (copy and paste what it is done, and just change the symbol `+`to `-`)

joan-saum:
Resolve #1378


necolas:
Thanks but I've already done this in a private branch

Issue #1249, opened 2019-02-20
Deal with Google Chrome ignoring autocomplete off


necolas:
https://bugs.chromium.org/p/chromium/issues/detail?id=468153#c164

PR #1404, opened 2019-07-23
adding missing `autoCompleteType` to `TextInput`

chrisdrackett:
not used on web, just pull it out so it doesn't end up on the DOM element.

### DIFF
--- a/packages/react-native-web/src/exports/TextInput/index.js
+++ b/packages/react-native-web/src/exports/TextInput/index.js
@@ -193,6 +193,7 @@ class TextInput extends Component<*> {
       /* react-native compat */
       accessibilityViewIsModal,
       allowFontScaling,
+      autoCompleteType,
       caretHidden,
       clearButtonMode,
       dataDetectorTypes,


necolas:
This looks like a recent RN change that's equivalent to the existing web-only `autoComplete` I added to RNW. TBH I don't know why the prop name is different and will suggest to the RN team that we change it to match web.

necolas:
https://github.com/facebook/react-native/issues/26003

CareLuLu-Gabriel:
On the same thread, on `Google Chrome Version 76.0.3809.100 (Official Build) (64-bit)`, the `autocomplete="noop"` doesn't work while `autocomplete="off"` works. It seems that they are finally following the standard "on"/"off" https://github.com/necolas/react-native-web/issues/1249

necolas:
🙄 I wish they would stop messing around with this. I don't see any official comment or resolution on the open chromium issue though https://bugs.chromium.org/p/chromium/issues/detail?id=587466

Issue #237, opened 2016-10-23
Image: new babel/webpack loader


necolas:
We need a new webpack image loader that can generate RN source objects from assets on the file system:

``` js
require('./image.png');
// => { uri: '...', width: 40, height: 40 }
```

It should also support RN's syntax for resolution-specific images:

```
image.png
image@2x.png
image@3x.png
```

...by adding extra data to the object so that `Image` can choose the most appropriate asset:

``` js
require('./image.png');
// => { uri: '...', uri@2x: '...', uri@3x: '...', width: 40, height: 40 }
```

MoOx:
Going to try this https://www.npmjs.com/package/react-native-image-loader
(can't find the repo, the only reference I found is https://github.com/willpiers/react-native-image-loader from https://github.com/mjohnston/react-native-webpack-server/issues/6 but leads to a 404)

MoOx:
Ok false alarm, this module is just 5 lines of code and totally don't work as expected :(
Will try to do one then.

MoOx:
I did this pretty quickly and it seems to work for now

```js
const fs = require("fs")

const ImageFile = require("image-file")

module.exports = function(content) {
  const cleanPublicPath = content
    .replace("module.exports = ", "")
    .replace(/;$/, "")

  const path = this.resourcePath
  const image = new ImageFile(fs.readFileSync(path).buffer)

  return `module.exports = {
    uri: ${ cleanPublicPath },
    width: "${ image.width }",
    height: "${ image.height }"
  }`
}
```

Usage in webpack config:

```js
        {
          test: /\.(jpg|png|gif)$/,
          loader: "./src/loaders/react-native-web-image-loader.js!file-loader",
        },
```

necolas:
Nice start. How do we get webpack to extract any other resolution-specific images?

MoOx:
Not handled yet in my case, but I guess the next step would be:

```js
const fs = require("fs")

const ImageFile = require("image-file")

module.exports = function(content) {

  //
  // @todo instead of using original file path, we should just reuse some
  // file-loader logic directly to support hash for file paths etc
  //

  const cleanPublicPath = content
    .replace("module.exports = ", "")
    .replace(/;$/, "")

  const absoluteFile = this.resourcePath
  const image = new ImageFile(fs.readFileSync(absoluteFile).buffer)

  const path = require("path")

  const absoluteData = path.parse(absoluteFile)
  const base = absoluteData.dir + absoluteData.name

[...]

PR #1408, opened 2019-07-31
[Image] Get the asset scale which is closest to the window scale

EvanBacon:
https://github.com/necolas/react-native-web/issues/237

### DIFF
--- a/packages/react-native-web/src/exports/Image/index.js
+++ b/packages/react-native-web/src/exports/Image/index.js
@@ -18,6 +18,7 @@ import ImageResizeMode from './ImageResizeMode';
 import ImageSourcePropType from './ImageSourcePropType';
 import ImageStylePropTypes from './ImageStylePropTypes';
 import ImageUriCache from './ImageUriCache';
+import PixelRatio from '../PixelRatio';
 import StyleSheet from '../StyleSheet';
 import StyleSheetPropType from '../../modules/StyleSheetPropType';
 import View from '../View';
@@ -55,7 +56,7 @@ const resolveAssetUri = source => {
     const asset = getAssetByID(source);
     let scale = asset.scales[0];
     if (asset.scales.length > 1) {
-      const preferredScale = window.devicePixelRatio;
+      const preferredScale = PixelRatio.get();
       // Get the scale which is closest to the preferred scale
       scale = asset.scales.reduce((prev, curr) =>
         Math.abs(curr - preferredScale) < Math.abs(prev - preferredScale) ? curr : prev
--- a/packages/react-native-web/src/exports/Image/index.js
+++ b/packages/react-native-web/src/exports/Image/index.js
@@ -53,7 +53,14 @@ const resolveAssetUri = source => {
   if (typeof source === 'number') {
     // get the URI from the packager
     const asset = getAssetByID(source);
-    const scale = asset.scales[0];
+    let scale = asset.scales[0];
+    if (asset.scales.length > 1) {
+      const preferredScale = window.devicePixelRatio;
+      // Get the scale which is closest to the preferred scale
+      scale = asset.scales.reduce((prev, curr) =>
+        Math.abs(curr - preferredScale) < Math.abs(prev - preferredScale) ? curr : prev
+      );
+    }
     const scaleSuffix = scale !== 1 ? `@${scale}x` : '';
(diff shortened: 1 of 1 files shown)


packages/react-native-web/src/exports/Image/index.js
```diff
@@ -53,7 +53,14 @@ const resolveAssetUri = source => {
   if (typeof source === 'number') {
     // get the URI from the packager
     const asset = getAssetByID(source);
-    const scale = asset.scales[0];
+    let scale = asset.scales[0];
```
EvanBacon: we should cache the pixel ratio to prevent this from being called multiple times.

packages/react-native-web/src/exports/Image/index.js
```diff
@@ -53,7 +53,14 @@ const resolveAssetUri = source => {
...
+    if (asset.scales.length > 1) {
+      const preferredScale = window.devicePixelRatio;
+      // Get the scale which is closest to the preferred scale
+      scale = asset.scales.reduce((prev, curr) =>
+        Math.abs(curr - preferredScale) < Math.abs(prev - preferredScale) ? curr : prev
```
EvanBacon: We need to round the size so you always get the most crisp image.

packages/react-native-web/src/exports/Image/index.js
necolas: `PixelRatio` (via `Dimensions`) does this already

EvanBacon: I mean to prevent this computation from running every time the asset is requested.

necolas:
Please could you add unit tests for this feature?

ericvicenti:
1. This fix looks great for most apps
2. It will totally screw up server-rendering, because `window.pixelRatio` is only on the client, and images will loaded twice at different resolutions

So I agree this is the simplest and preferred option for normal client apps. But for SSR+client apps, maybe we could consider an implementation with `<img srcset=..`?

necolas:
This is why it needs to use `PixelRatio`, where the values can be set on the server based on client hints. `source` is also not rendered on the server, so this would only be for `defaultSource`.

ericvicenti:
Relying on client hints is quite imprecise, and it breaks/complicates caching.

Even if you could predict the window size from the server, the required resolution depends on the layout in the app. Maybe a larger image is justified on mobile if it takes up the full screen, in contrast to an icon. 

@necolas, you've been much deeper into web eng than me for the past few years. Can I ask why you're not interested in using srcset? For these reasons I think it is more appropriate for the browser to determine which resolution to download.

Here's a nice article on it.. https://medium.com/hceverything/applying-srcset-choosing-the-right-sizes-for-responsive-images-at-different-breakpoints-a0433450a4a3

MoOx:
I also think we should keep cached SSR in mind (static rendering) by using `srcset` somewhere... I know it will be harder since we are currently using a background image with a img overlay for accessibility, but maybe we could  go for a simpler `img` + `srcset` + `object-fit`, don't you think?

msand:
Would it be possible to expose the resolveAssetUri helper? Or some way to get the uri from a require('./img.ext') call? It's needed for implementing svg raster image embedding in the web support of react-native-svg

Current ugly workaround / hack: (There must be a better way)

https://snack.expo.io/@msand/react-native-svg-example
used in ./Examples/Image.js
defined in ./components/SnackImageCompat.web.js

```jsx
import React, { Component, Fragment } from 'react';
import { Image as NativeImage, View } from 'react-native';
import { Image as SvgImage } from 'react-native-svg';

class WebImage extends NativeImage {
  constructor(props, context) {
    super(props, context);
    this._setImageRef = ref => {
      this._imageRef = ref;
      const attrs = ref && ref.attributes;
      const src = attrs && attrs.src;
      const uri = src && src.value;
      this.setState({ href: uri });
    };
    this.oldRender = this.render;
    this.render = () => {
      const uri = this.state.href;
      return (
        <Fragment>
          <View
            style={{
              visibility: 'hidden',
              position: 'absolute',
              width: 0,
              height: 0,
            }}>
            {this.oldRender()}
          </View>
          <SvgImage {...this.props} href={uri} />
        </Fragment>
      );
    };
  }
}

export default props => (
  <WebImage {...props} source={props.source || props.href} />
);

```

EvanBacon:
Closing in favor of #1456

msand:
@EvanBacon Is there any other way to get the uri of images in snack / expo web?

PR #1413, opened 2019-08-07
Implement returnKeyType RN attribute as enterkeyhint

comp615:
**Problem**
React Native has a `returnKeyType` attribute which controls how the enter key looks and behaves. This was previously filtered out by RNW.

**Solution**
Map it through to the `enterkeyhint` attribute. Although this is not widely supported yet, it is part of an established standard. So it seemed prudent to implement it.

https://www.chromestatus.com/feature/5654529808793600
https://html.spec.whatwg.org/multipage/interaction.html#attr-enterkeyhint


necolas:
Thanks. I moved the prop-mapping logic from `createElement` to `TextInput`.

freekingg:
Why install the latest version, still report this error, what should I do?

```
Warning: Invalid DOM property `enterkeyhint`. Did you mean `enterKeyHint`?
```

PR #1415, opened 2019-08-09
Fix styles hydration when marker is in single quotes

### LINKED ISSUE #1317: Insertion order of inline styles vs global styles causes issues in Edge
**The problem**
At a high-level, Twemoji on Edge are broken on Twitter. We did some preliminary looking into this (thanks @krosenberg)...notes below, but we haven't done an isolated repro yet.

In both Chrome and Edge, the span that wraps an emoji has two class names that each reference a css rule containing a display property:
```
<span dir="auto" class="css-76zvg2 css-16my406 r-4qtqp9 r-ip8ujx r-sjv1od r-zw8f10 r-bnwqim r-h9hxbl">

.r-4qtqp9 {
    display: inline-block;
}

.css-76zvg2 {
    display: inline;
}
```
In Chrome, the computed style for the span is `display: inline-block`, but in Edge it is `display: inline`.

I took a look at the style sheet where these rules live, and noticed that the order of CSSRuleList differs between Chrome and Edge.

On Chrome `.r-4qtqp9` comes after `.css-76zvg2`, so the inline-block takes precedence:

```
Array.from(document.styleSheets[1].cssRules).map((x, i) => x.selectorText).indexOf(".r-4qtqp9")
>> 18
Array.from(document.styleSheets[1].cssRules).map((x, i) => x.selectorText).indexOf(".css-76zvg2")
>> 11
```
But on Edge, it's the opposite:
```
Array.from(document.styleSheets[1].cssRules).map((x, i) => x.selectorText).indexOf(".r-4qtqp9")
>> 9
Array.from(document.styleSheets[1].cssRules).map((x, i) => x.selectorText).indexOf(".css-76zvg2")
>> 18
```
I'm not sure how the order of the CSSRuleList is determined, but if it's based on when JS code adds rules to the list (I'm assuming this happens in RNW somewhere), then I wonder if there's a difference in the order in which the code is called based on the browser or the JS engine or something.

**How to reproduce**
Have not had time to create and test an isolated case yet. With Edge access this might be more clear, but hoping @necolas's previous experience in #1300 might help shed light.

**Expected behavior**
CSS rules should be the same and work across major browsers

* React Native for Web (version): 0.11+ (worked previously)
* React (version): 16.4
* Browser: Edge ONLY

giuseppeg:
Fixes #1317

It turns out that Edge uses single quotes for attribute selectors. This broke hydration on Edge.

### DIFF
--- a/packages/react-native-web/src/exports/StyleSheet/__tests__/createOrderedCSSStyleSheet-test.js
+++ b/packages/react-native-web/src/exports/StyleSheet/__tests__/createOrderedCSSStyleSheet-test.js
@@ -86,5 +86,25 @@ describe('createOrderedCSSStyleSheet', () => {
       const clientSheet = createOrderedCSSStyleSheet(element.sheet);
       expect(clientSheet.getTextContent()).toMatchSnapshot();
     });
+
+    test('works when the group marker is in single quotes', () => {
+      // Setup SSR CSS
+      const serverSheet = createOrderedCSSStyleSheet();
+      serverSheet.insert('.a { color: red }', 0);
+      serverSheet.insert('.b { color: red }', 1);
+      const textContent = serverSheet.getTextContent().replace(/"/g, "'");
+
+      // Add SSR CSS to client style sheet
+      element.appendChild(document.createTextNode(textContent));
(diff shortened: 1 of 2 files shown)


necolas:
Good work! thank you

PR #1426, opened 2019-08-29
[Image] Make image draggable by default.

### LINKED ISSUE #1425: Images are not draggable by default
**The problem**
The `Image` component should match native behavior by setting draggable to `true` by default.

**How to reproduce**
Render an `<Image />` component without explicitly setting the draggable prop. Note in the browser how `draggable` is set to false. Also, note [the code](https://github.com/necolas/react-native-web/blob/37ca236d095d039b48899720ff39a630f4f0dff2/packages/react-native-web/src/exports/Image/index.js#L277)

**Expected behavior**
Draggable is true by default.

evcohen:
Fixes #1425


necolas:
See #1425

PR #1428, opened 2019-09-10
Link: Allow explicit rel="opener"

sstur:
We shouldn't unconditionally set noopener on all links. There's valid use cases to open a new tab that _can_ communicate back using `postMessage()`

Also, this cleans up the string concatenation so that we don't get extraneous leading space or duplicate values.

### DIFF
--- a/packages/react-native-web/src/modules/createDOMProps/index.js
+++ b/packages/react-native-web/src/modules/createDOMProps/index.js
@@ -187,8 +187,12 @@ const createDOMProps = (component, props, styleResolver) => {
   // https://mathiasbynens.github.io/rel-noopener/
   // Note: using "noreferrer" doesn't impact referrer tracking for https
   // transfers (i.e., from https to https).
-  if (component === 'a' && domProps.target === '_blank') {
-    domProps.rel = `${domProps.rel || ''} noopener noreferrer`;
+  // Note: Specifying rel="opener" will explicitly bypass this safeguard.
+  if (component === 'a' && domProps.target === '_blank' && domProps.rel !== 'opener') {
+    const existingRel = domProps.rel ? domProps.rel.split(' ') : [];
+    // Ensure that we don't end up with duplicates.
+    const newRel = new Set([...existingRel, 'noopener', 'noreferrer']);
+    domProps.rel = Array.from(newRel).join(' ');
   }
   // Automated test IDs
   if (testID && testID.constructor === String) {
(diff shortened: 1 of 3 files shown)


packages/react-native-web/src/modules/createDOMProps/index.js
```diff
@@ -187,8 +187,12 @@ const createDOMProps = (component, props, styleResolver) => {
   // https://mathiasbynens.github.io/rel-noopener/
   // Note: using "noreferrer" doesn't impact referrer tracking for https
   // transfers (i.e., from https to https).
-  if (component === 'a' && domProps.target === '_blank') {
-    domProps.rel = `${domProps.rel || ''} noopener noreferrer`;
+  // Note: Specifying rel="opener" will explicitly bypass this safeguard.
+  if (component === 'a' && domProps.target === '_blank' && domProps.rel !== 'opener') {
+    const existingRel = domProps.rel ? domProps.rel.split(' ') : [];
+    // Ensure that we don't end up with duplicates.
+    const newRel = new Set([...existingRel, 'noopener', 'noreferrer']);
```
necolas: Creating a Set, array, then flattening it is unnecessary overhead on a hot path

sstur:
OK, sorry. What I meant by "blindly" was without letting the user opt-out. I might have used a poor word choice.

A simple use case would be opening a tab on the same domain that we would need to communicate back to the original tab.

Another case might be an oAuth flow where you open a tab to ask the user to go login using Google/GH/etc then that tab lands back on a same-origin page that sends a postMessage to `window.opener`.

I think noopener is a super wise default but we should probably be able to opt out because this essentially introduced a breaking change when I upgraded RNW and one that seems not to be able to be worked around.

packages/react-native-web/src/modules/createDOMProps/index.js
sstur: OK, I guess I thought this was the simplest way to prevent the duplicates that I saw in my DOM when I was using this. I don't know that it's a bug necessarily to have duplicates but I'm not sure that it's well defined in the spec what should happen if multiple conflicting or duplicates are in the `rel` attribute. I'll revert this change for a more performant one.

packages/react-native-web/src/modules/createDOMProps/index.js
```diff
@@ -187,8 +187,16 @@ const createDOMProps = (component, props, styleResolver) => {
...
+  // Note: Specifying rel="opener" will explicitly bypass this safeguard.
+  if (component === 'a' && domProps.target === '_blank' && domProps.rel !== 'opener') {
+    const relList = domProps.rel ? domProps.rel.split(' ') : [];
+    if (relList.indexOf('noopener') === -1) {
+      relList.push('noopener');
+    }
+    if (relList.indexOf('noreferrer') === -1) {
+      relList.push('noreferrer');
+    }
+    domProps.rel = relList.join(' ');
```
sstur: @necolas, how's this? We have only one array allocation now. Want me to keep it in pure string land?

sstur:
Hey @necolas, let me know if there's anything I can do to make this better and get it landed. Thanks!

necolas:
What I'm going to do is remove all the internal `rel` values and push that into user-space. Then you'll have full control

PR #1432, opened 2019-09-15
Linking API: Support for New Tab

shahzore-qureshi:
I added support for opening a link in a new tab.

The second parameter "target" can accept the following values:
"_blank" : URL is loaded into a new window or tab (default)
"_self" : URL replaces the current page

### DIFF
--- a/packages/react-native-web/src/exports/Linking/index.js
+++ b/packages/react-native-web/src/exports/Linking/index.js
@@ -22,9 +22,13 @@ const Linking = {
   getInitialURL(): Promise<string> {
     return Promise.resolve(initialURL);
   },
-  openURL(url: string, target: string): Promise<Object | void> {
+  openURL(
+    url: string,
+    target: '_blank' | '_self' = '_blank',
+    noopener?: 'noopener'
+  ): Promise<Object | void> {
     try {
-      open(url, target);
+      open(url, target, noopener);
       return Promise.resolve();
     } catch (e) {
       return Promise.reject(e);
@@ -36,16 +40,11 @@ const Linking = {
   }
 };
 
-const open = (url, target) => {
+const open = (url, target, noopener) => {
   if (canUseDOM) {
     const urlToOpen = new URL(url, window.location).toString();
-    const targetToUse = validateTarget(target) ? target : "_blank";
-    window.open(urlToOpen, targetToUse);
+    window.open(urlToOpen, target, noopener);
   }
 };
-
-const validateTarget = target => {
-  return target === "_blank" || target === "_self"
-}
 
 export default Linking;
--- a/packages/react-native-web/src/exports/Linking/index.js
+++ b/packages/react-native-web/src/exports/Linking/index.js
@@ -22,9 +22,9 @@ const Linking = {
   getInitialURL(): Promise<string> {
     return Promise.resolve(initialURL);
   },
-  openURL(url: string): Promise<Object | void> {
+  openURL(url: string, target: string): Promise<Object | void> {
     try {
-      open(url);
+      open(url, target);
       return Promise.resolve();
     } catch (e) {
       return Promise.reject(e);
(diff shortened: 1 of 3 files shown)


sw-yx:
instead of `target: string`, you can use string literals: `target: "_blank" | "_self`

shahzore-qureshi:
Good call @sw-yx! Thank you for the feedback.

EyMaddis:
I would propose a settings object as the second parameter, like:
```
Linking.openURL('some_url', { 
  target: '_blank', 
  rel: 'noopener' 
})
```

I think it is safe to assume that more properties might follow.

gianpaj:
Any update on this?

brunolemos:
I don't think this belongs at this repository. React Native only receives the `url` param, so react-native-web should be the same. You can easily implement this on user land. For example, in my case, I called it `Browser.openURL` instead of `Linking` and made it support the params I want.

shahzore-qureshi:
Sorry for the delay!

I agree @EyMaddis . I can add a settings object to that call.

@brunolemos - since react-native-web is a separate repository and is targeted for the web (and not mobile apps), I am okay with differentiating react-native from react-native-web.

Also, "Linking" on the web is quite different from "Linking" on mobile. Therefore, I believe the "Linking" module should have better support for the web. In addition, why even have the "Linking" module in the react-native-web repository if we end up using a made-up "Browser" module that you mentioned?

EyMaddis:
By the way, I totally agree with @brunolemos 

> I don't think this belongs at this repository. React Native only receives the url param, so react-native-web should be the same. You can easily implement this on user land.

shahzore-qureshi:
@EyMaddis - I do not agree with @brunolemos due to the reasoning in my last comment.

Alercard:
and if we use this code:
    if (Platform.OS==='web') {
      window.open(link, '_blank');
    }else{
      Linking.canOpenURL(link).then(supported => {
        if (supported) {
          Linking.openURL(link);
        } else {
          console.log("Don't know how to open URI: " + link);
        }
      });
    }

EvanBacon:
Perhaps this should be in a package like [`expo-web-browser`](https://docs.expo.io/versions/latest/sdk/webbrowser/) which supports opening external web pages on native (similar to creating a new tab).

soroushm:
its gonna be merge?

shahzore-qureshi:
@necolas and @avanwinkle - it looks like my original PR has finally been implemented as of this commit: https://github.com/necolas/react-native-web/commit/ffd300ebfd13343f7396b63b225ec2240630108c. May I ask what changed your minds on this topic? Why did you decide to implement the "target" parameter about 3 years after my request? And why was there a commit that hardcoded the target instead of allowing a target parameter? I am referring to this commit: https://github.com/necolas/react-native-web/commit/b7b383098aac42609166da38808a65497b5469a5. Thanks in advance!

shahzore-qureshi:
In addition, are there plans to officially move the React Native Web repo from your personal GitHub account to the official Meta/Facebook organization? I am concerned that this project is not receiving much attention because it is not officially under the Meta/Facebook organization. I see so much potential with this project, and I really want it to succeed.

kidroca:
I believe the reason to not accept your proposal earlier is because it didn't match the interface of Linking of react-native and it was relatively easy to code this in user land 

But later it was decided it's convenient to make linking open everything in new window/tab 
And this didn't change the interface 

But then users started to complain that they don't want to open everything in new tab 

So it seems in order to keep the feature of opening in new tab by default. But satisfy the rest that didn't want that a target parameter was added 

---

My observations are that developer conveniences aren't a priory for this project 
(Or there's simply too little time / maintainers)

necolas:
You're complaining that 1) you wanted this feature and it was added, and 2) that you want contributor PRs to be merged and a contributor PR was merged? What you're implying is not supported by facts.

This project has dozens of contributor PRs merged every release. If people submit good PRs with explanations, tests, etc., (as requested in the contributing guidelines) then it makes it a lot easier to review and merge. Unfortunately a lot of PRs lack context, change too many things for prompt review, and have little to no tests. And the person who improves (e.g., add the unit tests you didn't) and then maintains those contributor patches "forever" is me, not you.

The only reason this project is still going is because I keep it going. There's little to no support from Meta engineering. This project has to dance between supporting RN APIs that frequently had no thought for web, and what web devs want to do via React DOM with little thought for compat with React Native. Not to mention that I've put a lot of work into landing changes to React Native so more web code works out-of-the-box on React Native in the future.

Keep that in mind next time you feel like venting. Thanks in advance!

PR #1434, opened 2019-09-17
Only inject responder event plugins client side

### LINKED ISSUE #1433: SSR with server Webpack bundle error when injecting ResponderEventPlugin after hot update
**The problem**

We have a web app with SSR and two separate Webpack bundles, one for the client and one for the server; both these bundles support hot updates in development. On the client all works fine. The server works initially, but after the first update when reloading the page there is an error in `createElement.js` where `injectEventPluginsByName` is called during SSR:

```
Invariant Violation: EventPluginRegistry: Cannot inject two different event plugins using the same name, `ResponderEventPlugin`.
```

This can be fixed by only injecting event plugins on the client, which seems like it should be fine because event handling isn't relevant on the server:

```js
if (typeof window !== 'undefined') {
  injectEventPluginsByName({
    ResponderEventPlugin: ResponderEventPlugin
  });
}
```

Just adding the condition is sufficient to resolve the issue. Would this be a reasonable fix, or is there something else we're missing that might resolve this problem?

**How to reproduce**

Simplified test case: Since it relates to SSR on online reproduction isn't practical. We can create a minimal repo if needed.

**Expected behavior**

A hot updating server bundle including RNW code should work.

**Environment. Did this work in previous versions?**

N/A. Our setup worked until we incorporated the latest version of RNW.

* React Native for Web (version): 0.11.7
* React (version): 16.9.0
* Browser: Chrome (but N/A)

marnusw:
Fixes #1433

### DIFF
--- a/packages/react-native-web/src/exports/createElement/index.js
+++ b/packages/react-native-web/src/exports/createElement/index.js
@@ -14,9 +14,11 @@ import normalizeNativeEvent from '../../modules/normalizeNativeEvent';
 import React from 'react';
 import ResponderEventPlugin from '../../modules/ResponderEventPlugin';
 
-injectEventPluginsByName({
-  ResponderEventPlugin
-});
+if (typeof window !== 'undefined') {
+  injectEventPluginsByName({
+    ResponderEventPlugin
+  });
+}
 
 const isModifiedEvent = event =>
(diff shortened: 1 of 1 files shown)


EugeneHlushko:
Can this get merged please?

PR #1456, opened 2019-10-13
Multi resolution images

dcalhoun:
Continue the progress within #1408, by adding the [requested unit test](https://github.com/necolas/react-native-web/pull/1408#issuecomment-536686086). @necolas please let me know if there is additional coverage you would like or a different testing approach you would prefer.

Closes #1408.

### DIFF
--- a/packages/react-native-web/src/exports/Image/__tests__/index-test.js
+++ b/packages/react-native-web/src/exports/Image/__tests__/index-test.js
@@ -1,9 +1,11 @@
 /* eslint-env jasmine, jest */
 /* eslint-disable react/jsx-no-bind */
 
+import * as AssetRegistry from '../../../modules/AssetRegistry';
 import Image from '../';
 import ImageLoader from '../../../modules/ImageLoader';
 import ImageUriCache from '../ImageUriCache';
+import PixelRatio from '../../PixelRatio';
 import React from 'react';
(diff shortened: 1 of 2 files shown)


necolas:
Thanks for this. Would you like to have a go at writing a short bit of documentation for this? If so, just leaving it in a comment on this PR would be great as I can then include it in the new docs I'm working on.

dcalhoun:
@necolas I began to attempt to write documentation for this only to realize that these changes feel like merely a portion of the spirit of #237. I'm not sure if this warrants documentation yet.

If I'm not mistaken, these changes to `Image` rely on receiving a `source` prop that is a number that represents an ID of an asset within `AssetRegistry`. The function `registerAsset` is not used anywhere in this project (i.e. RNW does not automatically handle storing assets for you) and it is not exported (i.e. a consumer of RNW cannot manually register their own assets with the registry).

While this scale selection work is good progress towards automatically selecting the appropriate scaled image for a device's pixel density, it's merely a first step. I'm not sure how to write documentation for this without more of #237 being accomplished. Namely, the creation of a webpack/babel loader to allow requiring an image and having its various scaled assets (e.g. `@2x`, `@3x`) required alongside it.

Let me know if I am misunderstanding the state of things.

dcalhoun:
@EvanBacon any thoughts on my previous comment?

PR #1466, opened 2019-10-28
StyleSheet.create return object

jbrumwell:
React native moved away from returning a registry id when creating styles and return the object that was passed.  I ran into an issue where `react-native-elements` expected an object but was being passed a id number so themeing was broken.

Relative Commit: https://github.com/facebook/react-native/commit/a8e3c7f5780516eb0297830632862484ad032c10


brunolemos:
react-native-web does optimizations with StyleSheet using this unique id. If the lib you are using is expecting an object, pass the style to them using `StyleSheet.flatten(style)`.

jbrumwell:
@brunolemos `react-native-elements` theme provide merges styles from every component with the theme provided. Your suggestion would have us add `flatten(style)` to every style property in the app and would be inefficient on the native side that doesn't require it.

Since `react-native` made this change a year and half ago I assume we are going to run into a lot more issues like this one.

What are your thoughts on a solution where we can turn on / off integer being returned? We should also update the Readme to reflect that the Stylesheet api has this divergence.

jbrumwell:
@brunolemos I just noticed that Readme says version `0.55` this change was made in `0.56` of `react-native`, unless we plan to support a newer version of `react-native` it would actually be a break to integrate this pr.  We will maintain a fork until `react-native-web` supports a higher version of `react-native`

Issue #1019, opened 2018-07-07
Image: support HTTP properties for 'source' objects


necolas:
Image can support HTTP-Verb, Headers or a Body along with the image request. The image loader may need to be implemented using `XMLHttpRequest`. 

https://facebook.github.io/react-native/docs/images#network-requests-for-images

necolas:
Below is some detail on additional functionality this would fix (from https://github.com/necolas/react-native-web/issues/1080 by @brunolemos). I don't think it's possible using `new Image()` loading but should be straightforward with `XMLHttpRequest` loading:

```js
xhr.onreadystatechange = function () {
  if (this.readyState == 4 && this.status == 200) {
    src = xhr.responseURL;
  }
};
```

--------------------------------------------------------------------

### The problem

When using an Image URI that redirects to another one, the image is not rendered.

```jsx
// does not work:
<Image source={{ uri: 'https://twitter.com/brunolemos/profile_image?size=original', width: 100, height: 100 }} />;

// works fine:
<Image source={{ uri: 'https://pbs.twimg.com/profile_images/1029249022903676928/GgKLxkT4.jpg', width: 100, height: 100 }} />;
```

### How to reproduce

- https://codesandbox.io/s/88lj20kw02
- https://snack.expo.io/@brunolemos/react-native-image-http-redirect

### Expected behavior

HTTP redirect should also work on web since it works on mobile

**Environment (include versions). Did this work in previous versions?**

* React Native for Web (version): 0.8.9
* React (version): 16.4.2
* Browser: Chrome 68

### Screenshots

<div>
<img width="300" src="https://user-images.githubusercontent.com/619186/44493612-fdbceb80-a63e-11e8-8cb6-04007a2ab8fb.png" />
<img width="300" src="https://user-images.githubusercontent.com/619186/44493583-eaaa1b80-a63e-11e8-9df4-6454bddd8d54.png" />
</div>

pwiszowaty:
Hey guys, any update on this one?

BenMoveo:
Did someone mange to overcome this issue?

kidroca:
Opened a PR to support image sources with `headers` (so far only GET method support) but I can open a follow PR to support POST and other methods
- https://github.com/necolas/react-native-web/pull/2442

PR #1470, opened 2019-11-04
fix: support image source objects

jaulz:
This PR introduces support for Image source objects (related issue: https://github.com/necolas/react-native-web/issues/1019) which allows you to pass different headers or fetch data via POST method.

In the example I added a dynamic chart from Google Charts. Not directly related but to fill the row I added WebP and the mentioned redirect case in the issue.

image

Please let me know if anything is missing.

### DIFF
--- a/packages/react-native-web/src/exports/Image/ImageUriCache.js
+++ b/packages/react-native-web/src/exports/Image/ImageUriCache.js
@@ -7,23 +7,42 @@
  * @flow
  */
 
+import ImageLoader from '../../modules/ImageLoader';
+
+type ImageSource =
+  | string
+  | number
+  | {
+      method: ?string,
+      uri: ?string,
+      headers: ?Object,
+      body: ?string
+    };
+
 export default class ImageUriCache {
   static _maximumEntries: number = 256;
   static _entries = {};
 
-  static has(cacheId: string) {
+  static createCacheId(source: ImageSource) {
+    return JSON.stringify(ImageLoader.resolveSource(source));
+  }
+
+  static has(source: ImageSource) {
     const entries = ImageUriCache._entries;
+    const cacheId = ImageUriCache.createCacheId(source);
     return Boolean(entries[cacheId]);
   }
 
-  static get(cacheId: string) {
+  static get(source: ImageSource) {
     const entries = ImageUriCache._entries;
+    const cacheId = ImageUriCache.createCacheId(source);
     return entries[cacheId];
   }
 
-  static add(cacheId: string, displayImageUri: string) {
+  static add(source: ImageSource, displayImageUri: ?string) {
     const entries = ImageUriCache._entries;
     const lastUsedTimestamp = Date.now();
+    const cacheId = ImageUriCache.createCacheId(source);
     if (entries[cacheId]) {
       entries[cacheId].lastUsedTimestamp = lastUsedTimestamp;
       entries[cacheId].refCount += 1;
@@ -31,13 +50,14 @@ export default class ImageUriCache {
       entries[cacheId] = {
         lastUsedTimestamp,
         refCount: 1,
-        displayImageUri
+        displayImageUri: displayImageUri || ImageLoader.resolveSource(source).uri
       };
     }
   }
 
-  static remove(cacheId: string) {
+  static remove(source: ImageSource) {
     const entries = ImageUriCache._entries;
+    const cacheId = ImageUriCache.createCacheId(source);
     if (entries[cacheId]) {
       entries[cacheId].refCount -= 1;
     }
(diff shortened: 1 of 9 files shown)


necolas:
Thanks for the PR! The unit tests are currently failing but at a glance this direction looks good to me, and should allow onProgress to be implemented too. So I'll work with you to get this merged

jaulz:
@necolas great! 😊 I just fixed some minor issues and now all tests are running fine. Also, I added the `onProgress` prop but that triggers only once for the example. Maybe we can find a better example?

packages/react-native-web/src/exports/Image/ImageUriCache.js
```diff
@@ -7,57 +7,80 @@
...
+      body: ?string
+    };
 
 export default class ImageUriCache {
   static _maximumEntries: number = 256;
   static _entries = {};
 
-  static has(uri: string) {
+  static createCacheId(source: ImageSource) {
+    return JSON.stringify(ImageLoader.resolveSource(source));
```
necolas: I'm a little concerned about using `JSON.stringify` here. We could use a Map instead

packages/react-native-web/src/exports/Image/index.js
```diff
@@ -33,46 +32,28 @@ const STATUS_LOADING = 'LOADING';
...
-    // get the URI from the packager
-    const asset = getAssetByID(source);
-    const scale = asset.scales[0];
-    const scaleSuffix = scale !== 1 ? `@${scale}x` : '';
-    uri = asset ? `${asset.httpServerLocation}/${asset.name}${scaleSuffix}.${asset.type}` : '';
-  } else if (typeof source === 'string') {
-    uri = source;
-  } else if (source && typeof source.uri === 'string') {
-    uri = source.uri;
+const getCacheUrl = e => {
```
necolas: This looks a bit fragile. Why do we need to get the url from the native event?

packages/react-native-web/src/exports/Image/ImageUriCache.js
jaulz: Yeah, I don't like it either but I could not find a smart way. The advantage of this approach is that you will always get the cached content for a combination of method, uri, headers and body. If we use a Map instead and use the source as a key then we somehow first need to get the reference to the source which leads to the same issue again. Or did you have something else in mind?

packages/react-native-web/src/exports/Image/index.js
jaulz: If an image is loaded via XHR, the actual image `src` will be replaced by a blob:
https://github.com/necolas/react-native-web/blob/acd3b335518a362842ecfe006a796fcfd2c1a076/packages/react-native-web/src/modules/ImageLoader/index.js#L110

Since the `onload` event is triggered by the image itself, we can retrieve the new blob source in `Image` via the image `src`  in the event handler without passing additional parameters. Though, of course we could  also make this more explicit and pass the event and the new blob url as separate parameters in `onLoad`. What do you prefer?

packages/react-native-web/src/exports/Image/ImageUriCache.js
necolas: A `Map` can use objects as keys

jaulz: Yes, it allows object as keys but the reference is stored as a key and not the content of the object:
```ts
const map = new Map()
const key = { a: 1 }
map.set(key, true)

map.get(key) !== map.get({ a: 1 }) // true
```

necolas: Yeah that's a problem given how image sources are defined. I guess we're stuck with stringifying for now

packages/react-native-web/src/exports/Image/index.js
necolas: Can we store the blob when we get it, so we don't have to read it back from the native event?

jaulz: Yep, the blob will be stored once the image was loaded:
```ts
ImageUriCache.add(source, getCacheUrl(e));
```

Afterwards it shouldn't be triggered again as long as the source does not change. Or do you mean something else?

necolas: I'm wondering why we need to get the blob from the event object when we know what it is before that

jaulz: I just reworked that part and now the blob is returned directly from the `ImageLoader`.

matthewwillian:
Hey is there any update with this PR?

jaulz:
@necolas I am just back from vacation and saw that there are conflicting files. Though, I don't see any changes in the affected files and am not sure why it says there are conflicts. Any idea what's wrong here?

necolas:
The conflicts are because the docs were rewritten. And now there is also a7ab961d95de892e4b41c96ea945aa344fd49e50

humaidk2:
Hi, 
Any update on this?

necolas:
#2442 is a contemporary fix for this. Thanks

PR #1485, opened 2019-11-15
Update babel-plugin-tester

necolas:


necolas:
Any ideas why babel is producing output with escaped quotes? `\\"react-native-web/dist/exports/View\\"`

paularmstrong:
> Any ideas why babel is producing output with escaped quotes?
> 
> `\\"react-native-web/dist/exports/View\\"`

This is Jest and it's snapshots that are producing the double-escaped double-quotes. 

You can modify babel-plugin-tester to use single quotes like this:

```js
pluginTester({
  babelOptions: {
    // ...
    generatorOpts: {
      jsescOption: {
        quotes: 'single'
      }
    }
  },
  plugin,
  tests
});
```

PR #1493, opened 2019-11-25
Update LayoutPropTypes to include gridArea

omonk:
This PR adds `gridArea as a valid key to be passed into style props.

Working on a project using grid we've come across this warning:

```
Warning: Failed prop type: Invalid props.style key `gridArea` supplied to `View`.
Bad object: {
  "gridArea": "2 / 2 / 3 / 3"
}
```


necolas:
Shorthand properties aren't supported in general

omonk:
Hi @necolas - just curious but is there a specific reason for shorthand not being supported?

PR #1498, opened 2019-12-05
[fix] Make ScrollView accept value 'zero' in scrollEventThrottle

LuisMiguelFilipe:
`scrollEventThrottle` default value is `0` but the `ScrollView` is currently logging if the default value zero is being used.

Reference: https://facebook.github.io/react-native/docs/scrollview#scrolleventthrottle


LuisMiguelFilipe:
Hello @necolas,
I'd just like to know if there is any action pending that i should do or is it just busy?

Thanks

PR #1500, opened 2019-12-08
Typescript guide to get started

yairopro:


packages/docs/src/introduction.stories.mdx
```diff
@@ -42,10 +42,43 @@ to setup a simple, web-only React app with built-in support for aliasing
...
+npm install --save react-native-web
+npm start
+```
+
+**With typescript:**
+```
+npx create-react-app my-app --template typescript
+cd my-app
 npm install react-native-web
+npm install --save-dev @types/react-native
```
necolas: Are these types correct for react-native-web though?

yairopro: RN-web has the same signatures as RN, doesn't it ?
I might be wrong. But which component or module would have a different interface ?

necolas: No, the types are not exactly the same, e.g., for styles. This is true of every fork of RN

yairopro: Ah ok

PR #1511, opened 2019-12-20
Allow style zIndex 'auto' value

hbj:
The style 'z-index' accepts also the value 'auto' and this PR adds it to the valid values.


necolas:
Thanks but at the moment it's intentional that auto isn't supported and that every view creates a new stacking context so that zindex is always relative to the parent

hbj:
Is this to be on par with the behaviour of React Native? The reason why we needed this is that we are using a component that displays correctly on mobile but has a display issue on web that can be solved by setting the z-index to 'auto', so I thought it would make sense if we would be able to apply web specific properties when our components are used on the web, i.e. by using `Platform.select()`.

PR #1530, opened 2020-01-26
Mark text div as user-select:text by default

SamyPesse:
When building an application with `react-native-web`, for example as an Electron app, we sometimes add a `body { user-select: none; }` CSS rule so that the user selection act like on a native app.

In that case adding the prop `selectable` to `<Text>` has no effect because `react-native-web` makes the assumption that the parent will be selectable.

With this PR, I think `react-native-web` will act more like `react-native` and enforce the text are selectable if the opposite is not specified.

### DIFF
--- a/packages/react-native-web/src/exports/Text/index.js
+++ b/packages/react-native-web/src/exports/Text/index.js
@@ -45,7 +45,8 @@ class Text extends React.Component<TextProps> {
     supportedProps.style = [
       style,
       numberOfLines != null && numberOfLines > 1 && { WebkitLineClamp: numberOfLines },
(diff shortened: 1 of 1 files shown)


SamyPesse:
**Update**: I've changed the logic to mark the text as selectable only if explicitly told so by the prop `selectable`: https://github.com/SamyPesse/react-native-web/commit/6a930af4b96aec0b76bac83b57f83c0eba031e12
It'll avoid impacting existing applications with a `user-select: text` if the prop `selectable` was not used.

necolas:
> we sometimes add a `body { user-select: none; }`

I don't think you should do this as it breaks accessibility expectations on the web. It is intentional that all text is selectable by default

PR #1539, opened 2020-02-20
fix injectEventPluginsByName throwing when re-evaluating code

vovaguguiev:
We (framer.com team) ran into this problem when investigating the errors thrown by the projects that were using `react-native-web`. Framer projects have the following setup:

- The project code is bundled with webpack while declaring `ReactDOM` as an external that is provided by Framer X during the evaluation of the produced bundle
- On every change to the user code, we re-bundle and re-evaluate the code but the instance of `ReactDOM` persists between those re-evaluations.
- This makes `injectEventPluginsByName` throw the error: `EventPluginRegistry: Cannot inject two different event plugins using the same name...` 

I'm happy to discuss alternative fixes is this one looks too hacky :)


necolas:
Thanks. Seems fine for now

Issue #1334, opened 2019-04-27
IE 11 Object.values problem


RooQI:
Hello.

**The problem**

In the file createStyleResolver.js and `Object.values` are not supported in IE11. You need to replace them with a supported combination `Object.keys(obj).map(function(e)`. 

**How to reproduce**

Open any example in IE11

**Expected behavior**

Instead of a blank page, the content should appear at startup.

**Environment (include versions). Did this work in previous versions?**

* React Native for Web (version): 0.11.2
* React (version): 16.8.4
* Browser: IE11

canaanites:
@RooQI @necolas any updates on this? We have the same problem..

PR #1541, opened 2020-02-24
Fix: Remove usages of Object.values, improve older browser compatibility

editkid:
Hello. I found an old issue https://github.com/necolas/react-native-web/issues/1334 which had a fix applied _but only partially_ it would appear. This PR simply applies the exact same fix in two other places in the same file.

Replace usages of `Object.values` to ensure compatibility with older browsers (without polyfilling). Ensures that IE10, IE11, Safari 8 etc. won't choke.

Hope this helps :v:


necolas:
Thanks

PR #1552, opened 2020-03-02
Support reason-react

jfrolich:
With Reason-React it is not possible to provide props with a "dash" (https://reasonml.github.io/reason-react/docs/en/invalid-prop-name).

This PR will make the `ariaLevel` prop work with react-native-web and `reason-react-native`. 

cc @MoOx


necolas:
This won't be a problem when https://github.com/necolas/react-native-web/commit/610b086566af5039f8ee9b9b4a4c8ef493c57784 lands. Not going to make changes in the meantime

jfrolich:
Ah was not aware of that commit, but that would solve it. Appreciated!

PR #1555, opened 2020-03-05
I18nManager: detect preferred language direction from document.dir

nickeng:
This commit updates I18nManager so it respects existing document.dir, this is useful for pages that contain server loaded RTL content with document.dir set appropriately. Not that its difficult to call `setPreferredLanguageRTL(true)` before `runApplication()` but there's no code like no code 😄


necolas:
I think I'll leave things as they are to avoid confusion

PR #1557, opened 2020-03-09
Allow compat mode for ViewPropTypes

simonbuerger:
Based on your comment in the release notes for 0.12.0 I'm not sure if you 'll be a fan of adding this back strictly for compat with libs that still use it? Just so that it doesn't crash

### DIFF
--- a/packages/react-native-web/src/exports/ViewPropTypes/index.js
+++ b/packages/react-native-web/src/exports/ViewPropTypes/index.js
@@ -1,0 +1,1 @@
+export default {};
--- a/packages/react-native-web/src/index.js
+++ b/packages/react-native-web/src/index.js
@@ -69,6 +69,7 @@ export { default as Settings } from './exports/Settings';
 export { default as Systrace } from './exports/Systrace';
 export { default as TimePickerAndroid } from './exports/TimePickerAndroid';
 export { default as TVEventHandler } from './exports/TVEventHandler';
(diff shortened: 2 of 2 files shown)


dehypnosis:
what about ```js
import PropTypes from "prop-types";

export default {
   style: PropTypes.any,
};
`

Some libraries of my project (may many of others) refer to VIewPropTypes.style directly..
`react-native-gallery-swiper`, `react-native-image-transformer`, `react-native-easy-view-transformer`, `react-native-page-list`, ...

Or proxy pattern.. or just copy actual viewPropTypes for more concrete compatibility?

EvanBacon:
I've migrated all of the Expo modules off of `prop-types` in favor of TypeScript to better support RNW 12+. I'd recommend that the packages still using `prop-types` remove their dependance since they will eventually break when `react-native` removes the deprecated `ViewPropTypes` module.

necolas:
I won't be adding deprecated APIs back. We don't use propTypes internally and neither do the biggest external consumers of this library.

matias91:
Right now the modules using `prop-types` works just fine in react native. So, I have a couple of doubts:

This module isn't supposed to behave exactly like `react-native`?
It is worth to block an entire platform instead of showing a warning?
There is a way to by-pass this without removing the modules that **are working perfectly** in `react-native`?

Having the development blocked due to this errors in third party modules is kind of frustrating.

PR #1573, opened 2020-03-31
Update supportedProps to include inputMode

dylan-gregory:
The inputMode property is exceptionally useful on native apps to render the correct type of keyboard necessary to fill in the type of input you're using. It is a globally supported prop, but gets filtered out at the moment.

### DIFF
--- a/packages/react-native-web/src/exports/View/filterSupportedProps.js
+++ b/packages/react-native-web/src/exports/View/filterSupportedProps.js
@@ -17,6 +17,7 @@ const supportedProps = {
   children: true,
   disabled: true,
   importantForAccessibility: true,
+  inputMode: true,
   nativeID: true,
   onBlur: true,
   onContextMenu: true,


necolas:
PRs should include more explanation. I'm not going to merge this because...

1. That filter list is for RN props. `inputMode` is not one of those.
2. The RN equivalent is `keyboardType`: https://reactnative.dev/docs/textinput.html#keyboardtype
3. This is implemented using `input[type]`, which sets the input mode as required.
4. Prop filtering is no longer being used in the `next` branch.
5. `inputMode` doesn't have broad enough browser support: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode

Thanks

dylan-gregory:
@necolas can you explain #3 from your comment?

necolas:
As per the link to `inputmode`:

> Inputs that require a telephone number should typically use `<input type="tel">` instead.

etc

dylan-gregory:
So my use case for inputMode was really just a workaround for not having access to the ReactNative decimal-pad (which is the phone-pad, but with a decimal - as seen here: https://lefkowitz.me/visual-guide-to-react-native-textinput-keyboardtype-options/), because I don't want it to be a telephone input. With the inputMode setting set to "decimal", I can make a number input show the number only without it incorrectly implying that it's a telephone input.

That being said - is there a reason you've intentionally omitted the decimal pad?

PR #1574, opened 2020-04-01
Deprecate AccessibilityInfo.fetch and add AccessibilityInfo.isScreenReaderEnabled

SuhairZain:
From this commit(https://github.com/facebook/react-native/commit/0090ab32c2aeffed76ff58931930fe40a45e6ebc) which was merged in 0.60.2, the fetch method is deprecated and the isScreenReaderEnabled method provides the same functionality.


ecreeth:
I think this should go in the **next** branch

ofirgeller:
@necolas Sorry to reopen this old issue, but I'm not sure it was fixed, unless the solution is that react-native-web does not polyfil this api and then it's up to the app developers.

PR #1577, opened 2020-04-06
Adds name property to CheckBox and Switch

jonsamp:
# Why

`CheckBox` and `Switch` both render `<input type="checkbox" />` components, however they do not add a `name` property to the native `<input>`, which makes it impossible for `<form>` components to read the control's value with its default `onSubmit` handler.

You can see how the default HTML form submission behavior works in [this snack](https://snack.expo.io/@jonsamp/form-checkbox-name).

# Changes

- Adds a `name` property to `CheckBox` and `Switch`, so that they can be used inside `<form>` components.
- Adds tests.
- Adds property in docs.

# Test plan

- View the docs and see the `name` property.
- Add a `name` property to `CheckBox` and `Switch` and ensure they resulting DOM element has a `name` property.


jonsamp:
After thinking about this more, I realize that this is only applicable to the web and not other react native platforms.

Issue #1538, opened 2020-02-14
The future of React Native for Web


necolas:
Over the next few months I'll be working through some significant changes to React Native for Web. These changes will be made on the [`next`](https://github.com/necolas/react-native-web/tree/next) branch. The motivation for these changes is to:

1. Update the library to use modern React features (e.g., Hooks) in preparation for Concurrent Mode.
2. Move away from depending on ReactDOM's `unstable-native-dependencies` export, which we'd like to remove from ReactDOM.
3. Resolve [long-standing issues](https://github.com/necolas/react-native-web/issues/1219) with the Responder Event Plugin.
4. Prototype high-level gesture systems for ReactDOM.
5. Simplify and improve the performance of the Touchable/Pressable components.

## Hooks rewrite (done)

Rewriting components to use Hooks is a prerequisite for all the other changes. Hooks simplify the implementations of components and offer an opportunity to resolve many existing bugs.

## Responder system rewrite (done)

Replacing the Responder Event System with a user-space implementation. See #1568 for more details.

## Touchables rewrite (done)

Better integrating the PressResponder with DOM expectations, to improve the UX of Touchables. See #1591 for more details.

## Canary releases

Most up-to-date canary is shown below. Please report regressions caused by canary releases. Post a comment below and include the canary version, as well as a codesandbox that reproduces the issue if possible.

### `0.0.0-d33e107ba` (5 June 2020)

* Changed: Remove `hitSlop` prop handling. Let browsers use their own automatic hitslop for touch interactions.
* Changed: Remove `TabBarIOS` and `TimePickerAndroid` exports
* Changed: Rewrite of the gesture responder system
* Changed: All components (except vendor ones) implemented using React Hooks. This build regresses `Image` caching, which will be reintroduced before a stable release.
* Changed: The `onLayout` prop now requires a `ResizeObserver` polyfill to work, and does not fallback to window `resize` events.
* Changed: Forwarding of `data-*` props is no longer supported. Use `dataSet` props, e.g., `dataSet={{ 'some-key': 1 }}`.
* Changed: Each component explicitly forwards supported props.
* Added: `Pressable`
* Added: `View` support for `accessibilityValue`.
* Fixed: `Image` support for variable resolution images (requires bundler integration).
* Fixed: `TextInput` support for `onContentSizeChange` to allow auto-grow textareas.
* Fixed: A limitation in setting styles using `ref.setNativeProps`.

## Open canary issues

[...]

PR #1588, opened 2020-04-12
[0.62 Roadmap] delete Long Press Error in Touchable 

ecreeth:
Original Summary: [This error is not actionable or valuable right now.](https://github.com/facebook/react-native/commit/9a3d722ccb523f227ffd7770a809996e6cfe75d9)


necolas:
The touchables are being rewritten so I won't be applying this patch. See  #1538

PR #1597, opened 2020-04-23
Appearance module for compatibility with RN 0.62

jfrolich:


packages/react-native-web/src/exports/Appearance/index.js
```diff
@@ -0,0 +1,43 @@
...
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ */
+import * as React from 'react';
+
+export type ColorSchemeName = 'light' | 'dark' | 'no-preference';
```
necolas, line 12: This file's export should be an object with the same API as the one in React Native: https://github.com/facebook/react-native/blob/f7b90336be25b78935549aa140131d4d6d133f7b/Libraries/Utilities/Appearance.js. It needs the change listener methods added.

packages/react-native-web/src/exports/Appearance/index.js
```diff
@@ -0,0 +1,43 @@
...
+  return window.matchMedia('(prefers-color-scheme: dark)');
+}
+
+const query = getQuery();
+
+export function getColorScheme(): ColorSchemeName {
+  return query && query.matches ? 'dark' : 'light';
+}
+
+export function useColorScheme(): ColorSchemeName {
```
necolas, line 25: Let's move this to the new "hooks" directory

PR #1600, opened 2020-04-27
Pass `disabled` props in TouchableWithoutFeedback

pveyes:
Recently upgraded to v0.12 and I notice `disabled` prop is no longer used in `TouchableWithoutFeedback` but it's used in `TouchableOpacity`. I think it should be added in `TouchableWithoutFeedback` as well. Let me know if I missed anything


jaulz:
@pveyes ah, I should have checked the PRs earlier. I was just about to raise the same PR when I saw yours :wink: Here is a basic demo of the issue:
https://codesandbox.io/s/react-native-web-lt5gi

necolas:
This is now fixed in master

PR #1616, opened 2020-05-14
[next] Fix host node measureLayout

janicduplessis:
The method on UIManager actually has the order of callbacks inverted compared to measureLayout.

This fixes measuring in VirtualizedList and maybe other things.


necolas:
Thanks. This was fixed

Issue #1568, opened 2020-03-27
Responder Event System for Web


necolas:
# Responder Event System for Web

The [Responder Event System](https://reactnative.dev/docs/next/gesture-responder-system) (RES) is a gesture system that manages the lifecycle of gestures. It was designed for React Native to help support the development of native-quality gestures. A pointer may transition through several different phases while the gesture (e.g., tap, scroll, swipe) is being determined, and a pointer may be used simultaneously alongside other pointers. The Responder Event System helps developers manage pointer interactions by providing a single, global “interaction lock” on views. For a view to become the “responder” means that pointer interactions are exclusive to that view and none other. A view can negotiate becoming the “responder” without requiring knowledge about other views.

## Using the Responder Event System

This system is currently provided via the [ResponderEventPlugin](https://github.com/facebook/react/blob/9def56ec0e1e71928ee999f48c00b1803ed8772a/packages/legacy-events/ResponderEventPlugin.js). The plugin and related injection helpers are also published in the ReactDOM npm package as a special bundle of unstable APIs under [`react-dom/unstable-native-dependencies`](https://unpkg.com/browse/react-dom@16.13.1/cjs/react-dom-unstable-native-dependencies.development.js). This bundle was created for, and is used exclusively by `react-native-web` to enable the RES in ReactDOM.

However, this setup has problems which are described below.

### Problems for React

1. The unstable APIs in the public release of ReactDOM are part of the legacy event system. The ReactDOM event system is being refactored and simplified, but OSS libraries depending on these unstable APIs will eventually block completion of this work.
2. The unstable APIs are not available in Facebook's internal build of ReactDOM. This blocks the experimental use of `react-native-web` by teams there.

### Problems for React Native for Web

1. Injecting the `ResponderEventPlugin` causes every host component (e.g., div) in ReactDOM to gain props for the Responder System.
2. The internals of the unstable APIs need to be mutated by `react-native-web` to monkey-patch issues with the implementation when it is running on the web. Occasionally, these internals have been changed in patches to React and caused unexpected downstream regressions.
[...]

PR #1643, opened 2020-06-19
fix scrollView&List Component press and slide conflict

wangjiafenghw:
Interrupting an event by determining if the focus has moved farther than 2 is compatible with Touchable in the scroll container to prevent scrolling from simultaneously starting the click event


necolas:
This issue is fixed in the canary channel by https://github.com/necolas/react-native-web/issues/1568

PR #1646, opened 2020-06-21
Support Modal Dialogs

### LINKED ISSUE #1020: Modal: implementation
Accessible modal implementation (in progress):

* Hides app content from screen-readers when active.
* Traps focus.
* Built-in `Escape`-to-close mechanism.

https://facebook.github.io/react-native/docs/modal

imnotjames:
Closes #1020 

This adds support for the [React Native Modal](https://reactnative.dev/docs/modal) on web.

While testing has been limited, this has been found to work for my use cases alongside [the React Native Community extensions of the React Native Modal.](https://github.com/react-native-community/react-native-modal)  The support for 

This is accomplished by creating a new portal that the modal lives within and setting various `aria` rules to improve accessibility.

## Accessibility Features

This hides app content from screen readers by setting the `aria-modal` flag on the modal.  This is implemented [per the W3C Accessible Design Pattern for Dialogs.](https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html)

This also traps focus within the modal - both when attempting to focus elsewhere using the mouse as well as attempting to focus elsewhere using the keyboard.  This is also implemented [per the W3C Accessible Design Pattern for Dialogs.](https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html)

Furthermore, a built-in <kbd>Escape</kbd> to close mechanism has been implemented which calls `onRequestClose`.  This only applies to the "top-most" dialog when multiple modal dialogs are open.  This matches the react-native behavior for pressing the back button, and again has been implemented  [per the W3C Accessible Design Pattern for Dialogs.](https://www.w3.org/TR/wai-aria-practices/examples/dialog-modal/dialog.html)

### ~~The `animated` property~~

~~The animated property was [deprecated in v0.26 - April 2016](https://github.com/facebook/react-native/commit/2bb1c263dbdce8f92ed5af41a729470238157799#diff-931cea019b5e9faceffef4e46f1fff08).  This property was [removed in v0.63 - February 2020](https://github.com/facebook/react-native/commit/1e9db7bd6df3055b9b81d23f15a54bb250621a41#diff-931cea019b5e9faceffef4e46f1fff08).  It was superseded by `animationType`.~~

### Platform-Specific Properties

The following properties are not implemented because they did not seem relevant for the web:

* `presentationStyle`
* `statusBarTranslucent`
* `hardwareAccelerated`
* `onOrientationChange`
* `supportedOrientations`


KingAmo:
do not work on web

imnotjames:
Could you elaborate?

Naturalclar:
I've been using a patched-version of react-native with this PR, and it's working great :+1:

imnotjames:
Looks like everything has been migrated to use functional components instead of class based components.  I'll update this PR to match :)

imnotjames:
Ran through features in `yarn compile && yarn docs`

- Hides app content from screen-readers when active.
- Traps focus per W3CAG
- Built-in Escape-to-close mechanism.
- Only top-most modal responds to Escape-to-Close
- Animations for fade and slide work as expected

All seem to be working as expected :)

However, still needs the flow types to be updated.

imnotjames:
@necolas you mentioned in another PR that you'd prefer this NOT re-use `Animated`.  Just want to confirm that's what you meant before I refactor the animation component.

RichardLindhout:
@imnotjames looks very good with animationEnd and css animations!

A few things to consider
- Instead of trying to focus every element we could maybe use de data-focusable of react native web or is that too high level?

imnotjames:
> Instead of trying to focus every element we could maybe use de data-focusable of react native web or is that too high level?

That's possible.  Do you have a link to where that's documented?  I'm having trouble finding it..

Would it work as expected to

- the W3C Accessibility Guidelines
- aria browser hints
- interactions with other focus management mechanisms (eg, TV UI)

If so, it might be worth exploring.

I'm always in the camp of "less is more".  With this PR I was mostly trying to cut down on the amount of overridden UX interactions & let the browser handle as much as possible - so we aren't re-implementing what's already been solved by the browsers. :)  Fewer cases for unexpected interactions.

For example, If someone _does_ want to make a TV UI with react native they'd still be able to - this feature does not override MUCH of the focus logic - only keeping the focus within the modal per W3CAG suggestion.

packages/react-native-web/src/exports/Modal/ModalAnimation.js
```diff
@@ -0,0 +1,160 @@
...
+    }
+
+    return animationType;
+  }, [animationType, animated]);
+
+  const isAnimated = computedAnimationType !== 'none';
+
+  // Wrap the callbacks so we don't have to worry about them being null
+  const onShowCallback = useCallback(() => { if (onShow) { onShow(); } }, [onShow]);
+  const onDismissCallback = useCallback(() => { if (onDismiss) { onDismiss(); } }, [onDismiss])
```
necolas, line 74: We don't need these extra `useCallback` wrappers, this can be moved inside the body of `animationEndCallback`

packages/react-native-web/src/exports/Modal/ModalAnimation.js
```diff
@@ -0,0 +1,160 @@
...
+      '100%': { transform: [ { translateY: '0%' } ] }
+    }
+  },
+  slideOut: {
+    transform: [ { translateY: '100%' } ],
+    animationKeyframes: {
+      '0%': { transform: [ { translateY: '0%' } ] },
+      '100%': { transform: [ { translateY: '100%' } ] }
+    }
+  },
```
necolas, line 154: Did you try using `animationDirection:'reverse'` instead of defining the "out" animations?

packages/react-native-web/src/exports/Modal/index.js
```diff
@@ -1,2 +1,251 @@
...
+  const focusRef = useRef<{ trapFocusInProgress: boolean, lastFocusedElement: ?HTMLElement }>({
+    trapFocusInProgress: false,
+    lastFocusedElement: null
+  });
+
+  const onDismiss = useCallback(() => {
+    // When we dismiss we can't assume that we're dismissing the
+    // top element in the stack - so search the stack and remove
+    // ourselves from it if need be.
+    if (visibleModalStack.includes(modalId)) {
```
necolas, line 109: IE11 doesn't support `.includes` so let's use an index check

packages/react-native-web/src/exports/Modal/index.js
```diff
@@ -1,2 +1,251 @@
...
+    // running it.
+    if (focusRef.current.trapFocusInProgress) {
+      return;
+    }
+
+    // If the underlying modal element reference hasn't been set yet
+    // we can't do much with trapping focus.
+    if (!modalElementRef.current) {
+      return;
+    }
```
necolas, line 149: Can we combine these into one if/return block

packages/react-native-web/src/exports/Modal/index.js
```diff
@@ -1,2 +1,251 @@
...
+  }, [modalId, visible, modalElementRef]);
+
+  const closeOnEscape = useCallback((e: KeyboardEvent) => {
+    if (!visible) {
+      return;
+    }
+
+    if (!isTopModal(modalId)) {
+      return;
+    }
```
necolas, line 182: Merge these too

packages/react-native-web/src/exports/Modal/FocusBracket.js
```diff
@@ -0,0 +1,44 @@
...
+
+  useEffect(() => {
+    if (ref.current) {
+      ref.current.setNativeProps({
+        tabIndex: 0
+      });
+    }
+  }, [ref]);
+
+  return <Text ref={ref} style={[styles.focusBracket]} />;
```
necolas, line 35: Let's make this `createElement('div', { accessibilityRole: 'none', focusable: true })`. The visibility should also be set to hidden.

packages/react-native-web/src/exports/Modal/ModalAnimation.js
imnotjames: I did.  The problem I ran into with that was why I've included a `transform` outside of the keyframes.

At the end of the animation but before `OnAnimationEnd` fires there's a brief moment of the animation resetting to the `0%` - even when I was fiddling with the fill modes and other animation options.

With the `transform` added in the `slideOut` as `translateY(100%)` after the animation completes it will maintain the `100%` position.

For reference, I saw this on Firefox.

packages/react-native-web/src/exports/Modal/index.js
imnotjames: Ah! Definitely, will do. Was there a check I should have run to detect this statically?

packages/react-native-web/src/exports/Modal/ModalAnimation.js
```diff
@@ -0,0 +1,160 @@
...
+
+
+const ANIMATION_DURATION = 300;
+
+function getAnimationStyle(animationType, visible) {
+  if (animationType === 'slide') {
+    return [
+      (visible ? styles.animatedIn : styles.animatedOut),
+      (visible ? styles.slideIn : styles.slideOut)
+    ]
```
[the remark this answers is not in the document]
imnotjames: Not sure I understand - aren't they already?  They're created via `StyleSheet.create`  Do you have an example where this is done?

Or do you mean combining animatedIn & slideIn / animaedOut & slideOut, etc?

packages/react-native-web/src/exports/Modal/FocusBracket.js
```diff
@@ -0,0 +1,44 @@
...
+
+/**
+ * This Component is used to "wrap" the modal we're opening
+ * so that changing focus via tab will never leave the document.
+ *
+ * This allows us to properly trap the focus within a modal
+ * even if the modal is at the start or end of a document.
+ */
+
+const FocusBracket = () => {
```
imnotjames, line 24: @necolas is there a spot that would be a good place for internal modules like this?  I feel bad polluting the `export` namespace with things that.. technically shouldn't be exported 'cause they're only for internal use.

packages/react-native-web/src/exports/Modal/ModalAnimation.js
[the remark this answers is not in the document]
RichardLindhout: I think they only work if they are applied dynamically 

> > Instead of trying to focus every element we could maybe use de data-focusable of react native web or is that too high level?
> 
> That's possible. Do you have a link to where that's documented? I'm having trouble finding it..
> 
> Would it work as expected to
> 
> * the W3C Accessibility Guidelines
> * aria browser hints
> * interactions with other focus management mechanisms (eg, TV UI)
> 
> If so, it might be worth exploring.
> 
> I'm always in the camp of "less is more". With this PR I was mostly trying to cut down on the amount of overridden UX interactions & let the browser handle as much as possible - so we aren't re-implementing what's already been solved by the browsers. :) Fewer cases for unexpected interactions.
> 
> For example, If someone _does_ want to make a TV UI with react native they'd still be able to - this feature does not override MUCH of the focus logic - only keeping the focus within the modal per W3CAG suggestion.


React Native Web adds a data-focusable=true on elements which are focusable. But I don't know if this will be forever.
https://github.com/necolas/react-native-web/pull/1698/files#diff-13924d58d5a7ddd791f0e9ae7b588c46R75

RichardLindhout:
> > Instead of trying to focus every element we could maybe use de data-focusable of react native web or is that too high level?
> 
> That's possible. Do you have a link to where that's documented? I'm having trouble finding it..
> 
> Would it work as expected to
> 
> * the W3C Accessibility Guidelines
> * aria browser hints
> * interactions with other focus management mechanisms (eg, TV UI)
> 
> If so, it might be worth exploring.
> 
> I'm always in the camp of "less is more". With this PR I was mostly trying to cut down on the amount of overridden UX interactions & let the browser handle as much as possible - so we aren't re-implementing what's already been solved by the browsers. :) Fewer cases for unexpected interactions.
> 
> For example, If someone _does_ want to make a TV UI with react native they'd still be able to - this feature does not override MUCH of the focus logic - only keeping the focus within the modal per W3CAG suggestion.

React Native Web adds a data-focusable=true on elements which are focusable. But I don't know if this will be forever.
https://github.com/necolas/react-native-web/pull/1698/files#diff-13924d58d5a7ddd791f0e9ae7b588c46R75

imnotjames:
One other Q - should I squash these commits or do you merge in via squash?

packages/react-native-web/src/exports/Modal/ModalAnimation.js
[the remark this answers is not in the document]
necolas: We don't need to create new arrays each time, they can be defined outside the function:

```js
const animationSlideInStyles = [ styles.animatedIn, styles.slideIn ];
const animationSlideOutStyles = [ styles.animatedOut, styles.slideOut ];

function getAnimationStyle(animationType, visible) {
  if (animationType === 'slide') {
    return visible ? animationSlideInStyles : animationSlideOutStyles;
  }
  // ...
}
```

packages/react-native-web/src/exports/Modal/FocusBracket.js
necolas: This is fine

packages/react-native-web/src/exports/Modal/ModalFocusTrap.js
```diff
@@ -0,0 +1,151 @@
...
+ */
+
+const FocusBracket = () => {
+  return createElement(
+    'div',
+    {
+      style: styles.focusBracket,
+      accessibilityRole: 'none',
+      focusable: true,
+      tabIndex: 0
```
necolas, line 34: You don't need this

packages/react-native-web/src/exports/Modal/ModalFocusTrap.js
```diff
@@ -0,0 +1,151 @@
...
+      <FocusBracket />
+    </>
+  );
+};
+
+export default ModalFocusTrap;
+
+const styles = StyleSheet.create({
+  focusBracket: {
+    visibility: 'none',
```
necolas, line 148: value should be `hidden`

packages/react-native-web/src/exports/Modal/ModalAnimation.js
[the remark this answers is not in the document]
imnotjames: Got it!  Makes much more sense that way. :) Thanks!

packages/react-native-web/src/exports/Modal/ModalFocusTrap.js
imnotjames: TLDR: Per testing, If `visibility` is set to `hidden` or `tabIndex` is unset it seems to break the focus trapping.
Without these being set they no longer are focusable after the elements within the modal.

The focus swaps from within the modal to outside the modal when you hit `tab` within the
modal.  Great!  Normally the focus trap catches & we move the focus back to the top.
This is done by listening to focus events on the document / user agent browser pane.

Except.. If the modal is the "last" focusable item on the page.  At that point the focus leaves
the user agent - to the browser chrome instead.  The focus event on the document never fires
for this because we aren't changing focus to another element within the document - we leave the document.

That's what the focus brackets are there to prevent - to always force an element leaving the modal to find another
element to focus on before leaving the document.

Not sure how to write a test to verify this, though.  Perhaps render the document with the focus trap enabled and search up and down from modal that there's at least one focusable element outside of the div?

packages/react-native-web/src/exports/Modal/ModalFocusTrap.js
imnotjames: > TLDR: Per testing, If visibility is set to hidden or tabIndex is unset it seems to break the focus trapping.
Without these being set they no longer are focusable after the elements within the modal.

packages/react-native-web/src/exports/Modal/ModalFocusTrap.js
imnotjames: Happy to remove the `tabIndex` and set `visibility: hidden` so we can verify this in storybook

imnotjames:
I dropped the `animated` property.  Since this PR was originally opened you've bumped the supported react Native compatibility to where it was dropped (0.63).

I also have tests ready for `ModalAnimation`, `ModalPortal`, and `ModalTrapFocus` .

Trying to figure out if there's a way to cleanly split out the rest of the code from the main modal export to simplify testing that logic too...

packages/react-native-web/src/exports/Modal/index.js
```diff
@@ -1,2 +1,147 @@
...
+  // Set a unique model identifier so we can correctly route
+  // dismissals and check the layering of modals.
+  const modalId = useMemo(() => uniqueModalIdentifier++, []);
+
+  const onDismissCallback = useCallback(() => {
+    // When we dismiss we can't assume that we're dismissing the
+    // top element in the stack - so search the stack and remove
+    // ourselves from it if need be.
+    if (visibleModalStack.indexOf(modalId) !== -1) {
+      visibleModalStack.splice(visibleModalStack.indexOf(modalId), 1);
```
imnotjames, line 56: If the modal just stops existing - eg, it's removed from the render tree - this won't fire and we end up with the modal stack in a bad spot.

We should probably use an effect to remove the element from the stack in the case of the modal being destroyed

```
useEffect(() => () => removeModal(modalId), [modalId])
```

necolas:
Let's only test the public API. That will make it easier to refactor code without having to move tests around

packages/react-native-web/src/exports/Modal/ModalFocusTrap.js
necolas: Oh yeah I haven't added focusable to this library yet, it's `accessible`. And the focus bracket should be visually hidden using the `clip` technique and probably hidden from screen readers too. There probably needs to be a bit of hidden ui that is available to screen readers and can be used to close the modal

imnotjames: > Oh yeah I haven't added focusable to this library yet, it's accessible.

I'll play with `accessible` a bit.

>  And the focus bracket should be visually hidden using the clip technique

Isn't it already hidden, though?  It has no content, no height, etc.  We can definitely clip it to hide it further, though, just want to make sure that's not overkill here.

> and probably hidden from screen readers too.

Sure, that makes sense. I'll add `aria-hidden` to the element.

> There probably needs to be a bit of hidden ui that is available to screen readers and can be used to close the modal

We do have the <kbd>Esc</kbd> key to close the dialog which might fulfill that purpose for screen readers.  The only other bit suggested per WAI-ARIA is ["It is strongly recommended that the tab sequence of all dialogs include a visible element with role button that closes the dialog, such as a close icon or cancel button."](https://www.w3.org/TR/wai-aria-practices-1.1/#keyboard-interaction-7) - but that refers to a visible elements which would affect the UI & design.

I don't believe the react-native implementation has much more than this either, but that was just through a cursory look [at the javascript bits](https://github.com/facebook/react-native/blob/master/Libraries/Modal/Modal.js) - but most handheld devices or TVs have a back button that might satisfy this.

imnotjames:
> Let's only test the public API. That will make it easier to refactor code without having to move tests around

I'll try - but the mix of portal / animation / etc - it does get a bit hairy to test.  If we want to refactor the internals we can always just delete the internals tests.

My hesitation with testing it all together is that it binds a lot of the assumptions together - eg, all tests will have to assume that they can't use the `@testing-library/react` `container` because of the portal & all tests will have to take into account that multiple modals at once will affect things in different ways.

packages/react-native-web/src/exports/Modal/ModalFocusTrap.js
imnotjames: To get `aria-hidden` I used `importantForAccessibility='no-hide-descendants'`

Daniel-Griffiths:
Are any other changes required on this PR? I would be happy to contribute. 🙂

packages/react-native-web/src/exports/Modal/__tests__/ModalAnimation.js
```diff
@@ -0,0 +1,173 @@
+/* eslint-env jasmine, jest */
+
+import ModalAnimation from '../ModalAnimation';
+import React from 'react';
+import { cleanup, render } from '@testing-library/react';
+
+describe('components/Modal/ModalAnimation', () => {
```
necolas, line 7: Please combine all the tests into a single test file that is testing the public API of `Modal` rather than individual components that are part of the current implementation

imnotjames: Ok - I had to drop a few of the tests to be able to test against only the public API,  Some of the tests are also no longer testing the functionality as intended (eg, the focus trap tests)

necolas:
Please can you elaborate on:

> Some of the tests are also no longer testing the functionality as intended

imnotjames:
Sure!  The example I gave was the focus trapping tests.

With the way the tests were set up before the focus trapping test was run against only the focus trapping element.

```jsx
<a data-testid={'outside'} href={'#outside'}>Outside</a>
<ModalFocusTrap active={false}>
  <a data-testid={'inside'} href={'#hello'}>Hello</a>
</ModalFocusTrap>
```

The way the test worked was that it would focus outside the trap and would verify that the `ModalFocusTrap` no longer attempted to trap the focus back into itself because `active` was set to `false`.  It'd check that `document.activeElement` was NOT the `inside` anchor.

However, with the way the new tests had to be written -

```
<a data-testid={'outside'} href={'#outside'}>Outside</a>
<Modal>
  <a data-testid={'inside'} href={'#hello'}>Hello</a>
</Modal>
```

We can verify that the focus is not trapped within the `Modal`.  HOWEVER, the difference here is that the `inside` anchor is not even rendered & the `ModalFocusTrap` isn't even in the DOM at all.

It's POSSIBLE that there's a case where the `ModalFocusTrap` _could_ be included in the DOM, that we wouldn't want it trapping & it still is trying to - but this test does not verify that.

sawa-zen:
When used with SSR, it seems to get an error because the document here doesn't exist. canUseDom may need to be used.
https://github.com/necolas/react-native-web/blob/next/packages/react-native-web/src/exports/Modal/ModalPortal.js#L23

imnotjames:
Neat!  If I wrap that with `canUseDom` will that break the modal when SSR is live?

sawa-zen:
I changed the following line to return null if canUseDom is false, and it works fine.
ttps://github.com/necolas/react-native-web/blob/next/packages/react-native-web/src/exports/Modal/index.js#L101

Personally, I don't think modals need to be render in SSR, but only in CSR, so I think the above modification method is fine.

sawa-zen:
By the way, react-native-paper seems to be controlled by rendered state.
https://github.com/callstack/react-native-paper/blob/master/src/components/Modal.tsx#L182

imnotjames:
Seems like the standard react-native modal always has the children of the modal mounted.  I'll be refactoring things a bit to make that match up.

packages/docs/src/components/Modal/Modal.stories.js
```diff
@@ -0,0 +1,45 @@
+import PropTypes from 'prop-types';
```
necolas, line 1: Please remove all the prop types stuff and export the examples from 'examples/index.js' to match the other docs

packages/docs/src/components/Modal/Modal.stories.mdx
```diff
@@ -0,0 +1,95 @@
...
+
+### onShow
+
+The `onShow` callback is called once the modal has been shown and may be visible.
+
+### onDismiss
+
+The `onDismiss` callback is called after the modal has been dismissed and is no longer visible.
+
+### presentationStyle
```
necolas, line 77: Let's remove all the unsupported props from the docs

packages/react-native-web/src/exports/Modal/ModalAnimation.js
```diff
@@ -0,0 +1,155 @@
...
+  if (animationType === 'fade') {
+    return visible ? animatedFadeInStyles : animatedFadeOutStyles;
+  }
+
+  return visible ? styles.container : styles.hidden;
+}
+
+export type ModalAnimationProps = {|
+  children?: any,
+
```
necolas, line 32: Please remove the blank lines here. There are several other places where the formatting is like this. I won't comment on them all but please make the same change elsewhere

packages/react-native-web/src/exports/Modal/ModalContent.js
```diff
@@ -0,0 +1,97 @@
...
+    active,
+    transparent,
+    children,
+    onRequestClose
+  } = props;
+
+  const closeOnEscapeCallback = useCallback((e: KeyboardEvent) => {
+    // If the modal that received this event is not considered the "active" modal we
+    // should ignore this event
+    if (!active) {
```
necolas, line 39: Move this into the condition below: active && e.key...

packages/react-native-web/src/exports/Modal/ModalContent.js
```diff
@@ -0,0 +1,97 @@
...
+    }
+  }, [active, onRequestClose]);
+
+  // Bind to the document itself for this component
+  useEffect(() => {
+    if (canUseDOM) {
+      document.addEventListener('keyup', closeOnEscapeCallback, false);
+    }
+
+    return () => {
```
necolas, line 58: This whole thing can be moved within the canusedom check above

packages/react-native-web/src/exports/Modal/ModalContent.js
```diff
@@ -0,0 +1,97 @@
...
+      if (canUseDOM) {
+        document.removeEventListener('keyup', closeOnEscapeCallback, false);
+      }
+    };
+  }, [closeOnEscapeCallback]);
+
+  const style = useMemo(() => {
```
necolas, line 65: No need for memo here. Move the style inline on the view

packages/react-native-web/src/exports/Modal/ModalFocusTrap.js
```diff
@@ -0,0 +1,175 @@
...
+      }
+    } finally {
+      focusRef.current.trapFocusInProgress = false;
+    }
+
+    focusRef.current.lastFocusedElement = getActiveElement();
+  }, [active]);
+
+  // Bind to the document itself for this component
+  useEffect(() => {
```
necolas, line 142: Move everything in this effect into canUseDOM check and then you can avoid the check in all the other functions. This will also need to use the "isomorphic" useLayoutEffect to attach these handlers before paint, which I'll add when committing this

packages/react-native-web/src/exports/Modal/types.js
```diff
@@ -0,0 +1,36 @@
...
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ */
+
+export type OrientationChangeEvent = {|
+  orientation: 'portrait' | 'landscape'
+|};
+
+export type ModalProps = {|
```
necolas, line 15: Move these types into the entry file please

packages/react-native-web/src/exports/Modal/ModalContent.js
imnotjames: Won't the array cause a re-render 'cause it'd be a new array?

imnotjames:
@sawa-zen can you confirm we're all good with next.js? :)

necolas:
I noticed that in the docs the modals appear on page load and then fade/slide out

imnotjames:
Oh no!  I didn't see that when testing but it's possible.

They should be loading without being visible but it's possible something is up with that.

I'll try both Firefox and chrome to replicate and find the root cause.

packages/react-native-web/src/exports/Modal/ModalAnimation.js
```diff
@@ -0,0 +1,141 @@
...
+  // If the `visible` flag is changing we want to set the rendering flag to true
+  // before the animations ever will start
+  useEffect(() =>  {
+    if (!isAnimated) {
+      // If !isAnimated we have to manually call `animationEndCallback` - if we
+      // don't it will never get called
+      animationEndCallback();
+    }
+  }, [isAnimated, visible, animationEndCallback]);
```
necolas, line 73: `visible` is in the deps by not used by this effect

imnotjames: Right - this is to force it to be called any time the `visible` flag changes.   Perhaps a better comment or a `ref` that it can compare against to prevent it from calling too many times?

sawa-zen:
Modals are now shown, but the following warnings are now shown
```
Warning: Expected server HTML to contain a matching <div> in <div>.
    in div (created by ModalAnimation)
    in ModalAnimation (created by ForwardRef)
    in ModalPortal (created by ForwardRef)
    ...
```
I'll see what I can find out.

packages/react-native-web/src/exports/Modal/ModalAnimation.js
necolas: It's probably a sign that this and the surrounding code should be refactored to maintain correct dependencies

imnotjames: I've added a ref to keep track of the previous visibility to compare it to the current & also ended up adding back another check for the `visible` flag to prevent the flash of fade / slide on an initial render of modals that aren't visible.

Let me know if that works for you.

imnotjames:
> Modals are now shown, but the following warnings are now shown
> 
> ```
> Warning: Expected server HTML to contain a matching <div> in <div>.
>     in div (created by ModalAnimation)
>     in ModalAnimation (created by ForwardRef)
>     in ModalPortal (created by ForwardRef)
>     ...
> ```
> 
> I'll see what I can find out.

Let me know!  Or, if you've got a repository for replicating this I'd also be happy to dig into the root cause.

packages/react-native-web/src/exports/Modal/ModalPortal.js
```diff
@@ -0,0 +1,52 @@
...
+    }
+
+    return () => {
+      if (canUseDOM && element && document.body) {
+        document.body.removeChild(element);
+      }
+    }
+  }, [element]);
+
+  if (!canUseDOM || !element) {
```
sawa-zen, line 43: Depending on the result of useCanDom, there was a difference between the server side and the client side DOM to be rendered. So I solved this problem by creating the element in useEffect and modifying the first rendering result for the server side and the client side in the same way.

```suggestion
  const [element, setElement] = useState()

  useEffect(() => {
    const newElement = document.createElement('div')
    setElement(newElement)
    document.body.appendChild(newElement);
    return () => {
      document.body.removeChild(newElement);
    }
  }, [])

  if (!element) {
```

imnotjames: Perhaps I don't understand how `next.js` works - won't this fail because it can't use the DOM?

imnotjames: Ah, found this - https://github.com/vercel/next.js/tree/canary/examples/with-portals reading through it now

sawa-zen: SSR does not execute useEffect, and only when JS is executed in the browser will it be processed in useEffect. So it won't fail.

imnotjames: Ok - I've implemented per that - I had to add some extra checks, unfortunately.  The flow-typing failed if I didn't guard against some scenarios.

sawa-zen: I can confirm that it works well on SSR! Thanks! 🎉

sawa-zen (approved):

packages/react-native-web/src/exports/Modal/ModalAnimation.js
```diff
@@ -45,7 +45,9 @@ function ModalAnimation(props: ModalAnimationProps) {
   } = props;
 
   const [isRendering, setIsRendering] = useState(false);
-  const wasVisible = useRef(false);
+
+  // Undefined by default so the initial `onDismiss` callback is fired
```
necolas, line 49: This doesn't seem right to me. `onDismiss` shouldn't be getting called on initial render but it currently is

imnotjames: That's.. true.

Looking at the [`react-native` implementation](https://github.com/facebook/react-native/blob/master/Libraries/Modal/Modal.js#L166) `onDismiss` is ONLY called when the modal is being unmounted - not when the modal is being hidden.

Does my interpretation align with yours on this?

packages/docs/src/components/Modal/Modal.stories.mdx
```diff
@@ -0,0 +1,83 @@
...
+
+The Modal component is a basic way to present content above an enclosing view.
+Modals may be nested within other Modals.
+
+## Props
+
+| Name                      | Type           | Default |
+| ------------------------- | -------------- | ------- |
+| children                  | ?any           | |
+| visible                   | ?boolean       | false |
```
imnotjames, line 16: `visible` should be `true` by default

https://github.com/facebook/react-native/blob/master/Libraries/Modal/Modal.js#L149

packages/react-native-web/src/exports/Modal/ModalAnimation.js
imnotjames: My confusion is in that https://reactnative.dev/docs/modal.html#ondismiss doesn't align with the code.

imnotjames: Through testing on android I'm finding the behavior is this as well.  Perhaps the react docs should be updated.

necolas:
A couple of other things I noticed from the examples

1. Opening "modalception" doesn't seem to move the focus onto the buttons of the 2nd/3rd/etc modals.
2. You can't shift-tab backwards past the first tab-stop, but you can tab forwards past the last tab-stop

imnotjames:
> Opening "modalception" doesn't seem to move the focus onto the buttons of the 2nd/3rd/etc modals.

The focus trapping currently only prevents focus from being made against items not in the modal.  It does allow for there to be nothing focused.  Is that what you're seeing?  Or do you mean the focus stays on the other modals?

> You can't shift-tab backwards past the first tab-stop, but you can tab forwards past the last tab-stop

This does seem to be a bug that was introduced with changes for these edge cases.  I'll create a test to validate this behavior & fix it.

packages/react-native-web/src/exports/Modal/ModalContent.js
```diff
@@ -0,0 +1,86 @@
...
+import { canUseDOM } from 'fbjs/lib/ExecutionEnvironment';
+
+import View from '../View';
+import StyleSheet from '../StyleSheet';
+
+export type ModalContentProps = {|
+  children?: any,
+  active?: ?(boolean | () => boolean),
+  transparent?: ?boolean,
+  onRequestClose?: ?() => void,
```
necolas, line 22: Alphabetical

packages/react-native-web/src/exports/Modal/ModalContent.js
```diff
@@ -0,0 +1,86 @@
...
+ *
+ * @flow
+ */
+
+import React, { forwardRef, useCallback, useMemo, useEffect } from 'react';
+
+import { canUseDOM } from 'fbjs/lib/ExecutionEnvironment';
+
+import View from '../View';
+import StyleSheet from '../StyleSheet';
```
necolas, line 16: Remove whitespace between imports

packages/react-native-web/src/exports/Modal/ModalContent.js
```diff
@@ -0,0 +1,86 @@
...
+  onRequestClose?: ?() => void,
+|};
+
+const ModalContent = forwardRef<ModalContentProps, *>((props, forwardedRef) => {
+  const {
+    active,
+    transparent,
+    children,
+    onRequestClose
+  } = props;
```
necolas, line 31: Alphabetical

packages/react-native-web/src/exports/Modal/ModalContent.js
```diff
@@ -0,0 +1,86 @@
...
+  const closeOnEscapeCallback = useCallback((e: KeyboardEvent) => {
+    // If the modal that received this event is not considered the "active" modal we
+    // should ignore this event
+    if (active && e.key === 'Escape') {
+      e.stopPropagation();
+
+      if (onRequestClose) {
+        onRequestClose();
+      }
+    }
```
necolas, line 42: Remove comment and blank line

packages/react-native-web/src/exports/Modal/ModalContent.js
```diff
@@ -0,0 +1,86 @@
...
+  }, [active, onRequestClose]);
+
+  // Bind to the document itself for this component
+  useEffect(() => {
+    if (canUseDOM) {
+      document.addEventListener('keyup', closeOnEscapeCallback, false);
+
```
necolas, line 49: Remove blank line

packages/react-native-web/src/exports/Modal/ModalContent.js
```diff
@@ -0,0 +1,86 @@
...
+    if (active && e.key === 'Escape') {
+      e.stopPropagation();
+
+      if (onRequestClose) {
+        onRequestClose();
+      }
+    }
+  }, [active, onRequestClose]);
+
+  // Bind to the document itself for this component
```
necolas, line 45: Remove comment

packages/react-native-web/src/exports/Modal/ModalContent.js
```diff
@@ -0,0 +1,86 @@
...
+
+const ModalContent = forwardRef<ModalContentProps, *>((props, forwardedRef) => {
+  const {
+    active,
+    transparent,
+    children,
+    onRequestClose
+  } = props;
+
+  const closeOnEscapeCallback = useCallback((e: KeyboardEvent) => {
```
necolas, line 33: Don't need a separate useCallback. Fold this into the useEffect below

packages/react-native-web/src/exports/Modal/ModalFocusTrap.js
```diff
@@ -0,0 +1,165 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ */
+
+import React, { useRef, useCallback, useEffect } from 'react';
+
+import { canUseDOM } from 'fbjs/lib/ExecutionEnvironment';
+
```
necolas, line 14: Remove blank lines

packages/react-native-web/src/exports/Modal/ModalFocusTrap.js
```diff
@@ -0,0 +1,165 @@
...
+const FocusBracket = () => {
+  return createElement(
+    'div',
+    {
+      style: styles.focusBracket,
+
+      // Sets `aria-hidden` to true & remove the semantic meaning
+      accessible: true,
+      importantForAccessibility: 'no-hide-descendants',
```
necolas, line 35: There doesn't seem to be much point using these props if you're setting tabIndex below. So just add aria-hidden and remove these. Also remove the blank lines please

packages/react-native-web/src/exports/Modal/ModalFocusTrap.js
```diff
@@ -0,0 +1,165 @@
...
+    if (attemptFocus(child) || focusLastDescendant(child)) {
+      return true;
+    }
+  }
+  return false;
+}
+
+export type ModalFocusTrapProps = {|
+  children?: any,
+  active?: boolean | () => boolean,
```
necolas, line 82: Alphabetical

packages/react-native-web/src/exports/Modal/ModalFocusTrap.js
```diff
@@ -0,0 +1,165 @@
...
+    lastFocusedElement: null
+  });
+
+  const trapFocus = useCallback(() => {
+    // We should not trap focus if:
+    // - The modal hasn't fully initialized with an HTMLElement ref
+    // - Focus is already in the process of being trapped (eg, we're refocusing)
+    // - isTrapActive prop being false-ish tells us to do nothing
+    // - we cannot use the DOM
+    if (!trapElementRef.current || focusRef.current.trapFocusInProgress || !active || !canUseDOM) {
```
necolas, line 101: Shouldnt need canUseDOM here, because this function is only called from within a canUseDOM condition.

packages/react-native-web/src/exports/Modal/ModalFocusTrap.js
```diff
@@ -0,0 +1,165 @@
...
+  const trapElementRef = useRef<?Node>();
+
+  // Ref used to track trapping of focus and to prevent focus from leaving a modal
+  // for accessibility reasons per W3CAG.
+  const focusRef = useRef<{ trapFocusInProgress: boolean, lastFocusedElement: ?HTMLElement }>({
+    trapFocusInProgress: false,
+    lastFocusedElement: null
+  });
+
+  const trapFocus = useCallback(() => {
```
necolas, line 95: Since this is only used within the useEffect below, I don't think it needs to be memoized in a useCallback and can be moved into the effect fn body

packages/react-native-web/src/exports/Modal/ModalPortal.js
```diff
@@ -0,0 +1,57 @@
...
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ */
+
+import { useEffect, useState, useRef } from 'react';
+
+import { canUseDOM } from 'fbjs/lib/ExecutionEnvironment';
+
```
necolas, line 14: Remove blank lines

packages/react-native-web/src/exports/Modal/ModalPortal.js
```diff
@@ -0,0 +1,57 @@
...
+      if (element && document.body) {
+        document.body.appendChild(element);
+        elementRef.current = element;
+        setMounted(true);
+      }
+
+      return () => {
+        // Check for document.body existence 'cause if we don't
+        // the flow typing checks get unhappy.
```
necolas, line 39: Remove comment

packages/react-native-web/src/exports/Modal/ModalPortal.js
```diff
@@ -0,0 +1,57 @@
...
+  }, []);
+
+  // For Next.JS compatibility, we keep track of the DOM actually being
+  // mounted & if it's not we don't render anything - per the next.js docs
+  // on thinking with portals.
+  if (!mounted || !elementRef.current || !canUseDOM) {
+    return null;
+  }
+
+  return ReactDOM.createPortal(children, elementRef.current);
```
necolas, line 54: Comment: "Only render if the portal is mounted"

Combine this into a ternary with the truthy condition first: i.e., conditions ? createPortal: null

packages/react-native-web/src/exports/Modal/index.js
```diff
@@ -1,2 +1,151 @@
...
+ * Copyright (c) Facebook, Inc. and its affiliates.
+ *
+ * This source code is licensed under the MIT license found in the
+ * LICENSE file in the root directory of this source tree.
+ *
+ * @flow
+ */
+
+import React, { forwardRef, useCallback, useMemo, useEffect, useState } from 'react';
+
```
necolas, line 12: Remove blank line

packages/react-native-web/src/exports/Modal/index.js
```diff
@@ -1,2 +1,151 @@
...
+import ModalPortal from './ModalPortal';
+import ModalAnimation from './ModalAnimation';
+import ModalContent from './ModalContent';
+import ModalFocusTrap from './ModalFocusTrap';
+
+type OrientationChangeEvent = {|
+  orientation: 'portrait' | 'landscape'
+|};
+
+export type AnimationType = 'none' | 'slide' | 'fade';
```
necolas, line 22: Inline these types

packages/react-native-web/src/exports/Modal/index.js
```diff
@@ -1,2 +1,151 @@
...
+  transparent?: ?boolean,
+  onOrientationChange?: ?(e: OrientationChangeEvent) => void,
+  supportedOrientations?: ?Array<
+    'portrait' | 'portrait-upside-down' | 'landscape' | 'landscape-left' | 'landscape-right'
+    >,
+  statusBarTranslucent?: ?boolean,
+  hardwareAccelerated?: ?boolean,
+  onRequestClose?: ?() => void,
+  onShow?: ?() => void,
+  onDismiss?: ?() => mixed
```
necolas, line 38: Alphabetical

packages/react-native-web/src/exports/Modal/index.js
```diff
@@ -1,2 +1,151 @@
...
+  }
+
+  const activeModalId = activeModalStack[activeModalStack.length - 1];
+
+  for (const modalId of activeModalStack) {
+    if (modalId in activeModalListeners) {
+      activeModalListeners[modalId](modalId === activeModalId);
+    }
+  }
+}
```
necolas, line 58: Remove blank lines

packages/react-native-web/src/exports/Modal/index.js
```diff
@@ -1,2 +1,151 @@
...
+    return;
+  }
+
+  const activeModalId = activeModalStack[activeModalStack.length - 1];
+
+  for (const modalId of activeModalStack) {
+    if (modalId in activeModalListeners) {
+      activeModalListeners[modalId](modalId === activeModalId);
+    }
+  }
```
necolas, line 57: Please can we iterate over the stack using forEach rather than `for..of`

packages/react-native-web/src/exports/Modal/index.js
```diff
@@ -1,2 +1,151 @@
...
+  if (modalId in activeModalListeners) {
+    // Before removing this listener we should probably tell it
+    // that it's no longer the active modal for sure.
+    activeModalListeners[modalId](false);
+    delete activeModalListeners[modalId];
+  }
+
+  const index = activeModalStack.indexOf(modalId);
+
+  if (index !== -1) {
```
necolas, line 70: remove blank lines

packages/react-native-web/src/exports/Modal/index.js
```diff
@@ -1,2 +1,151 @@
...
+    notifyActiveModalListeners();
+  }
+}
+
+function addActiveModal(modalId, listener) {
+  removeActiveModal(modalId);
+
+  activeModalStack.push(modalId);
+  activeModalListeners[modalId] = listener;
+
```
necolas, line 81: blank lines

packages/react-native-web/src/exports/Modal/index.js
```diff
@@ -1,2 +1,151 @@
...
+const Modal = forwardRef<ModalProps, *>((props, forwardedRef) => {
+  const {
+    visible = true,
+    animationType,
+    transparent,
+    children,
+    onShow,
+    onDismiss,
+    onRequestClose
```
necolas, line 93: alphabetical

packages/react-native-web/src/exports/Modal/index.js
```diff
@@ -1,2 +1,151 @@
...
+    }
+  }, [modalId, onDismiss]);
+
+  const onShowCallback = useCallback(() => {
+    addActiveModal(modalId, setIsActive);
+
+    if (onShow) {
+      onShow();
+    }
+  }, [modalId, onShow]);
```
necolas, line 119: blank lines

necolas (commented):
Nearly there! Thanks

packages/react-native-web/src/exports/Modal/ModalAnimation.js
imnotjames: Ok - I've also looked at the last known working code for the dismissal in 0.62 to confirm the behavior matches your suggestion (it does)

while the code that [wraps firing off the onDismiss looks somewhat similar to `onShow`](https://github.com/facebook/react-native/blob/d0871d0a9a373e1d3ac35da46c85c0d0e793116d/React/Views/RCTModalHostView.m#L195)

It's [very different from how it's called in the manager](https://github.com/facebook/react-native/blob/v0.61.5/React/Views/RCTModalHostViewManager.m#L82)

So while `onShow` fires if the modal starts visible, `onDismiss` won't if it starts not-visible.

imnotjames:
Could be useful to look into https://www.npmjs.com/package/eslint-plugin-sort-destructure-keys for enforcing the sorted destructuring

packages/react-native-web/src/exports/Modal/__tests__/index.js
```diff
@@ -0,0 +1,447 @@
...
+          <div data-testid={'inside'}>There are no focusable contents.</div>
+        </Modal>
+      </>
+    );
+
+    const outsideElement = document.querySelector('[data-testid="outside"]');;
+
+    outsideElement.focus();
+
+    expect(document.activeElement).toBe(document.body);
```
necolas, line 403: Why is the body the expected active element if the test is expecting focus to be trapped within the modal?

imnotjames: Because the modal has nothing within it to focus, so the focus ends up being to the entire document body - preventing an outside element from being focused.

necolas: I think focus should probably go to the modal. It doesn't really make sense to throw focus to the body and will be confusing for users. The modal can be made programmatically focusable via tabindex -1

imnotjames: Sure!

Just to clarify - focus on the `document.body` just mean that nothing is currently focused - we haven't explicitly focused the body.  It's if you take the currently active element and `blur` it. It's really up to the browser how this occurs, but we can override this behavior.

This edge-case [is not particularly WCAG WAI ARIA Modal Best Practices & Guidelines document](https://www.w3.org/TR/wai-aria-practices-1.1/#dialog_modal) so I think that it's a reasonable path to make the modal technically focusable and focus that if there are no other focusable elements within the trap.

necolas:
James, thank you for all the work you did on this PR and sticking with it! I really appreciate this contribution. I'm going to close this PR as it has been merged into the `next` branch along with some minor adjustments I made to the code and examples. If we encounter any need for more fixes, those can be made as PRs against `next` and I'll probably squash/rebase them into the single commit that will land this feature.

imnotjames:
Thanks for workin' with me on it & reviewing all the code!

PR #1657, opened 2020-06-29
[fix] setNativeProps and pointerEvent prop

### LINKED ISSUE #1655: pointerEvents and Animated conflict each other
**The problem**

It is not possible to use `pointerEvents` prop of `Animated.View`.

**How to reproduce**

Simplified test case: https://codesandbox.io/s/icy-waterfall-tfnc2?file=/src/App.js

Steps to reproduce:
1. Try to press the button

**Expected behavior**

The button should be pressable because the `pointerEvents` of the overlay above are set to `none`.

* React Native for Web (version): 0.13.0 (it worked with 0.12.3)
* React (version): 16.3.1
* Browser: Chrome 83

necolas:
The pointerEvent prop is converted into a DOM style property and needs to be
accounted for by setNativeProps.

Close #1656
Fix #1655

### DIFF
--- a/packages/react-native-web/src/exports/Picker/index.js
+++ b/packages/react-native-web/src/exports/Picker/index.js
@@ -54,8 +54,6 @@ const Picker = forwardRef<PickerProps, *>((props, forwardedRef) => {
     }
   });
 
-  usePlatformMethods(hostRef, [], style);
-
   function handleChange(e: Object) {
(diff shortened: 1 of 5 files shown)


necolas:
@jaulz try this PR instead

jaulz:
@necolas perfect, thanks!

Issue #1319, opened 2019-04-02
[iOS Safari] Images with tintColor do not render predictably


EvanBacon:
**The problem**

It appears that the images rendered by RNWeb in mobile safari do not show up. 
Additionally the tint color will sometimes not render (if you cause some re-render)

## Expected: 

<img width="1177" alt="Screen Shot 2019-04-02 at 11 19 11 PM" src="https://user-images.githubusercontent.com/9664363/55436889-c5e73080-559d-11e9-8f7d-78850420f34b.png">

## iOS mobile safari:

![IMG_58875F9559E4-1](https://user-images.githubusercontent.com/9664363/55436990-0b0b6280-559e-11e9-8305-f4d0684a21b2.jpeg)

necolas:
> It appears that the images rendered by RNWeb in mobile safari do not show up.

I suspect I would have heard about that from Twitter :) Do you have an example of that independent from the tint color?

> Additionally the tint color will sometimes not render (if you cause some re-render)

I won't have time to look into this myself but can review a PR

williamliangwl:
It looks like the changes of the `tintColor` is not reflected in Safari

I forked the above's repro and add 2 seconds interval of changing the color of the `tintColor`
in the firefox and chrome, they are working fine, but not Safari

Even if I inspect the `feflood` html attribute, its `flood-color` value is not changing

https://codesandbox.io/s/rnweb-tint-color-tdcv7

zamponotiropita:
Similar issue with safari. I'm getting proper `tintColor` on first load, but on page refresh the `tintColor` isn't coloring anymore

markymc:
I'm seeing this on React Navigation too. My back button's tint color is white, but sometimes it goes black. After some debugging it seems to be the same as this problem, somewhere around here:

https://github.com/react-navigation/react-navigation/blob/1c4bd6813bc6f5151fd2d99a0245331ff5631c38/packages/stack/src/views/Header/HeaderBackButton.tsx#L77-L85

I made a hacky fix by changing the colour of the PNG file and replacing it in postinstall, but of course this isn't ideal.

PR #1664, opened 2020-07-07
fix: Unpredictable mobile tintColor

cbstauss:
This fixes a bug with SVG tintColor where the color would not update properly on mobile browsers. Fix for #1319


necolas:
Thanks! Available in 0.13.3

jottenlips:
Still experiencing this on `"react-native-web": "0.14.9"`

Issue #1670, opened 2020-07-13
TextInput component: persistent black underline


JakubKubista:
**The problem**
When you start typing to the TextInput component, then every letter has black underline as a default style on Smart TV (Samsung). This style is not possible to turn off for such a device at the moment. Styles like [textdecorationline](https://reactnative.dev/docs/text-style-props#textdecorationline) are not working, but for this case has been released [underlineColorAndroid](https://reactnative.dev/docs/textinput#underlinecolorandroid) prop, which is currently missing at React Native Web. There is also a small chance, that it's related to known [issue 1023](https://github.com/necolas/react-native-web/issues/1023).

**How to reproduce**
1. Include the `TextInput` component at Smart TV app on Android or Samsung Tizen.
2. Focus on the `TextInput` component.
3. Start writing some text.

**Result**
_Black underline_ appears below the text without any option to disable it.

**Expected behavior**
There should be an option to change text-decoration for these devices.

* Environment (include versions): Samsung Smart TV19 - UE43RU7445UXXC
* Did this work in previous versions? No
* React Native for Web (version): 0.11.7
* React (version): 16.8.5
* Browser: TBC

necolas:
Is this the issue / fix? https://www.quora.com/How-do-I-remove-the-underline-feature-while-typing-in-Android-keyboard

Obviously a prop added to RN for the native Android platform isn't going to do anything on web though

JakubKubista:
@necolas Thank you very much for a quick response. No, this underline appears every time, when a new character appears and the world is not finished even if the world makes sense or not, so it's not related to spelling correction.

necolas:
Try turning off autocorrect. If not, let me know when you know what the cause is.

JakubKubista:
There is not even an autocorrect option on this type of Smart TVs. Since it works with RN then RNW update would be the option on how to make this works. We will try to investigate further about details.

necolas:
Autocorrect is a web attribute

JakubKubista:
Disabling native mobile features such as `autocomplete="off"`, `autocorrect="off"`, `autocapitalize="off"` or `spellcheck="false"` won't help.

PR #1671, opened 2020-07-13
Add TextInput underlineColorAndroid prop

JakubKubista:
Close: https://github.com/necolas/react-native-web/issues/1670


necolas:
Not supported on web

PR #1675, opened 2020-07-16
Add support for dataSet for ScrollView

RyogaK:
This supports to set ::-webkit-scrollbar-thumb, etc., for a specific scrollview.


r-bman:
Will this also work for `FlatList` given that it inherits `ScrollView` props?

RyogaK:
@r-bman This my PR is only focusing on ScrollView. I can't answer your question because it depending on whether the FlatList relays the dataSet prop to the internal ScrollView.

Issue #1668, opened 2020-07-11
Typescript types do no support "dataSet" attribute in 0.13.3


danawoodman:
**The problem**
Based on the latest release, the recommended approach to pass `data-` attributes to a view is `dataSet` but it appears the types have not been updated to reflect this.

Based on the [release notes](https://github.com/necolas/react-native-web/releases/tag/0.13.0) I should be able to do this:

> Forwarding of data-* props is no longer supported. Use dataSet instead. For example, dataSet={{ someName: 1 }}.

However, when passing `dataSet` I get a TS error (see below)

**How to reproduce**
Create a Typescript React component with `dataSet`:

```tsx
<View dataSet={{ foo: 'bar' }} />
```

View TS errors.

**Expected behavior**
I don't get TS errors for a supported attribute.

**Environment (include versions). Did this work in previous versions?**

* React Native for Web (version): 0.13.3
* React (version): 16.13.1
* Browser: Brave (Chromium)

Typescript error:

```
(JSX attribute) dataSet: {
    foo: string;
}
No overload matches this call.
  Overload 1 of 2, '(props: Readonly<ViewProps>): View', gave the following error.
    Type '{ dataSet: { foo: string; }; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<View> & Readonly<ViewProps> & Readonly<{ children?: ReactNode; }>'.
      Property 'dataSet' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<View> & Readonly<ViewProps> & Readonly<{ children?: ReactNode; }>'.
  Overload 2 of 2, '(props: ViewProps, context?: any): View', gave the following error.
    Type '{ dataSet: { foo: string; }; }' is not assignable to type 'IntrinsicAttributes & IntrinsicClassAttributes<View> & Readonly<ViewProps> & Readonly<{ children?: ReactNode; }>'.
      Property 'dataSet' does not exist on type 'IntrinsicAttributes & IntrinsicClassAttributes<View> & Readonly<ViewProps> & Readonly<{ children?: ReactNode; }>'.ts(2769)
```

necolas:
Someone needs to update the TS types wherever they are maintained

danawoodman:
Sorry, I assumed they were maintained here. 

Do you know who is the maintainer?

necolas:
I don't know who the maintainer is. I started this discussion about this if you would like to share your experience https://github.com/necolas/react-native-web/issues/1684

[...]

PR #1682, opened 2020-07-23
Missing accessibilityRole="option"

jonstuebe:
I've been building a combobox and found that `react-native-web` is missing the `accessibilityRole="option"` necessary for proper voice over. I might be missing some other changes below as I'm not super familiar with the codebase.

Here's a link to the spec examples detailing the role usage:

https://www.w3.org/TR/wai-aria-practices/examples/combobox/aria1.1pattern/listbox-combo.html

### DIFF
--- a/packages/react-native-web/src/modules/AccessibilityUtil/propsToAriaRole.js
+++ b/packages/react-native-web/src/modules/AccessibilityUtil/propsToAriaRole.js
@@ -19,7 +19,8 @@ const accessibilityRoleToWebRole = {
   none: 'presentation',
   search: 'search',
   summary: 'region',
-  text: null
+  option: 'option',
+  text: null,
 };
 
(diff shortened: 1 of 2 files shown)


necolas:
`<View accessiblityRole="option" />` already produces `<div role="option" />`

jonstuebe:
Is there a fall through in the code somewhere @necolas? I'm using TS so it sounds like a type issue then.

necolas:
Ah, yes it might be typescript. There was another types issue that was brought up recently https://github.com/necolas/react-native-web/issues/1668.

As time goes I'm starting to wonder more about types. Maybe I'll make another issue to solicit input about types.

necolas:
If you have any thoughts: https://github.com/necolas/react-native-web/issues/1684

PR #1686, opened 2020-07-27
Create RN components that relies on different dom elements

MatheusrdSantos:
This PR contains the implementation of createRNElement.
Now the developers are able to create ReactNative elements like Text that relies on ```<span>```, ```<a>```, ```<h1>```, ```<h2>``` or a Button that relies on ```<button>```. So whatever is the needing the developer will be able to adapt the web behavior of the RN elements.
There are a lot of possible combinations.

### DIFF
--- a/packages/babel-plugin-react-native-web/src/moduleMap.js
+++ b/packages/babel-plugin-react-native-web/src/moduleMap.js
@@ -61,6 +61,7 @@ module.exports = {
   VirtualizedList: true,
   YellowBox: true,
   createElement: true,
+  createRNElement: true,
   findNodeHandle: true,
   processColor: true,
   render: true,
--- a/packages/react-native-web/src/exports/createRNElement/__tests__/__snapshots__/index-test.js.snap
+++ b/packages/react-native-web/src/exports/createRNElement/__tests__/__snapshots__/index-test.js.snap
@@ -1,0 +1,29 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`exports/createRNElement prop "href" 1`] = `
+<a
+  class="css-reset-4rbku5 9"
+  data-focusable="true"
+  href="http://mylink.com"
+/>
+`;
+
+exports[`exports/createRNElement renders default style 1`] = `
+<div
+  class="8"
+  style="height: 10px; width: 10px;"
+/>
+`;
+
+exports[`exports/createRNElement renders different DOM elements 1`] = `
+<h3
+  class="6"
+/>
+`;
+
+exports[`exports/createRNElement renders different DOM elements 2`] = `
+<button
+  class="css-reset-4rbku5 7"
+  data-focusable="true"
+/>
+`;
--- a/packages/react-native-web/src/exports/createRNElement/__tests__/index-test.js
+++ b/packages/react-native-web/src/exports/createRNElement/__tests__/index-test.js
@@ -40,12 +40,4 @@ describe('exports/createRNElement', () => {
     fireEvent.click(container.firstChild);
     expect(myFn).toHaveBeenCalledTimes(1);
   });
-
-  test('onLayout callback', () => {
-    const myFn = jest.fn();
-    const View = createRNElement('div');
-    const { rerender } = render(<View onLayout={myFn} style={{ height: 10, width: 10 }} />);
-    rerender(<View onLayout={myFn} style={{ height: 15, width: 15 }} />);
-    expect(myFn).toHaveBeenCalledTimes(2);
-  });
 });
--- a/packages/react-native-web/src/exports/createRNElement/__tests__/index-test.js
+++ b/packages/react-native-web/src/exports/createRNElement/__tests__/index-test.js
@@ -1,0 +1,51 @@
+/* eslint-env jasmine, jest */
+
+import createRNElement from '../index';
+import React from 'react';
+import { render, fireEvent } from '@testing-library/react';
+
+describe('exports/createRNElement', () => {
+  test('renders different DOM elements', () => {
+    const Text = createRNElement('h3', { ref: true });
+    const { container: h3Container } = render(<Text>test</Text>);
+    expect(h3Container.firstChild).toMatchSnapshot();
+    const Button = createRNElement('button', { ref: true });
+    const { container: btnContainer } = render(<Button />);
+    expect(btnContainer.firstChild).toMatchSnapshot();
+  });
+
+  test('renders default style', () => {
+    const View = createRNElement('div', {}, { backgoundColor: 'red' });
+    const { container } = render(<View style={{ height: 10, width: 10 }} />);
+    expect(container.firstChild).toMatchSnapshot();
+  });
+
+  test('prop "href"', () => {
+    const propList = {
+      ref: true,
+      href: true
+    };
+    const Text = createRNElement('a', propList);
+    const { container } = render(<Text href={'http://mylink.com'}>test</Text>);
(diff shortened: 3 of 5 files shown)


Sharcoux:
@necolas  I'm sorry to insist on this one, but can we discuss it a bit?

Can we first agree on these points:

1. react-native-web is especially interesting for cross-platform projects using react-native and also targeting web

necolas:
This has already been discussed in issues where I said I wouldn't add this API.

> those tags cannot be used with the current implementation of react-native-web

Yes they can. Use `unstable_createElement`

>  there is no easy way to integrate svg in a cross platform fashion

Yes there is, use `unstable_createElement`.

Sharcoux:
What do you want me to do with unstable_createElement? It doesn't accept react native style, it doesn't accept onLayout, it doesn't integrate at all in the react native ecosystem, it just requires me to create a completely different component and everywhere I use it I have to handle it differently than the others. How can you consider this a good solution?

Sharcoux:
Or maybe you can explain me better how to use unstable_createElement in that case?

```javascript
const SVG = ({ svg, onLayout, width, height, style, ...props }) => {
  // This is the easy part
  if (RN.Platform.OS !== 'web') return <SvgXml xml={svg} style={[style, { width, height }]} {...props} />

  // Here starts the trouble
  // first, I need to transform all the attributes that might be set to the svg tag to convert them into an object.
  const attrs = svg.match(/<svg(\s*)?(.*?)>/i)[2]
  const attrsMap = [...attrs.matchAll(/([a-z][-.:0-9_a-z]*)(\s*=\s*("([^"]*)"|'([^']*)'|(\S+)))?/ig)].reduce((object, match) => object[match[1]]=match[4] || match[5] || match[6] || "", {})
  // I need to transform the kebab-case attributes into camelCase
  const camelAttrsMapObject.keys(attrsMap).reduce((object, key) => object[key.replace(/-./g, x=>x.toUpperCase()[1]))] = attrsMap[key])
  // now, I need to create the style object that will behave as the other react native components:
  const rnStyle = {
    alignItems: 'stretch',
    boxSizing: 'border-box',
    display: 'flex',
    flexBasis: 'auto',
    flexDirection: 'column',
    flexShrink: 0,
    position: 'relative';
    z-index: 0;
    // ect, ect, you know the list better than I do
  }

  // Now, I can add the style I received for my component:
  const finalStyle = [ rnStyle, style, { width, height } ] // <- I should actually parse attrsMap.style, but let's keep it 'simple'

  // Finally I can return the component:
  return require('react-native-web').unstable_createElement('svg', {
    ...camelAttrsMapObject,
    style: finalStyle
  })
})
```

Well, I don't consider that it is easy, but ok. Oh but wait, that's not over. Now if I want onPress to work on this, I need to reimplement your useResponderEvent hook... Ah, and if I want to use onLayout, I have to reimplement your useElementLayout hook...

Do you see now why I think that all this is a bit complicated for users?

So, why can't we make all this a bit easier to everyone and provide a feature that does all this in much easier way:

`createRNElement('svg', props)`

Please note that all this is also valid for section, header, nav, article, and all those tags.

necolas:
> What do you want me to do with unstable_createElement?

Read the docs. Unlike your claims, you can get it to render any element you want and it supports RN styles. https://necolas.github.io/react-native-web/docs/?path=/docs/guides-unstable-uses--page

> Do you see now why I think that all this is a bit complicated for users?

Yes, but it's not designed for that use. The react-svg library API isn't going to drive the APIs exported here. You can wrap your SVG in a View and get those layout, etc., props. I'm not going to export implementation details for one use case that can be achieved via other means.

In the future, we might break down the RN APIs so that features like layout and the responder system are available as hooks. But RN isn't there yet.

Sharcoux:
> Yes, but it's not designed for that use. The react-svg library API isn't going to drive the APIs exported here.

svg is not the only use case, and it is not linked with the react-native-svg lib but just the way to handle any html tag within react-native-web in a more convenient way because, as I've just shown you, right now it is very limited and unfriendly. SVG is indeed an excellent use case because all devs will struggle with this which is really a stupid subject to lose time on, but there are also SEO issues, as I told you, that would benefit from using proper html5 tags.

> You can wrap your SVG in a View and get those layout, etc., props

This approach brings just a lot more issues as I have to split the styles and the props between those that need to be applied to the wrapper, and those that needs to be applied to the svg tag.

>  I'm not going to export implementation details for one use case that can be achieved via other means.

If you see an easy way to achieve what I need could you provide a working demo?

@necolas , I have a lot of respect for you for all the incredible work you did with react-native-web. I'm just trying to understand here. You say you don't want this but can you detail the reason why you don't want? Does it have any bad impact?

Sharcoux:
@necolas , please. Give me a reason.
In the current state of things, useElementLayout and other hooks are not being exported, and as you don't provide the createRNElement feature, some very basic stuff becomes absurdly complicated. I don't see any reason why you would prefer limit the usages we can make of react-native-web for creating web content.

paularmstrong:
`unstable_createElement` already does what you're asking for. You're just going a hugely roundabout and unnecessary way. Here's an example SVG that I have: https://github.com/paularmstrong/build-tracker/blob/master/src/app/src/icons/ArrowLeft.tsx You don't need to set innerHtml or any of the extra stuff.

Sharcoux:
@paularmstrong If I have an svg string like this:

```javascript
const svg = `<svg width="10" height="10" stroke="currentColor" viewBox="0 0 10 10" stroke-linecap="round" stroke-width="2" xmlns="http://www.w3.org/2000/svg" xmlns:svg="http://www.w3.org/2000/svg">
  <path d="M1,5 L9,5"/>
  <path d="M5,1 L5,9"/>
</svg>`
```

Can you show me how you represent this as a React-Native object?

Can you show me also how you can use onLayout, refs, and measure on it ?

Also, can you show me how you make it accept and respect flex styles (flexGrow, alignSelf, etc)

Thanks a lot.

Sharcoux:
Also, I would be very happy to know how you can make this React-Native code:

```javascript
<Section>
  <Header>
    <Title>Title</Title>
  </Header>
  <Article>
    <Text>content</Text>
  </Article>
</section>
```

to render as:

```xml
<section>
  <header>
    <h1>Title</h1>
  </header>
  <article>
    <p>Content</p>
  </article>
</section>
```

And each of these elements still accept onLayout and other react-native behavior.

natew:
Just want to voice my strong support here. Just exporting the hooks would be enough to let [my component kit](https://github.com/snackui/snackui) have two big upgrades:

- allow it to use className internally so it can keep atomic styles nicely.
- avoid double the tree depth in many areas when all it needs is a single hook to get the event responders.

Both are pretty critical to it having and real shot at being useful for the community. And I’d be happy to help workshop maybe some midway point we can meet at to make this workable for library authors until RN exports the hooks as well. Or perhaps i can make a PR adding the hooks export to RN itself so we have the excuse to support theme here?

I know of another person not on this thread, a library author, that also is very keen to have this work, so that makes 4 people strongly advocating it, all library authors in the community making helpful things with a few thousand stars between them!

Sharcoux:
@necolas Ok, I wrapped the svg tag within a View to be able to use unstable_createElement and still have access to onLayout, measure, panResponder, etc. But now, my issue is that I'm trying to represent an svg inside a `p` tag. The View is represented as a `div`. But div tag inside p is illegal. So if I want to do the things right, I'll need to use `unstable_createElement`. But by doing so, I'm getting back to the original problem that using this View was supposed to solve: I can't use the hooks...

I really don't understand the benefits of blocking this PR and all the problems it solves for no real reason... Unless you can provide me another workaround?

PR #1689, opened 2020-07-29
add 'title' prop support to View and Text components

Simek:
Currently when the [`title`](https://www.w3schools.com/tags/att_title.asp) property is used the attribute and value is being ignored. This prevents from adding typical for the web platform hint appearing on element hover. 

This PR adds `title` to the property picker `forwarProp` list in View and Text components, test for both components and extends also the TS types. As a result `title` prop is forwarded correctly and defined hint appears on the screen.

### Preview
demo

### DIFF
--- a/packages/react-native-web/src/exports/Text/__tests__/__snapshots__/index-test.js.snap
+++ b/packages/react-native-web/src/exports/Text/__tests__/__snapshots__/index-test.js.snap
@@ -26,3 +26,11 @@ exports[`components/Text nested 1`] = `
   />
 </div>
 `;
+
+exports[`components/Text prop "title" 1`] = `
+<div
+  class="css-text-901oao"
+  dir="auto"
+  title="I like the hover!"
+/>
+`;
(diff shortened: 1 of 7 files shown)


necolas:
Using the title attribute is strongly discouraged on the web due to accessibility shortcomings, and it isn't something that can be done cross-platform either. The recommended approach is to use tooltips and aria attributes. Those work with keyboard focus too at least.

Simek:
>Using the title attribute is strongly discouraged on the web due to accessibility shortcomings...
I strongly disagree, I don't know where you information is coming from. A lot of major and global sites are using this property (Wikipedia, GitHub, YouTube, etc.)

> ...it isn't something that can be done cross-platform either. 
I don't think that `react-native-web` is a cross-platform tool. `title` properties are working fine in every OS and in every browser. 

Besides that every out-of-tree platform supports some platform specific `props` and there is no problem with that at all. 

@necolas Please reconsider your approach and let me know how can I can do to improve the PR to fit you requirements.

necolas:
Just because the attribute is in the html spec isn't a reason to add it to the API of this framework. The link you included explains the many reasons it should not be used in the first place.

steida:
@Simek The point of RNfW is not to do anything that React allows. On the contrary, RNfW is a meaningful subset.

Issue #1680, opened 2020-07-20
Issue with react-test-renderer's create function (Part 2)


shogunatealex:
**The problem**
react-test-render isn't working for the Animated library from react-native. 

The simplified test will be a lot quicker to understand if you want to just skip to that.

Exact error is:
`TypeError: Cannot read property 'getScrollableNode' of null`

Full stack is:
   TypeError: Cannot read property 'getScrollableNode' of null

      4 | 
      5 | it('renders', () => {
    > 6 |       const tree = createRender(<Animated.View />).toJSON();
        |                    ^
      7 |       expect(tree).toMatchSnapshot();
      8 | });
      9 | 

      at AnimatedComponent._attachNativeEvents (node_modules/react-native-web/dist/cjs/vendor/react-native/Animated/createAnimatedComponent.js:113:44)
      at AnimatedComponent.componentDidMount (node_modules/react-native-web/dist/cjs/vendor/react-native/Animated/createAnimatedComponent.js:105:12)
      at commitLifeCycles (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:9432:22)
      at commitAllLifeCycles (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:10829:7)
      at HTMLUnknownElement.callCallback (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:2377:14)
      at Object.invokeGuardedCallbackDev (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:2427:16)
      at invokeGuardedCallback (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:2480:31)
      at commitRoot (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:11041:7)
      at node_modules/react-test-renderer/cjs/react-test-renderer.development.js:12492:5
      at Object.unstable_runWithPriority (node_modules/react-test-renderer/node_modules/scheduler/cjs/scheduler.development.js:255:12)
      at completeRoot (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:12491:13)
      at performWorkOnRoot (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:12420:9)
      at performWork (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:12328:7)
      at performSyncWork (node_modules/react-test-renderer/cjs/react-test-renderer.development.js:12302:3)
[...]

PR #1695, opened 2020-08-06
fix: createAnimatedComponent calls getScrollableNode on null component

cbstauss:
Checks for component's existence before calling getScrollableNode. Not sure if this is the right solution, but it passes the test case provided in issue #1680


necolas:
Yeah I don't know what the root issue is, and I'm tired of dealing with `Animated` changes. The implementation isn't suitable for performant use on web anyway, and it's only here for API compat with React Native. So I'll merge this to unblock the test renderer issue and circle back on the problem another time.

PR #1713, opened 2020-08-22
Support css animation events

RichardLindhout:
Since react-native-web does have CSS animation support I would like to subscribe to these event's since I'm improving the ripple effect for react-native-paper I need the onAnimationEnd property.

React-native-web will also use this event internally in the Modal PR.

### DIFF
--- a/packages/react-native-web/src/exports/View/index.js
+++ b/packages/react-native-web/src/exports/View/index.js
@@ -64,6 +64,10 @@ const forwardPropsList = {
   onMouseUp: true,
   onScroll: true,
   onWheel: true,
+  onTransitionEnd: true,
+  onAnimationStart: true,
+  onAnimationEnd: true,
+  onAnimationIteration: true,
   href: true,
   rel: true,
   target: true


RichardLindhout:
I don't get why, as it will add boilerplate.

Why these
```
  onBlur: true,
  onClick: true,
  onClickCapture: true,
  onContextMenu: true,
  onFocus: true,
  onKeyDown: true,
  onKeyUp: true,
  onTouchCancel: true,
  onTouchCancelCapture: true,
  onTouchEnd: true,
  onTouchEndCapture: true,
  onTouchMove: true,
  onTouchMoveCapture: true,
  onTouchStart: true,
  onTouchStartCapture: true,
onMouseDown: true,
  onMouseEnter: true,
  onMouseLeave: true,
  onMouseMove: true,
  onMouseOver: true,
  onMouseOut: true,
  onMouseUp: true,
  onScroll: true,
  onWheel: true,
```

But only skip these: https://reactjs.org/docs/events.html#animation-events
```
  onTransitionEnd: true,
  onAnimationStart: true,
  onAnimationEnd: true,
  onAnimationIteration: true,
```

necolas:
Because most of those are in RN and RNWindows. I have no intention of adding all the DOM event props to these components. And future versions of these frameworks may start using hooks and native event handlers anyway

RichardLindhout:
Ah thanks, I see, that would indeed result in some confusion maybe (as some people may think this work in RN too with the Animated.View etc.)

PR #1723, opened 2020-09-03
Add event parameter into Image.onLoad callback

euharrison:
Screen Shot 2020-09-03 at 13 57 00

It's missing the event parameter into onLoad callback from Image.

This PR just preserve the original event param and flow through the functions.

This is useful, for example, to catch the image dimensions when it's fully loaded with `e.path[0].width`

### DIFF
--- a/packages/react-native-web/src/exports/Image/index.js
+++ b/packages/react-native-web/src/exports/Image/index.js
@@ -236,10 +236,10 @@ const Image = forwardRef<ImageProps, *>((props, ref) => {
         function load(e) {
           updateState(LOADED);
           if (onLoad) {
-            onLoad();
+            onLoad(e);
           }
           if (onLoadEnd) {
-            onLoadEnd();
+            onLoadEnd(e);
           }
         },
         function error() {
--- a/packages/react-native-web/src/exports/Image/index.js
+++ b/packages/react-native-web/src/exports/Image/index.js
@@ -239,7 +239,7 @@ const Image = forwardRef<ImageProps, *>((props, ref) => {
             onLoad(e);
           }
           if (onLoadEnd) {
-            onLoadEnd(e);
+            onLoadEnd();
           }
         },
         function error() {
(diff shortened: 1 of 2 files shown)


packages/react-native-web/src/exports/Image/index.js
```diff
@@ -236,10 +236,10 @@ const Image = forwardRef<ImageProps, *>((props, ref) => {
         function load(e) {
           updateState(LOADED);
           if (onLoad) {
-            onLoad();
+            onLoad(e);
           }
           if (onLoadEnd) {
-            onLoadEnd();
+            onLoadEnd(e);
```
necolas, line 242: onLoadEnd doesn't include an event

packages/react-native-web/src/exports/Image/index.js
```diff
@@ -236,10 +236,10 @@ const Image = forwardRef<ImageProps, *>((props, ref) => {
         function load(e) {
           updateState(LOADED);
           if (onLoad) {
-            onLoad();
+            onLoad(e);
```
necolas, line 239: This should be `{ nativeEvent: e }` to mirror React Native

packages/react-native-web/src/modules/ImageLoader/index.js
```diff
@@ -119,7 +119,7 @@ const ImageLoader = {
     image.onerror = onError;
     image.onload = e => {
       // avoid blocking the main thread
-      const onDecode = () => onLoad();
+      const onDecode = () => onLoad(e);
```
necolas, line 122: This should be `{ nativeEvent: e }` to mirror React Native

necolas (changes requested):

packages/react-native-web/src/exports/Image/index.js
euharrison: If I understood correctly, this `onLoad` use the `ImageLoader.load`, so, if I change it there to `{ nativeEvent: e }`, here I can't recreate the event because it will be duplicated like `{ nativeEvent: { nativeEvent: e } }`

euharrison:
@necolas thanks for review, I updated the code and add a comment

ps: thanks for this amazing library, it's helping me a lot!

PR #1734, opened 2020-09-10
[chore]: remove useless code

hosseinmd:
In this PR I removed a code that was never running.
```js
if (isCustomProperty) {
      const name = isCustomProperty ? styleName : hyphenateStyleName(styleName);
``` 
This condition always assign styleName to name, because above this code you defined a condition which says isCustomProperty is true.

### DIFF
--- a/packages/react-native-web/src/vendor/react-dom/setValueForStyles/index.js
+++ b/packages/react-native-web/src/vendor/react-dom/setValueForStyles/index.js
@@ -11,7 +11,6 @@
  */
 
 import dangerousStyleValue from '../dangerousStyleValue';
-import hyphenateStyleName from 'hyphenate-style-name';
 
 /**
  * Sets the value for multiple styles on a node.  If a value is specified as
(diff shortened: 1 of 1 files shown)


necolas:
This code is still needed

hosseinmd:
I know you are one of the best, but I think this code `hyphenateStyleName(styleName)` clearly never runs. Why do you think this code is still needed?

sfcgeorge:
@necolas yep I think @hosseinmd is right. The duplicated outer conditional forces the inner conditional to always run one way. And so I assume the inner conditional is also backwards. I think it's meant to be like either of these:

```js
const name = isCustomProperty ? hyphenateStyleName(styleName) : styleName;
style.setProperty(name, styleValue);
// ^ using `setProperty` isn't needed here so equivalent to:
style[name] = styleValue;
```

or the more verbose way

```js
if (isCustomProperty) {
  style[hyphenateStyleName(styleName)] = styleValue;
} else {
  style[styleName] = styleValue;
}
```

But currently the hyphenation never runs so while the above makes more sense it is a behaviour change. I assume that fixes a bug; currently custom properties may be broken / not hyphenated as intended.

necolas:
Oh right I see

Issue #1735, opened 2020-09-13
I18Manager change direction not work well


hosseinmd:
**The problem**
i18Manager works well if you change direction before the components are rendered. But if change after components are rendered that has a problem with start end margins.
When changing i18Manager to RTL,  marginStart and marginEnd for some elements  won't change. But it works when defined before rendering components.

I provided an example that shows the issue.
In this example, after changing direction margin between buttons isn't right.
Simplified test case: https://codesandbox.io/s/muddy-sky-tjz8x?file=/src/App.js

**Expected behavior**
The expected margin between buttons to be equal in both directions.

**Environment (include versions). Did this work in previous versions?**

* React Native for Web (version): 0.13.7
* React (version): 16.13.1
* Browser: Chrome 85.0.4183.102

necolas:
You still have to trigger a react update, e.g., https://github.com/necolas/react-native-web/blob/master/packages/docs/src/apis/I18nManager/examples/LayoutRTL.js#L477

hosseinmd:
In an App, how can I update all layouts which are memorized? This is not about a page, this is about a few pages where users can go back to those.

PR #1736, opened 2020-09-13
feat: add logical inline style if supported

hosseinmd:
Issue: https://github.com/necolas/react-native-web/issues/1735
Logical inline style resolves the problem of an update needed when changing direction.

Description: https://developer.mozilla.org/en-US/docs/Web/CSS/margin-inline-start

Example: https://codesandbox.io/s/zealous-sammet-ejxzr?file=/src/App.js

### DIFF
--- a/packages/react-native-web/src/exports/StyleSheet/i18nStyle.js
+++ b/packages/react-native-web/src/exports/StyleSheet/i18nStyle.js
@@ -69,6 +69,19 @@ const PROPERTIES_I18N = {
   start: left
 };
 
+const isLogicalInlineStyleSupported = propertyKey => propertyKey in document.body.style;
+
+const LOGICAL_INLINE_STYLES = {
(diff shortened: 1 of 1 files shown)


hosseinmd:
Some tests failed, I could fix those, but I think tests should change.

necolas:
> Logical inline style resolves the problem of an update needed when changing direction.

It doesn't, because there are other properties that would still need a re-render to occur as well as any JS checking `isRTL`. This also doesn't work for SSR. And since the direction is generally changed only when locale changes, you usually often need to re-render the app anyway. Plus, this is the expectation in React Native too

hosseinmd:
Right, other properties are in logical inline styles too, they are supported by Firefox, but in chrome some of them are still experimental.
We can add all of them and check if browser support uses those.
React-Native has a bad idea, We have to restart App.

PR #1738, opened 2020-09-17
Implement BackHandler for web

Sharcoux:
I got that idea while working on a problem a bit similar in my RN + RNW app.

I don't know if this can be useful to other people.

I let it as a draft as I didn't test this exact version of the code, but I can do it if it has a chance to be accepted.

The idea is to push a fake duplicate of the app url on load, so that when we press "back", we detect that the history has changed, and we can take actions.

### DIFF
--- a/packages/react-native-web/src/exports/BackHandler/index.js
+++ b/packages/react-native-web/src/exports/BackHandler/index.js
@@ -8,16 +8,22 @@
  * @flow
  */
 
-function emptyFunction() {}
-
 const BackHandler = {
-  exitApp: emptyFunction,
-  addEventListener() {
+  listeners: [],
+  exitApp: history.previous,
+  addEventListener: (listener) => {
+    listeners.push(listener)
     return {
-      remove: emptyFunction
+      remove: this.removeEventListener(listener)
     };
   },
-  removeEventListener: emptyFunction
+  removeEventListener: (listener) => this.listeners.splice(this.listeners.indexOf(listener), 1)
 };
+
+history.pushState(null, '', window.location.href)
+window.addEventListener('popstate', (event) => {
+  if(!BackHandler.listeners.find(callback => callback()) history.previous()
+  else history.pushState(null, '', window.location.href)
+})
 
 export default BackHandler;
--- a/packages/react-native-web/src/exports/BackHandler/index.js
+++ b/packages/react-native-web/src/exports/BackHandler/index.js
@@ -36,6 +36,6 @@ const exitApp: () => {
 
 export default {
   addEventListener,
-  exitApp: history.previous,
+  exitApp,
   removeEventListener
 };
--- a/packages/react-native-web/src/exports/BackHandler/index.js
+++ b/packages/react-native-web/src/exports/BackHandler/index.js
@@ -7,7 +7,7 @@
  *
  * @flow
  */
-const listeners: Array<() => boolean | null | undefined)> = []
+const listeners: Array<() => boolean | null | undefined> = []
 const addEventListener = (event: 'hardwareBackPress', listener: () => boolean | null | undefined) => {
   listeners.unshift(listener);
 };
--- a/packages/react-native-web/src/exports/BackHandler/index.js
+++ b/packages/react-native-web/src/exports/BackHandler/index.js
@@ -30,7 +30,7 @@ window.addEventListener('popstate', (e) => {
 });
 
 const exitApp: () => {
-  listeners = [];
+  listeners.length = 0;
   history.previous();
 };
 
--- a/packages/react-native-web/src/exports/BackHandler/index.js
+++ b/packages/react-native-web/src/exports/BackHandler/index.js
@@ -7,35 +7,35 @@
  *
  * @flow
  */
-const listeners = [] as (() => boolean | null | undefined)[]
+const listeners: Array<() => boolean | null | undefined)> = []
 const addEventListener = (event: 'hardwareBackPress', listener: () => boolean | null | undefined) => {
-  listeners.unshift(listener)
-}
+  listeners.unshift(listener);
+};
 
 const removeEventListener = (event: 'hardwareBackPress', listener: () => boolean | null | undefined) => {
-  listeners.splice(listeners.indexOf(listener), 1)
-}
+  listeners.splice(listeners.indexOf(listener), 1);
+};
 
 // Hack for Chrome. See this issue: https://stackoverflow.com/questions/64001278/history-pushstate-on-chrome-is-being-ignored?noredirect=1#comment113174647_64001278
 const onFirstPress = () => {
   history.pushState(null, '', window.location.href)
   window.removeEventListener('focusin', onFirstPress)
-}
-window.addEventListener('focusin', onFirstPress)
+};
+window.addEventListener('focusin', onFirstPress);
 
 // Detect pressing back key on web browsers
 window.addEventListener('popstate', (e) => {
-  const handled = listeners.find(l => l())
-  if (handled) e.preventDefault()
-})
+  const handled = listeners.find(l => l());
+  if (handled) e.preventDefault();
+});
 
 const exitApp: () => {
-  listeners = []
-  history.previous()
-}
+  listeners = [];
+  history.previous();
+};
 
 export default {
   addEventListener,
   exitApp: history.previous,
   removeEventListener
-}
+};
--- a/packages/react-native-web/src/exports/BackHandler/index.js
+++ b/packages/react-native-web/src/exports/BackHandler/index.js
@@ -7,23 +7,35 @@
  *
  * @flow
  */
+const listeners = [] as (() => boolean | null | undefined)[]
+const addEventListener = (event: 'hardwareBackPress', listener: () => boolean | null | undefined) => {
+  listeners.unshift(listener)
+}
 
-const BackHandler = {
-  listeners: [],
-  exitApp: history.previous,
-  addEventListener: (listener) => {
-    listeners.push(listener)
-    return {
-      remove: this.removeEventListener(listener)
-    };
-  },
-  removeEventListener: (listener) => this.listeners.splice(this.listeners.indexOf(listener), 1)
-};
+const removeEventListener = (event: 'hardwareBackPress', listener: () => boolean | null | undefined) => {
+  listeners.splice(listeners.indexOf(listener), 1)
+}
 
-history.pushState(null, '', window.location.href)
-window.addEventListener('popstate', (event) => {
-  if(!BackHandler.listeners.find(callback => callback()) history.previous()
-  else history.pushState(null, '', window.location.href)
+// Hack for Chrome. See this issue: https://stackoverflow.com/questions/64001278/history-pushstate-on-chrome-is-being-ignored?noredirect=1#comment113174647_64001278
+const onFirstPress = () => {
+  history.pushState(null, '', window.location.href)
+  window.removeEventListener('focusin', onFirstPress)
+}
+window.addEventListener('focusin', onFirstPress)
+
+// Detect pressing back key on web browsers
+window.addEventListener('popstate', (e) => {
+  const handled = listeners.find(l => l())
+  if (handled) e.preventDefault()
 })
 
-export default BackHandler;
+const exitApp: () => {
+  listeners = []
+  history.previous()
+}
+
+export default {
+  addEventListener,
+  exitApp: history.previous,
+  removeEventListener
+}
--- a/packages/react-native-web/src/exports/BackHandler/index.js
+++ b/packages/react-native-web/src/exports/BackHandler/index.js
@@ -29,7 +29,7 @@ window.addEventListener('popstate', (e) => {
   if (handled) e.preventDefault();
 });
 
-const exitApp: () => {
+const exitApp = () => {
   listeners.length = 0;
   history.previous();
 };


necolas:
I don't understand what this is trying to achieve

Sharcoux:
The idea is to handle the "previous" button of the browser the same way as the "back" button on devices

necolas:
Can you be more specific? How does this achieve that?

Sharcoux:
The idea is that our app will have 2 pages in the history. Using `history.pushState` make us able to create that entry without reloading the page. We can learn when this state is being popped using `popstate`. This `popstate` becomes a reliable way to detect when we pressed the "previous" button.

From there we should probably define what we expect from a web implementation of BackHandler. In my own use case, I wanted this:
- 1 press on previous would trigger an action in my app, or display a custom message to ask leaving confirmation
- 2 press on previous within 1s would leave the app.

In the implementation I proposed here, it is a bit different:
- each time we press "previous", we call the list of listeners. If one listener return `true`, we stop propagating the event, and we stay in the app. If all listeners return `false` or if there is no listener, we leave the app and let the browser handle the "previous" as usual.

necolas:
Maybe this is something @EvanBacon can provide feedback about.

RichardLindhout:
This could be interesting to fix e.g. the route going back while there is a Modal opened. When the user presses the native back button it would be expected that the modal closes first instead of the route going back.

Note that this behaviour would also be the case on desktop devices with the back button.

This line is a little bit confusing, this is always the first item, correct? It also always empty a.t.m. since BackHandler. addEventListener is never called.
```
BackHandler.listeners.find(callback => callback()
``

**Comment by RichardLindhout** [comment]
We should also think about possibly interference of other libraries/code which use history.pushState. + popstate is called on page load if I remember correctly. But I think this would be a really great addition on UX on the web which a lot of (web) libraries have not fixed yet.

**Comment by Sharcoux** [comment]
Right now, I'm using this system to close modals : I add the close function to the listeners using BackHandler. addEventListener

The same way as for RN, this function is supposed to be called in the user application.

I'm not sure if I correctly understood and answered your question....

**Comment by RichardLindhout** [comment]
You're 100% right, I was just too focused on the code that I forgot about the external api which is used and returning booleans! 👍

**Comment by Sharcoux** [comment]
In our app, we chose to not use the browser history, because in my opinion, this is a feature that makes sense for a website, but not so much for an application, and moreover, passing data through the url tends to break the normal react data flow.

So, for projects that followed that same line, it would be a great addition. But for projects that still make use of history, we first need to decide exactly what the expected behavior of BackHandler should be. Or maybe we should just advice them to not use it for their case.

**Comment by RichardLindhout** [comment]
I think a lot of webapps use the url navigation, I also think this is very important for apps running in a browser for the UX of the user since they are used to the back button in their browser. But I think it could be possible to make both cases working.

**Comment by Sharcoux** [comment]
I'm not too much afraid of the technical challenge to make it work. I more wondering if it is possible to define a universal expected behavior for this feature in such a case and if it still makes sense.

I don't understand the part of your remark about the back button. Of course, the user can use the back button of their browser. It's just that instead of relying on urls we handle the state changes programmatically. I find this better because we can do it the react way.

**Comment by RichardLindhout** [comment]
I think it depends on the app but I like if an app support ctrl+click with url etc. On-topic: I would really like this feature!

**Comment by Sharcoux** [comment]
Line 26, if we change history.pushState with history.forward(), the behavior might start looking more like what we could expect, and it should play nicely with an app that uses history. It would just a an extra entry at the very beginning of the stack.
We should also remove this entry when leaving the page. We could do that on the beforeunload event.

Does anyone has a react native app using history that we could use for testing?

**Comment by Sharcoux** [comment]
Well, if someone comes with a project where this could be used, just tag me, we will see if we can make this work.

**Comment by RichardLindhout** [comment]
@Sharcoux I'm working on a codesandbox with react-router-dom. I'll use a custom modal where we can test the backhandler code ;)

**Comment by RichardLindhout** [comment]
This is hard, the pushState function works very differently from what I would have imagined

**Comment by RichardLindhout** [comment]
Here is the sandbox, but it won't work when you register multiple back handlers, we will need to connect the popState event to the right listener but the event don't has the right content in my case + it can interfere with other things.

**Comment by RichardLindhout** [comment]
I think we need to wait for a better api for navigation on the web, I think it will be too fragile

The Chrome team has a proposal for this which looks like it could fix what we are trying to fix
https://github.com/slightlyoff/history_api/issues

```
let previousRoutes = history.states()
if (previousRoutes.length === 0) {
    history.push('/product')
} else {
    history.goBack()
}
```

**Comment by RichardLindhout** [comment]
You can see in my example that too much states are pushed when multiple Backhandlers are registered (we need to remove the pushedState when the .remove() is called but this is not possible as far as I know.

This results in the back button not working for a few clicks after opening multiple modals and closing them

**Comment by RichardLindhout** [comment]
ezgif-7-b1131e0e2360

Here is the CodeSandbox: https://codesandbox.io/s/winter-sun-euxm7?file=/src/BackHandler.js

**Comment by RichardLindhout** [comment]
ezgif-5-47262fcdf03e


Here is a GIF showing the problem, and here is the codesandbox: https://codesandbox.io/s/winter-sun-euxm7?file=/src/BackHandler.js

**Comment by Sharcoux** [comment]
Ok, I don't know what could be the best way to work on this together, but here are my thoughts:

1. You used my code the way it was, but actually, I think that you should include the change I mentioned: line 26, use history.forward() instead of pushing a new state.
2. If we cannot remove a state and if we can use `history.states()`, we might check if history.states().length > 1 before pushing the state line 23, but I'm afraid that this could create weird behavior

**Comment by RichardLindhout** [comment]
history.states() can't be used unfortunately since this is a proposal and not implemented in browsers yet.

**Comment by RichardLindhout** [comment]
>1. You used my code the way it was, but actually, I think that you should include the change I mentioned: line 26 of my PR, use history.forward() instead of pushing a new state (line 42 of your code).

We can't go forward if there's no state?

> 2. If we cannot remove a state and if we can use history.states(), we might check if history.states().length > 1 before pushing the state line 23 of my PR, but I'm afraid that this could create weird behaviour

history.states() can't be used unfortunately since this is a proposal and not implemented in browsers yet. That proposal would make this feature a lot easier.


> 3. Line 8 of your code, I don't understand why you are pushing a new state when adding a listener
We need to push to the url so we can hook onto going back events, ideally we could maybe register this just once per page instead of every listener.

>4. Line 42 of your code, isn't it the opposite? If the event is handled, we don't want to go back, so we should move forward, but if the event is not handled, we should let the browser go back on step?

We can't prevent the browser from going back, so we need to work the other way around

**Comment by Sharcoux** [comment]
I changed some stuff, and this seems to be more or less what you wanted to achieve, but I might be wrong.

I'm just surprised about the history.forward(). If we are pressing back, this means that we are adding something in the stack isn't it? Maybe we need to delay a bit the call using setImmediate to wait for the history to be updated? This might be needed if the popstate listeners are called before the popped state is being added to the forward list.

**Comment by RichardLindhout** [comment]
Your solution works great!!

One thing: ```window.history.pushState(null, undefined, window.location.href);``` when importing the file seems off to do at import level.e. I'm curious if this one time pushState works if we add them to multiple pages.

And we need also need to change the window.history.previous to window.history.back

**Comment by Sharcoux** [comment]
The addState should really be added at import level. You need it only once and only if the BackHandler is being used.

**Comment by RichardLindhout** [comment]
I've removed some things which are not needed anymore + added multiple backHandler imports to test that out

I found 1 problem with the current implementation and that is when you want to go back and show a confirm alert (on the 'Prevent going back' going back page) it don't respects the history.goBack() function

https://codesandbox.io/s/thirsty-worker-se765?file=/src/BackHandler.js

**Comment by RichardLindhout** [comment]
I tested it out on an Android phone. It does not work the way it should.

**Comment by Sharcoux** [comment]
> I found 1 problem with the current implementation and that is when you want to go back and show a confirm alert (on the 'Prevent going back' going back page) it does not respect the history.goBack() function

I need to check how they implement their goBack function

> I tested it out on an Android phone. It does not work the way it should.

Could you elaborate @RichardLindhout ?

**Comment by RichardLindhout** [comment]
It closes all modals at once + the confirm modal does not show up on the 'prevent going back' page
ezgif-6-8d39424cfe1f

**Comment by RichardLindhout** [comment]
I think the current web implementation is just not good enough to support all use-cases of the BackHandler unfortunately, but let's hope I'm wrong.

**Comment by Sharcoux** [comment]
I'm sure that we can reach something. It might just take a bit more effort in referrencing all the specific cases than I expected.
I don't know how much more time I can spend on the subject right now because we are launching our product, but I'll get back on it when I'm more free.

**Comment by Sharcoux** [comment]
@RichardLindhout I think that I found the issue: [this thread](https://support.google.com/chrome/thread/8721521?hl=en) reveal that Chrome is waiting for the user to do something on the page before accepting to edit the history.

**Comment by Sharcoux** [comment]
@RichardLindhout I think that I found a solution using this code to execute the initial pushState (the one at import level in the BackHandler)

```javascript
  const onFirstPress = () => {
    history.pushState(null, '', window.location.href)
    window.removeEventListener('focusin', onFirstPress)
  }
  window.addEventListener('focusin', onFirstPress)
```

This will wait for the first user interaction with the website. In my use case, it worked. This needs to be tested, though.

Sharcoux:
@RichardLindhout I made an update. I'm using it almost like that in my application and it works. Can you confirm that it works as expected in your case?

hosseinmd:
BackHandler needs to handle escape (at least).
Probably you implemented it in the modal.

hosseinmd:
I tested this pr, not work in my project, I used react-navigation in that.

Sharcoux:
Could you describe what you expected to happen and what exactly went wrong?
As I said, on my side it is working as expected. Or you could be using it wrong, or you are using it in a way I didn't think of, or you are using it with a browser that doesn't handle this code the same way as Chrome does, or there is indeed a bug in this code, and it just happens to not matter in my use case. Anyway, any information  that can help me analyse those hypothesis is most welcome

hosseinmd:
I fork sandbox example, changed to react-navigation instead of react-router 
Please check it
https://codesandbox.io/s/goofy-kapitsa-txlm1?file=/src/App.js

Sharcoux:
Thanks. I will check this to understand what is going wrong exactly, and the difference with my usecase.

RichardLindhout:
> BackHandler needs to handle escape (at least).
> Probably you implemented it in the modal.

The escape is already handled by the modal and react-navigation modal I don't think we should handle it in the backhandler.

hosseinmd:
> The escape is already handled by the modal and react-navigation modal I don't think we should handle it in the backhandler.

I don't agree with you, react-navigation using React-Native backhandler in the native side if we implement backhandler in React-native-web, probably React navigation doesn't need to implement it for web separately.
However, The escape needed to implement in backhandler too, because we needed it in other wheres, that is not just for the modal.

RichardLindhout:
Why not use a onKeyDown listener in the other places?

hosseinmd:
So that could be cool if we can handle it just by backhandler for native and web both.

Sharcoux:
Wait a minute guys, I'm not sure to follow. Are you suggesting that BackHandler should react to the escape key? If so, I totally disagree. The role of BackHandler is just to provide the ability to react to the button "back" which would not be possible otherwise. Reacting to "Escape" is totally possible by various other ways and we cannot be sure that the expected behavior of Escape key would be the same as the Back key.

Sharcoux:
> I fork sandbox example, changed to react-navigation instead of react-router
> Please check it
> https://codesandbox.io/s/goofy-kapitsa-txlm1?file=/src/App.js

I checked. Indeed there is an issue when using react-navigation. For some reason, despite calling e.preventDefault() on the popstate event, the state is still being removed from the history. I also notice that the initial pushState is also being ignored in some cases... I wonder if react-navigation is not itself doing some nasty treatment to the history.
I didn't succeed yet to find the root cause behind this.

hosseinmd:
@satya164
I added @satya164, I hope he would help to solve this conflict.

satya164:
If you're looking for my feedback, I don't think `BackHandler` should react to `popstate`. The `popstate` event is not equivalent to user pressing browser back button. The name is misleading, think of it more like a `statechange` event rather than an entry being "popped".

It is triggered on both back and forward buttons, long pressing one of those buttons and selecting a history entry, calling `history.go` in the code etc., basically every time history changes, except on `pushState` and `replaceState`.

It's also not possible to prevent the event which makes it impossible to implement the `BackHandler` API.

> In our app, we chose to not use the browser history, because in my opinion, this is a feature that makes sense for a website, but not so much for an application

It definitely makes sense for apps as well. Native apps implement deep linking which isn't that different.

> I wonder if react-navigation is not itself doing some nasty treatment to the history

Nasty treatment is a strong word, React Navigation manages history on web unless you disable it. If you intefere with history as well, of course it won't be able to properly keep track of changes and things will break.

So you either let React Navigation manage history or pass `linking={{ enabled: false }}` to disable it and handle it yourself.

> I checked. Indeed there is an issue when using react-navigation. For some reason, despite calling e.preventDefault() on the popstate event, the state is still being removed from the history

You can't prevent `popstate`, the equivalent would be to keep track of history, and after `popstate`, push an entry with previous URL (though it's still not correct because `popstate` doesn't mean back).

The use case for "preventing going back" needs to be implemented in React Navigation. Since React Navigation keeps track of history already, it has a better idea of how to handle preventing a histiry change.

> I don't agree with you, react-navigation using React-Native backhandler in the native side if we implement backhandler in React-native-web, probably React navigation doesn't need to implement it for web separately.

React Navigation doesn't use `BackHandler` on web. It has separate code for history management which will React to browser history and keep the URL in sync with the navigation state. Implementing `BackHandler` here won't help because it's not possible to detect browser's back button, and even if we could, React Navigation needs to react to other URL changes as well, so it's not enough.

IIRC the `Linking` API integrates with history API, but I might be wrong.

RichardLindhout:
I agree with @satya164, I like the way react-navigation works with url's by default!

My opinion: I think the browser api in its current state can't work well enough with the possible use-cases of the BackHandler

Maybe this proposal: https://github.com/slightlyoff/history_api could fix some things but I don't know it will ever be implemented.

Sharcoux:
Thanks a lot @satya164 , This helped me understand better our options. And indeed, react-navigation is not doing anything nasty, I just didn't know that popState could not be prevented with preventDefault. It just happened to look as if it was working in my use case while it's not.

I see 2 options to solve the issue and keep it transparent for react-navigation or any other use of history :

### 1) Monkey patching the history. You can see the updated sandbox [here](https://codesandbox.io/s/modest-jennings-ysvfk?file=/src/BackHandler.js) (**warning** for some reason it doesn't work if you open the sandbox in a separated tab, but it works in the embed window and it will work in a real app)

This is cool because we can make sure that we target only the back button.

2) Using `history.forward()` to imitate a preventDefault on the popState event. The issue with this approach is that we don't know if the button pressed is back or forward, and when we reach the initial screen after a back button press was let through and the history is empty, we cannot press the back button until a new item calls history.pushState.

You can see it live [here](https://codesandbox.io/s/goofy-kirch-enq72?file=/src/BackHandler.js).

I have more faith in the monkey patching approach, but we need to see how the different browser react to this. I tested only on Firefox.

satya164:
You can't override browser back button by monkeypatching `window.history.back`. It works in your codesandbox probably because they have a JS implementation for their custom browser frame. Try it in a normal browser and it won't do anything.

What you're trying to do is not possible with the current browser APIs.

Sharcoux:
Indeed. The back button is not calling `window.history.back` contrary to what is happening on the sandbox.

Ok, so [here](https://codesandbox.io/s/goofy-kirch-enq72?file=/src/BackHandler.js) is my last proposal taking all this into account. It's basically mixing both approach into something that will work as intended I think. But, this time, it will definitely "do nasty things to the history". The forward button will always be disabled so that we can reliably use `popstate` event to detect press on the `back` button. It should really be used on an app where the main goal is to prevent the user from accidentally leaving the app, and provide him with the best possible behavior when pressing the `back` button.

It might be debatable if these constraints are acceptable or not for something that would reach the code base, but I think that React-Native-Web is aiming at creating cross-platform apps and with this in mind, even with its limitation, this BackHandler brings a most needed feature that is hard to achieve otherwise.

I'd like to hear your own opinion about it.

satya164:
> It should really be used on an app where the main goal is to prevent the user from accidentally leaving the app

You should use `window.addEventListener('beforeunload', callback)` if you want to prevent user from leaving the page for some reason (e.g. have some unsaved data etc.).

> But, this time, it will definitely "do nasty things to the history". The forward button will always be disabled so that we can reliably use popstate event to detect press on the back button.

Please don't, users have certain expectations from things running in the browser and have certain workflows. Just because you have a web app, it doesn't mean users will have different expectations. If you don't want to integrate your app with browser history, then that's fine (even though not ideal), but messing with back/forward buttons, and especially from framework code is a big nope.

Not to mention, users can still right click the back button and go back to any point in browser history and there's nothing you can do to prevent it. If you want to be able to prompt user from leaving the page, then use the API that browsers provide you to do that.

> even with its limitation, this BackHandler brings a most needed feature that is hard to achieve otherwise

I don't think this sort of code belongs in React Native Web. It shouldn't mess with browser history.

It makes more sense for navigation/routing libraries such as React Navigation to manage browser history. And routing libraries won't be able to have proper browser history integration if something else is messing with browser history.

React Navigation already has browser history integration [and API for preventing going back](https://reactnavigation.org/docs/preventing-going-back) from a screen. That API has some bugs, which makes it not work in nested navigators, however, it's a bug in the library and fixable. And it doesn't need doing unexpected things to browser history.

Sharcoux:
`beforeunload` is very limited and not well supported at all, making it a no go in the current state of things.

Sharcoux:
> It makes more sense for navigation/routing libraries such as React Navigation to manage browser history. And routing libraries won't be able to have proper browser history integration if something else is messing with browser history.

I totally understand your point of view. The thing is that, the same way that BackHandler doesn't belong to history libraries on native, it should not belong to react-navigation on browser. Reacting to the user input through the different buttons at their disposal should not be exclusive to some libraries. Especially because from my point of view, URLs management and React are not getting well along and so we should still be able to produce app without depending on navigation libraries. On the other hand, I agree that the BackHandler implementation should not destroy the workflow of the history and the integration with the other libs.

I might have a new idea. Let me try it out.

satya164:
> I totally understand your point of view. The thing is that, the same way that BackHandler doesn't belong to history libraries on native, it should not belong to react-navigation on browser.
> Reacting to the user input through the different buttons at their disposal should not be exclusive to some libraries.

Sure, but there's no API to detect back button in browser. It's not equivalent to the History API. You're not reacting to buttons. What's equavalent is the `Linking` API which already integrates with URLs and is already available in React Native Web.

On all other platforms, `BackHandler` exposes native back button to JavaScript. There's no such API to expose on Web. History API is about URL/history state changes, it's not about browser back button. The browser back button is just one of the ways to modify history. It's unlike the `BackHandler` API.

> Especially because from my point of view, URLs management and React are not getting well along and so we should still be able to produce app without depending on navigation libraries.

Would it be cool if we didn't need to use third party libs? Yes. But every addition of code has a cost and increases maintenance burden. I'd rather have a good base in core and let third party libs provide good quality integrations than have poorly maintained code in core. Not long ago, a lot of code was extracted to third party libraries from React Native core due to maintenance issues.

While it'll be nice to have such full integration built-in, it's not a trivial problem. I have spent weeks to implement it in React Navigation at minimum, and countless other weeks debugging edge cases, and there are still problems with it. There are libraries like React Router which spent years trying to solve this space. History API has a lot of problems, whatever hacks you end up with are going to break in ways you wouldn't have anticipated. I have already tried and I don't recommend it.

Sharcoux:
You've got me wrong. I'm not saying that using `react-navigation` might be wrong because it is a third-party library. I'm saying that for single pages applications, according to me (and we can disagree on this, no problem), it is wrong to use history because urls were conceived for accessing static data like a website, and applications are usually conceived with another paradigm where it is the state of the application that determine exactly what should be displayed to the user, and not the urls. Imagine if a video game was using different urls to decide what should be shown to the user. I think that it would completely break the state machine and make it very hard to maintain. There is another issue which is, from what I remember, that the url changes **before** React components are unmounted, breaking the way props are usually being propagated. For instance, if I have a component that I associate to a route `/client?id={clientId}`, when the url changes, the component will try to render without an id which lead to a lot of cases to handle and they are not always that trivial.

If we disagree on that it is probably because the kind of applications we have in mind are very different and it should be enough to just agree that both should be achievable, and that single page application would greatly profit from a common API for back buttons on mobile and browser.

About the current problem, there is actually a way to know if the `back` or the `forward` button got pressed using the history state to keep track of the history stack and know if we are moving forward or moving back. I have a working implementation, but before showing it I'm trying to understand something with react-navigation:

In [this fork](https://codesandbox.io/s/interesting-ellis-x0bwq?file=/src/Product.js) where I just commented out all the meaningful code from the BackHandler, so it is just vanilla `react-navigation`, when I navigate to the different tabs, the back button doesn't appear. Same goes when I run the project for real.

Sharcoux:
@hosseinmd maybe you can help me understand the fact that the back button doesn't work in your sandbox even when I remove the BackHandler part. I must be missing sth.

necolas:
Closing this PR because there doesn't seem to be a good way forward and the behavior isn't supported by APIs on web. Thanks for the discussion

hosseinmd:
@Sharcoux I tried but i don't have a good result. however, thank you. 
@necolas On some the web apps modals supporting this feature they implemented that by push state. i think this could be implemented but we have conflict with routes hanlder modules like react-navigation.

Sharcoux:
@hosseinmd @necolas I will provide my last code this weekend. I think that it can be considered as working as expected and respecting the constraints that were raised here. It should be compatible with react-navigation. The only issue is that I don't know react navigation and in the provided demo, it doesn't work as expected even without my code.
I think that this PR should remain opened and should be closed only if this last proposal is rejected.

Sharcoux:
@hosseinmd @satya164 [this implementation](https://codesandbox.io/s/elated-pine-v73kf?file=/src/BackHandler.js:0-2523) of the BackHandler. It should work fine with react-navigation.

hosseinmd:
checked it, but I don't understand what going on.

Sharcoux:
@hosseinmd well it's a simplified test case. WithUrls is using history to handle navigation and pressing back in the browser will move you one step back. WithoutUrls is using only the BackHandler and state management to navigate in the app. If someone knows how to correctly configurate react navigation to be able to use the history, you can replace the current navigation system.
What would you like to see happen there? I can improve the test case of you want.

RichardLindhout:
Let's hope this api will be accepted in all browsers it would probably fix this use case
https://github.com/WICG/app-history

Sharcoux:
For those interested, here is my final version that I'm using in my app with success on all browsers:

### BackHandler.ts

```ts
import type { BackHandler, BackPressEventName } from 'react-native'

const listeners = [] as (() => boolean | null | undefined | void)[]
const addEventListener = (
  event: BackPressEventName,
  listener: () => boolean | null | undefined | void
) => {
  listeners.unshift(listener)
  return { remove: () => removeEventListener(event, listener) }
}

const removeEventListener = (
  _event: 'hardwareBackPress',
  listener: () => boolean | null | undefined | void
) => {
  listeners.splice(listeners.indexOf(listener), 1)
}

const back = window.history.back
window.history.back = function () {
  const handled = listeners.find(l => l())
  if (!handled) {
    return back.apply(window.history, [])
  }
}

window.addEventListener('popstate', () => {
  const handled = listeners.find(l => l())
  if (!handled) {
    back.apply(window.history, [])
  }
  window.history.pushState(null, '', window.location.href)
})

// Hack for Chrome. See this issue: https://stackoverflow.com/questions/64001278/history-pushstate-on-chrome-is-being-ignored?noredirect=1#comment113174647_64001278
const onFirstPress = () => {
  window.history.pushState(null, '', window.location.href)
  window.removeEventListener('focusin', onFirstPress)
}
window.addEventListener('focusin', onFirstPress)

const exitApp = () => {
  listeners.length = 0
  back.apply(window.history, [])
}

export default {
  addEventListener,
  exitApp,
  removeEventListener,
} as BackHandler
```

You can also use this helper if you want to

```ts
import React from 'react'

import BackHandler from './BackHandler'

/** This helper will call onBackPress when we press the hardware back button */
const useBackKey = (onBackPress?: () => void) => {
  React.useEffect(() => {
    if (!onBackPress) return
    const onBack = () => {
      onBackPress()
      return true
    }
    BackHandler.addEventListener('hardwareBackPress', onBack)
    return () => BackHandler.removeEventListener('hardwareBackPress', onBack)
  }, [onBackPress])
}

export default useBackKey
```

zachdixon:
Thanks @Sharcoux , saved my life. The code above didn't work for me for some reason but the codesandbox you linked did. I feel like this should be the default behavior for react navigation if Platform.OS === 'web', no? Anyway, thanks again, this'll work in the meantime.

Issue #1741, opened 2020-09-17
PointerEvents="none" not working as expected on text elements


RichardLindhout:
**The problem**
![ezgif-7-69f064634e14](https://user-images.githubusercontent.com/6492229/93519478-9dd4c400-f92e-11ea-825f-210c5149bc94.gif)

I'm building a time picker component for React Native Paper with the PanResponder (which works great on the web!)
But I would expect the pointerEvents="none" to prevent the children from selecting text. But it does not. On iOS and Android text selection does not work in views with pointerEvents="none" even when the selectable={true} is used on <Text>

**How to reproduce**

Simplified test case: https://codesandbox.io/s/bold-shape-efdkb?file=/src/App.js

Steps to reproduce:
1. Click twice on the button
2. Text get selected while on Android/iOS it won't be selectable (even with selectable={true} on the Text

**Expected behavior**

I think we should add userSelect: 'none' (https://developer.mozilla.org/en-US/docs/Web/CSS/user-select) to the css when pointerEvents="none" is used. This way text inside buttons won't be selected if they have the pointerEvents disabled

**Environment (include versions). Did this work in previous versions?**

* React Native for Web (version): 0.13.8
* React Native: 0.63.2
* React (version): 16.13.1
* Browser: Google Chrome

**Additional context**
https://github.com/necolas/react-native-web/blob/9ed9231b04ae495718e76749036ca3cd2a81eb80/packages/react-native-web/src/modules/createDOMProps/index.js#L38
https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/StyleSheet/compile.js#L150

RichardLindhout:
I just forgot about selectable={false} on <Text> elements which is implemented already! However it would still be nice to have the same behaviour on the web but less important as I first thought :)

necolas:
That's right, you can disable text selection as needed. I think React Native gets text selection wrong by default and produces inaccessible experiences, especially on web. It looks like it also has a different relationship between `pointerEvents` and text selection behavior, and the fact that you cannot select text within a tree that has no pointer events looks to me like a bug in React Native. So I don't think it's worth changing RNWeb 

One thing we should do though, is make it so that this works:

```js
<Text selectable={false}>Some <Text selectable={true}>text</Text></Text>
```

RichardLindhout:
I agree this is more or less a bug in React Native!

> on the nested text
[...]

PR #1742, opened 2020-09-17
Fix text selecting in nested Text elements

RichardLindhout:
Based on this comment: https://github.com/necolas/react-native-web/issues/1741#issuecomment-694483457

### DIFF
--- a/packages/react-native-web/src/exports/Text/index.js
+++ b/packages/react-native-web/src/exports/Text/index.js
@@ -116,7 +116,7 @@ const Text = forwardRef<TextProps, *>((props, forwardedRef) => {
   const style = [
     props.style,
     numberOfLines != null && numberOfLines > 1 && { WebkitLineClamp: numberOfLines },
-    selectable === false && styles.notSelectable,
+    selectable === true ? styles.selectable : styles.notSelectable,
     onPress && styles.pressable
   ];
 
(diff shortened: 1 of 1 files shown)


RichardLindhout:
We can't use the 'auto' value because that does not work for nested elements
https://developer.mozilla.org/en-US/docs/Web/CSS/user-select

> The used value of auto is determined as follows:
> 
> On the ::before and ::after pseudo elements, the used value is none
> If the element is an editable element, the used value is contain
> Otherwise, if the used value of user-select on the parent of this element is all, the used value is all
> **Otherwise, if the used value of user-select on the parent of this element is none, the used value is none**
> Otherwise, the used value is text

Since we don't use before/after, editable or the all/contain property we can safely use the 'text' option :)

packages/react-native-web/src/exports/Text/index.js
```diff
@@ -116,7 +116,7 @@ const Text = forwardRef<TextProps, *>((props, forwardedRef) => {
   const style = [
     props.style,
     numberOfLines != null && numberOfLines > 1 && { WebkitLineClamp: numberOfLines },
-    selectable === false && styles.notSelectable,
+    selectable === true ? styles.selectable : styles.notSelectable,
```
necolas, line 119: This is not right as it makes text non-selectable by default (this is why the snapshots are failing). If `selectable` is `false`, it should set `userSelect:none`. If it is `true` it should set `userSelect:text`. If undefined/null do nothing

RichardLindhout: Changed!

Issue #162, opened 2016-07-08
External Linking


jesskrich:
Is there a way to create external links? I am using react-router for navigation but need to be able to connect to social media sites.

necolas:
`<Text href={href} />` should work

jesskrich:
<Text href={"http://google.com"}/>? Not working for me.

necolas:
Sorry that should have been:

```js
<Text accessibilityRole='link' href={href} />
```

See https://github.com/necolas/react-native-web/blob/master/docs/guides/accessibility.md

Example: http://codepen.io/necolas/pen/yJzVXB

skidding:
Is `<Text accessibilityRole='link' href={href} />` supported to work in native environments as well, or just on the web?

It didn't work for me, so this is the best "universal" Url component I could come with:

> Note that I just started with both RN and RNW

```jsx
export class Url extends Component {
  handlePress = () => {
    Linking.openURL(this.props.url);
  };

  render() {
    const { url } = this.props;

    // The purpose of TouchableWithoutFeedback is just to stop propagating
    // press events to parents
    // The View in the middle exists because otherwise TouchableWithoutFeedback 
    // would render the inner link unclickable.
    if (Platform.OS === 'web') {
      return (
        <TouchableWithoutFeedback>
          <View>
            <Link href={url} accessibilityRole="link" target="_blank">
              {url}
            </Link>
          </View>
        </TouchableWithoutFeedback>
      );
    }

    return (
      <TouchableWithoutFeedback onPress={this.handlePress}>
        <Link>{url}</Link>
      </TouchableWithoutFeedback>
    );
  }
}
```

Please let me know if there's an easier way. Thanks!

necolas:
https://github.com/necolas/react-native-web/blob/master/README.md#integrations

PR #1743, opened 2020-09-18
Provide better API for opening in new tab

Sharcoux:
Opening link in a new tab is something quite common on web nowadays. This use case should be handled by react-native-web.

Currently, the only solution is [described here](https://github.com/necolas/react-native-web/issues/162). I think that we can all agree that this is not very satisfying.

Here is my proposal. It should be transparent for native and provide the ability for web.

### DIFF
--- a/packages/react-native-web/src/exports/Linking/index.js
+++ b/packages/react-native-web/src/exports/Linking/index.js
@@ -22,9 +22,9 @@ const Linking = {
   getInitialURL(): Promise<string> {
     return Promise.resolve(initialURL);
   },
-  openURL(url: string): Promise<Object | void> {
+  openURL(url: string, target?: '_blank'): Promise<Object | void> {
     try {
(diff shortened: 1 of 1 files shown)


packages/react-native-web/src/exports/Linking/index.js
```diff
@@ -36,9 +36,9 @@ const Linking = {
   }
 };
 
-const open = url => {
+const open = (url: string, target?: '_blank') => {
   if (canUseDOM) {
-    window.location = new URL(url, window.location).toString();
+    window.open(url, target).focus();
```
ElForastero, line 41: `window.open()` is different from `window.location` assignment.

They are not interchangeable. 🤔 This will open a new tab.

Sharcoux: This is indeed the goal

PR #1746, opened 2020-09-21
fix: Prevent `ref` function from being being called more frequently than needed

### LINKED ISSUE #1665: Issue with opening modal after updating version to '0.13.1' from '0.12.0'
**The problem**
Issue with package: 'modal-enhanced-react-native-web' .
I am using this package for Modal. In v0.12.0, It was working fine. But I have to update `react-native-web` package due to [these](https://github.com/necolas/react-native-web/issues/1219) issues. But after updating, it is breaking modal functionalities. I am able to open Modal once, after that it doesn't  open (In 0.13.1). And different behaviours are there in each version above 0.13.0

**How to reproduce**
1. Fork it - https://codesandbox.io/s/6lx6ql1w5r
2. Add package `modal-enhanced-react-native-web`
3. Add code to open modal
4. Final Sandbox URL - https://codesandbox.io/s/2c4y1

**Note**: It has different behaviour in all version above v0.12.0 of `react-native-web`. I want modal behaviour same as its above 0.12.0

**Steps to reproduce:**
1. Open https://codesandbox.io/s/2c4y1
2. Click on button added to open modal
3. You will  see modal opened.
4. Click anywhere to close modal. Now it can be seen that modal fluctuates and then closes. In my use case, It closes opens modal when it is clicked anywhere on modal. (It has different and random behaviour on each version ^0.13.0). 

**Expected behavior**
- It should work same as it works in `v0.12.0` in the latest version also (v0.13.3)

**Environment (include versions). Did this work in earlier versions?**

* React Native for Web (version): Issue on versions above 0.12.0
* React (version): 16.12.0
* Browser: Chrome@latest

imnotjames:
In the `Picker`, `Pressable`, `Text`, `TextInput`, `TouchableHighlight`, `TouchableOpacity`, `TouchableWithoutFeedback`, and `View` component - when the `ref` prop is a function it will be called once per render of the component.  This is a new regression as of `0.13.0` - it was not present in `0.12.x` and was a side effect of the refactor to functional components.

Fixes #1665

### DIFF
--- a/packages/react-native-web/src/exports/Picker/__tests__/index-test.js
+++ b/packages/react-native-web/src/exports/Picker/__tests__/index-test.js
@@ -83,5 +83,36 @@ describe('components/Picker', () => {
       const { container } = render(picker);
       expect(findSelect(container).value).toBe('22');
     });
+
+    test('ref function is called when ref changes', () => {
+      const refMock = jest.fn();
+      const otherRefMock = jest.fn();
+
+      const { rerender } = render(<Picker ref={refMock} />);
+      expect(refMock).toHaveBeenCalled();
+
+      rerender(<Picker ref={otherRefMock} />)
+      expect(otherRefMock).toHaveBeenCalled();
+    });
+
+    test('ref function is not called every render', () => {
+      const refMock = jest.fn();
+
+      const { rerender } = render(<Picker ref={refMock} />);
+      expect(refMock).toHaveBeenCalledTimes(1);
+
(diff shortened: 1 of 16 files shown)


necolas:
Thanks for spotting the problem. Going to close this PR though, as I ported the approach we've been taking elsewhere to merging refs and memoizing the results: #1747

imnotjames:
No worries!  However it ends up getting fixed is great :)

PR #1750, opened 2020-09-22
fix: refs must be set before `mount` in modal content

imnotjames:
creates div for ModalPortal during the initial `render`
call before the `Modal` settles so we can capture refs before
the first component mount.

in libraries such as `react-native-modal` the `componentDidMount`
and `useEffect` hook expect the `ref` to be set before they are
called - and in the case of `Modal` it wasn't because of
the ModalPortal waiting for the mount to create the div for
a portal


necolas:
Thanks, this was folded into the Modal commit. Latest canary should include it

Issue #1602, opened 2020-04-30
AccessibilityInfo: add isReduceMotionEnabled


necolas:
**Is your feature request related to a problem? Please describe.**

React Native 0.62 added `isReduceMotionEnabled()` and the `reduceMotionChanged` event to `AccessibilityInfo` facebook/react-native@0090ab3

**Describe a solution you'd like**

Equivalent implementation for web using `matchMedia`.

**Additional context**

Looking for community PR.
cc @MoOx @jfrolich

necolas:
Fixed by #1754

PR #1754, opened 2020-09-24
AccessibilityInfo: add isReduceMotionEnabled

jaulz:
This is a first attempt to support `isReduceMotionEnabled` as mentioned in https://github.com/necolas/react-native-web/issues/1602. Please let me know if we should follow this approach.

### DIFF
--- a/packages/react-native-web/src/exports/AccessibilityInfo/index.js
+++ b/packages/react-native-web/src/exports/AccessibilityInfo/index.js
@@ -7,13 +7,20 @@
  * @flow
  */
 
-function emptyFunction() {}
-
 function isScreenReaderEnabled(): Promise<*> {
   return new Promise((resolve, reject) => {
     resolve(true);
   });
 }
+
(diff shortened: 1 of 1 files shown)


packages/react-native-web/src/exports/AccessibilityInfo/index.js
```diff
@@ -7,14 +7,21 @@
...
 
-function emptyFunction() {}
-
 function isScreenReaderEnabled(): Promise<*> {
   return new Promise((resolve, reject) => {
     resolve(true);
   });
 }
 
+const prefersReducedMotionMedia = window.matchMedia('(prefers-reduced-motion: reduce)')
```
necolas, line 16: This needs a canUseDOM guard for SSR context

jaulz: Okay, here you go 😊

necolas:
Thanks!

PR #1762, opened 2020-10-09
Skip ResizeObserver warning during SSR

marklawlor:
The ResizeObserver is not available when using `react-native-web` during a server side render, however the warning of a missing polyfill still triggers.

This PR skips the warning when `react-native-web` is used during SSR.


necolas:
Thanks!

PR #1767, opened 2020-10-13
Remove collapsable from createAnimatedComponent.js

TikiTDO:
Sending `false` to `collapsable` causes the browser to complain that it is not a valid value. Suggest change to `undefined` as per error in browser.

### DIFF
--- a/packages/react-native-web/src/vendor/react-native/Animated/createAnimatedComponent.js
+++ b/packages/react-native-web/src/vendor/react-native/Animated/createAnimatedComponent.js
@@ -169,7 +169,7 @@ function createAnimatedComponent(Component: any, defaultProps: any): any {
           // have to make sure the view doesn't get optimized away because it cannot
           // go through the NativeViewHierarchyManager since it operates on the shadow
           // thread.
-          collapsable={false}
+          collapsable={undefined}
         />
       );
     }


necolas:
You'll have to explain a bit more. How are you using this function? If you're using it directly with DOM element types that isn't supported in this library

TikiTDO:
The warning I'm trying to fix derives from here:

https://github.com/danilowoz/react-content-loader/blob/master/src/native/Svg.tsx#L14

In this case `react-content-loader/native` creates `AnimatedLinearGradient` by wrapping the `LinearGradient` from `react-native-svg` using `Animated.createAnimatedComponent`. The `react-native-web` implementation of `Animated.createAnimatedComponent` then adds `collapsable={false}` to the component, which in turn causes a big "DON'T DO THIS" type of warning below:

image

I figured that this was best tackled at the source, since neither of the other two libs have anything to do with the `collapsable` prop.

necolas:
One problem lies in the SVG library, because it is forwarding all props to the DOM. And another is in React Native because there is no platform guard around `collapsable`, which is an Android-only feature, and happens to collide with a web attribute that has a different purpose.

I'll monkey-patch the module here, since both of those projects are a bit slow to fix bugs for the web platform.

TikiTDO:
It's not too hard to work around the issue, since the `react-content-loader` lib offers a web alternative as well. My solution was just to `Platform` guard it. 

With this PR I just searched through `react-native-web` and didn't see anything actually using the code, which is why I figured the `undefined` would work more generically.

PR #1770, opened 2020-10-15
fix: prevent constant image reloading when source given as an object

comp615:
**Problem**
Image now uses hooks for memoization, as part of this the image loader is re-created or re-run when the source changes. One of the options for setting the source, however, is an object, and it's a common pattern to create these inline.

**Solution**
To help prevent people from footgunning themselves when creating Image sources inline, memoize on just the uri instead of the entire source. By doing this, we can do a string comparison instead of an object comparison which runs a high risk of preventing memoization (see #1766 too)

The uri computation is cheap, and the source width and height do not currently affect the useEffect loader, so this should be a safe optimization


necolas:
This patch only contains a new test. Is there a missing commit?

comp615:
Eek yes. One second.

PR #1777, opened 2020-10-20
fix: Don't re-run View ref processes when styles change

comp615:
**Problem**
Build on top of previous commits (#1770 #1766), we ideally want to only trigger ref changes (and the work associated with them) if the dom node has actually changed. 

Right now the argument `platformMethodsRef` is most subject to change which causes a new ref. This is because as an argument to its own memoize function, it takes style. Previously, I patch this so that classList was constant. However a more complete solution is to actually not memoize on these props at all.

NOTE: I'm less sure that this is the right thing to do as I don't have as full an understanding of the nuance in setNativeProps, but I _think_ it should work, so take it with an additional grain of salt.

**Solution**
Instead of memoizing on classList and style which are very hard to control (since they come from the developer), we can observe that these values do not actually affect what usePlatformMethods does. Instead, we simply need the values to be available _if_ `setNativeProps` is called. We can achieve this by reworking the argument into a ref and memoizing on that instead. This means it should be constant and in general usePlatformMethods will now only be computed once per component.

In addition, by keeping it as a ref and argument. We are able to accurately use the values when it comes time to setNativeMethods.

### DIFF
--- a/packages/react-native-web/src/modules/usePlatformMethods/index.js
+++ b/packages/react-native-web/src/modules/usePlatformMethods/index.js
@@ -44,29 +44,33 @@ function setNativeProps(node, nativeProps, classList, pointerEvents, style, prev
  * Adds non-standard methods to the hode element. This is temporarily until an
  * API like `ReactNative.measure(hostRef, callback)` is added to React Native.
  */
-export default function usePlatformMethods(setNativePropsStyles: {|
-  current: ?{|
-    classList?: Array<string | boolean>,
-    style?: $PropertyType<ViewProps, 'style'>,
-    pointerEvents: $PropertyType<ViewProps, 'pointerEvents'>
-  |}
-|}) {
+export default function usePlatformMethods({
+  classList,
+  pointerEvents,
+  style
+}: {
+  classList?: Array<string | boolean>,
+  style?: $PropertyType<ViewProps, 'style'>,
+  pointerEvents: $PropertyType<ViewProps, 'pointerEvents'>
+}) {
   const previousStyleRef = React.useRef(null);
+  const setNativePropsArgsRef = React.useRef(null);
+  setNativePropsArgsRef.current = { classList, pointerEvents, style };
 
-  return React.useMemo(
-    () => (hostNode: any) => {
+  return React.useCallback(
+    (hostNode: any) => {
       if (hostNode != null) {
         hostNode.measure = callback => UIManager.measure(hostNode, callback);
         hostNode.measureLayout = (relativeToNode, success, failure) =>
           UIManager.measureLayout(hostNode, relativeToNode, failure, success);
         hostNode.measureInWindow = callback => UIManager.measureInWindow(hostNode, callback);
         hostNode.setNativeProps = nativeProps => {
-          const { classList, style, pointerEvents } = setNativePropsStyles.current || {};
+          const { classList, style, pointerEvents } = setNativePropsArgsRef.current || {};
           setNativeProps(hostNode, nativeProps, classList, pointerEvents, style, previousStyleRef);
         };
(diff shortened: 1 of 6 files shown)


packages/react-native-web/src/modules/usePlatformMethods/index.js
```diff
@@ -43,22 +44,33 @@ function setNativeProps(node, nativeProps, classList, pointerEvents, style, prev
...
+  style?: $PropertyType<ViewProps, 'style'>,
+  pointerEvents: $PropertyType<ViewProps, 'pointerEvents'>
+}) {
+  const previousStyleRef = React.useRef(null);
+  const setNativePropsArgsRef = React.useRef(null);
+  setNativePropsArgsRef.current = { classList, pointerEvents, style };
 
-  return useMemo(
-    () => (hostNode: any) => {
+  return React.useCallback(
```
magus, line 60: We should use React.useRef here instead to create this function once and return the .current of that ref. This will guarantee ref equality but always use the latest values. since this matches the intent and correctly is guaranteed by the behavior of useRef I think it’s better than relying on memoization which is not guaranteed ( https://reactjs.org/docs/hooks-reference.html  )

useCallback in particular seems odd here semantically too.

comp615: Got it. Updated!

packages/react-native-web/src/modules/usePlatformMethods/index.js
```diff
@@ -43,22 +44,30 @@ function setNativeProps(node, nativeProps, classList, pointerEvents, style, prev
...
  * API like `ReactNative.measure(hostRef, callback)` is added to React Native.
  */
-export default function usePlatformMethods(props: Object) {
+export default function usePlatformMethods({
+  classList,
+  pointerEvents,
+  style
+}: {
+  classList?: Array<string | boolean>,
+  style?: $PropertyType<ViewProps, 'style'>,
```
necolas, line 53: This type can't be correct because components other than `View` set `style` too

packages/react-native-web/src/modules/usePlatformMethods/index.js
```diff
@@ -43,22 +44,30 @@ function setNativeProps(node, nativeProps, classList, pointerEvents, style, prev
...
+      hostNode.measureLayout = (relativeToNode, success, failure) =>
+        UIManager.measureLayout(hostNode, relativeToNode, failure, success);
+      hostNode.measureInWindow = callback => UIManager.measureInWindow(hostNode, callback);
+      hostNode.setNativeProps = nativeProps => {
+        const { classList, style, pointerEvents } = setNativePropsArgsRef.current || {};
+        setNativeProps(hostNode, nativeProps, classList, pointerEvents, style, previousStyleRef);
+      };
+    }
+    return hostNode;
+  }).current;
```
necolas, line 72: This no longer returns a ref. I don't think we want to set the initial value of the ref to be a function either. The code that was deleted was memoizing a function ref.

packages/react-native-web/src/modules/usePlatformMethods/index.js
comp615: Good point, and I see they have different style props. OK, I made this generic now which is hand-wavy right. But because I used a star it's not going to really do anything.

As an aside, the type of this function doesn't really matter because we'd have to fix all the call points. Many of them use `pick` to get the props. And `pick` causes all type-safety to be lost because it uses Object. I think the correct signature is `T => $Shape<T>` or something like that. But that causes a whole bunch of issues which would need correcting (for instance because we later assign domProps to what would then be viewProps so it gets mad). So maybe an issue for another diff...

PR #1780, opened 2020-10-22
fix: Prevent onClick from being called when certain roles are disabled

### LINKED ISSUE #1779: View should not fire onClick when disabled is set in 0.13+
**The problem**

**How to reproduce**
https://codesandbox.io/s/laughing-nobel-21jlt?file=/src/App.js

Steps to reproduce:
1. Click the text, observe callback is fired
2. Change RNW version to 0.12
3. Observe clicking does not fire the callback

* React Native for Web (version): 0.13+

**Additional context**
Previously this was handled by code which looked for disabled on button-like roles and removed the onClick handler: https://github.com/necolas/react-native-web/blob/0.12.3/packages/react-native-web/src/exports/createElement/index.js#L58-L76

In the new code, there's a conflict...the disabling is only done if data-focusable is true: https://github.com/necolas/react-native-web/blob/6d04e7243e8e0958eec473429f0f4563bc8e41b9/packages/react-native-web/src/modules/createDOMProps/index.js#L225-L236, but that's only set if the element is focusable, which never happens when disabled is set: https://github.com/necolas/react-native-web/blob/6d04e7243e8e0958eec473429f0f4563bc8e41b9/packages/react-native-web/src/modules/createDOMProps/index.js#L154-L177

comp615:
Closes #1779 

**Problem**
In older versions of RNW, there was code in `createElement` which would remove event handlers from elements with the disabled prop: https://github.com/necolas/react-native-web/blob/0.12.3/packages/react-native-web/src/exports/createElement/index.js#L58-L76

In newer versions, this code was moved to `createDOMProps`. However, it changed a few things:
- It no longer handles events aside from onClick
- It does not respect "disabled" anymore (see attached issue)

This PR aims to address just the second issue with the disabled conflict

**Solution**
Currently, we require `data-focusable` to be set to true in order to apply any override handling. However, when disabled is set...`data-focusable` is never set, which means we never override `onClick`.

Instead, we can more directly check the roles in the outer if, so that the code still runs when disabled is true. While this makes it a little more detached from the logic higher up in the file, it fixes the functionality. It should not let anything else through because when disabled is true, only the onClick overriding runs

**Questions**
- I see `textbox` had similar logic in some places, but I think adding it to this handling would cause weird keydown behavior so I skipped it.
- I moved the tests from createElement and expanded them, I figured the logic isn't in createElement anymore so that could go. These did not catch the bug because they only tested the keyDown behavior...not the click behavior.
- I didn't bring back the [big list of prevented events](https://github.com/necolas/react-native-web/blob/0.12.3/packages/react-native-web/src/exports/createElement/index.js#L39-L54). The rest are probably less used...but there are the same bugs present there...however it seemed like an intentional choice was made to only worry about `onClick` (as opposed to say `onContextMenu`). Maybe recommended to convert the onclick override to do all those other ones as well

### DIFF
--- a/packages/react-native-web/src/modules/createDOMProps/__tests__/index-test.js
+++ b/packages/react-native-web/src/modules/createDOMProps/__tests__/index-test.js
@@ -140,6 +140,112 @@ describe('modules/createDOMProps', () => {
     });
   });
 
+  describe('onClick handlers', () => {
+    const callsOnClick = (component, accessibilityRole, disabled = false) => {
+      const onClick = jest.fn();
+      const event = { stopPropagation: jest.fn() };
+      const finalProps = createDOMProps(component, { accessibilityRole, disabled, onClick });
+      finalProps.onClick(event);
+      return onClick.mock.calls.length === 1;
+    };
+
+    test('are passed through and called for various roles', () => {
+      expect(callsOnClick('div', 'link')).toBe(true);
+      expect(callsOnClick('div', 'button')).toBe(true);
+      expect(callsOnClick('div', 'textbox')).toBe(true);
+      expect(callsOnClick('div', 'menuitem')).toBe(true);
+      expect(callsOnClick('div', 'bogus')).toBe(true);
+      expect(callsOnClick('a')).toBe(true);
+      expect(callsOnClick('button')).toBe(true);
+      expect(callsOnClick('input')).toBe(true);
+      expect(callsOnClick('select')).toBe(true);
+      expect(callsOnClick('textarea')).toBe(true);
+      expect(callsOnClick('h1')).toBe(true);
+    });
+
+    test('are not passed through or called for various roles when disabled is true', () => {
+      expect(callsOnClick('div', 'link', true)).toBe(false);
+      expect(callsOnClick('div', 'button', true)).toBe(false);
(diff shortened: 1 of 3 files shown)


necolas:
Thanks. This is where I wish the web had an API to disable "all" events, as it makes it very tedious to develop the behaviors for various WAI-ARIA patterns. If you look at the repo I gave you access to, you'll see things become a bit simpler with that model as every behavior is a hook that you opt into, and they can be disabled.

packages/react-native-web/src/modules/createDOMProps/__tests__/index-test.js
```diff
@@ -140,6 +140,112 @@ describe('modules/createDOMProps', () => {
...
+    // These tests fire two events, one space and one enter to ensure each is handled correctly
+    const callsOnClick = key => (
+      component,
+      accessibilityRole,
+      disabled = false,
+      callOnSpace = false
+    ) => {
+      const onClick = jest.fn();
+      const onKeyDown = jest.fn();
+      const event = { key: 'Enter' };
```
necolas, line 192: Just noting that I changed this in the merged patch so that `' '` was used for the spacebar tests

PR #1811, opened 2020-11-13
Fix effect dependency in useElementLayout

giuseppeg:
`ref` is immutable and the effects should run when its current value changes


necolas:
This change is not safe because it is essentially reading the ref value during render. What is the scenario in which this issue surfaces?

giuseppeg:
Oh right sorry for the confusion. I was confused by the fact that you referenced the ref in the deps array.

The scenario is not relevant for React Native for Web nor for the app I was working on which broke for other reasons.

giuseppeg:
@necolas by the way do you know if these kind of subscriptions (ResizeObserver/matchMedia) are concurrent mode safe?

PR #1821, opened 2020-11-25
feat: implement refocusing trigger-element after closing modal

RafikiTiki:
In a project which I'm part of right now we're working towards compatibility with WCAG spec. When using current implementation of the `<Modal />` component I've seen that you've handled focus trap problem in very clean manner but there's one more thing that I believe that should be implemented to be even more compliant with the spec.

> A Web page implements modal dialogs via scripting. When the trigger button is activated, a dialog opens and focus is set to the first interactive element in the dialog. As long as the dialog is open, focus is limited to the elements of the dialog. **When the dialog is dismissed, focus returns to the button or the element following the button.**

source: https://www.w3.org/WAI/WCAG21/Understanding/focus-order Examples point 3

Right now after closing modal, focus is set to the `<body>` element. User relaying solely on keyboard for navigation now needs to _tab_ his/hers way to the button once more which is frustrating and definitely not a good UX.

This PR is implementing _refocus element that triggered modal after closing it_ mechanism.

### DIFF
--- a/packages/react-native-web/src/exports/Modal/ModalFocusTrap.js
+++ b/packages/react-native-web/src/exports/Modal/ModalFocusTrap.js
@@ -131,7 +131,7 @@ const ModalFocusTrap = ({ active, children }: ModalFocusTrapProps) => {
       lastFocusedElementOutsideTrap.current = document.activeElement;
       return () => {
         if (lastFocusedElementOutsideTrap.current) {
-          lastFocusedElementOutsideTrap.current.focus();
+          UIManager.focus(lastFocusedElementOutsideTrap.current);
         }
       };
     }
--- a/packages/react-native-web/src/exports/Modal/ModalFocusTrap.js
+++ b/packages/react-native-web/src/exports/Modal/ModalFocusTrap.js
@@ -71,6 +71,7 @@ export type ModalFocusTrapProps = {|
 |};
 
 const ModalFocusTrap = ({ active, children }: ModalFocusTrapProps) => {
+  const lastFocusedElementOutsideTrap = useRef<?HTMLElement>();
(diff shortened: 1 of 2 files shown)


packages/react-native-web/src/exports/Modal/ModalFocusTrap.js
```diff
@@ -123,6 +124,19 @@ const ModalFocusTrap = ({ active, children }: ModalFocusTrapProps) => {
     }
   }, [active]);
 
+  // To be fully compliant with WCAG we need to refocus element that triggered opening modal
+  // after closing it
+  useEffect(() => {
+    if (canUseDOM) {
+      lastFocusedElementOutsideTrap.current = document.activeElement;
+      return () => {
+        if (lastFocusedElementOutsideTrap.current) {
```
necolas, line 133: This value needs to be stored in the closure rather than read in the cleanup callback

packages/react-native-web/src/exports/Modal/ModalFocusTrap.js
```diff
@@ -123,6 +124,19 @@ const ModalFocusTrap = ({ active, children }: ModalFocusTrapProps) => {
...
+          UIManager.focus(lastFocusedElementOutsideTrap.current);
```
necolas, line 134: Probably need to check if this element is still in the document

necolas (changes requested):
We will need unit tests for this behavior

RafikiTiki:
@necolas sorry that it did take so long but I've finally found some time to update the PR. I've refactored things you've pointed out and also implemented tests for refocusing mechanism. Have a good day :)

necolas:
Thanks!

necolas (approved):

PR #1824, opened 2020-11-26
fix: send onRequestClose to the correct modal even with animations

### LINKED ISSUE #1817: Escape key will close all modals
**The problem**
Pressing escape key will close all modals when `animationType !== 'none'`

**How to reproduce**
https://codesandbox.io/s/red-snowflake-qcstb

Steps to reproduce:
1. click open modal 
2. click open modal in modal 1
3. click open modal in modal 2
4. press escape

**Expected behavior**
Pressing escape should only close the top most modal like `animationType === 'none'`

* React Native for Web (version): 0.14.8
* React (version): 16.13.1
* Browser: chrome 86

imnotjames:
the animation end handler does not correctly check that the animation end event is coming from the correct location - as such, the wrong modal gets marked as the "active" modal.  when that happens, the `Escape` key does not get routed to the correct modal.

fixes #1817

### DIFF
--- a/packages/react-native-web/src/exports/Modal/__tests__/index.js
+++ b/packages/react-native-web/src/exports/Modal/__tests__/index.js
@@ -2,7 +2,7 @@
 
 import Modal from '..';
 import React from 'react';
-import { render } from '@testing-library/react';
+import { fireEvent, render } from '@testing-library/react';
 
 describe('components/Modal', () => {
   test('visible by default', () => {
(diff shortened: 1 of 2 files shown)


packages/react-native-web/src/exports/Modal/__tests__/index.js
```diff
@@ -426,4 +426,74 @@ describe('components/Modal', () => {
...
+          </Modal>
+        </Modal>
+      </>
+    );
+
+    // This is kind of ugly but I can't find a better way to target just the animation div
+    const animationAElement = getByTestId('a').parentElement.parentElement.parentElement
+      .parentElement;
+    const animationBElement = getByTestId('b').parentElement.parentElement.parentElement
+      .parentElement;
```
imnotjames, line 489: I don't like these lines ): But I couldn't find a better way to target the element in particular that I wanted

packages/react-native-web/src/exports/Modal/__tests__/index.js
```diff
@@ -426,4 +426,74 @@ describe('components/Modal', () => {
...
+
+    fireEvent.animationEnd(animationAElement);
+    fireEvent.animationEnd(animationBElement);
```
necolas, line 492: If these lines are removed, the test fails. Shouldn't it still pass if animation events didn't occur?

packages/react-native-web/src/exports/Modal/__tests__/index.js
```diff
@@ -426,4 +426,74 @@ describe('components/Modal', () => {
...
+      </>
+    );
+
+    fireEvent.keyUp(document, { key: 'Escape' });
+
+    expect(spyA).toHaveBeenCalledTimes(0);
+    expect(spyB).toHaveBeenCalledTimes(1);
+  });
+
+  test('escape key fires onRequestClose for top modal only with animation', () => {
```
necolas, line 457: I think this test also needs to pass if animations are not used

necolas (changes requested):

necolas:
Is there going to be any follow up?

packages/react-native-web/src/exports/Modal/__tests__/index.js
imnotjames: If an animation is used, we wait until the animation completes to send the `onDismiss` event from the modal's animation layer.  This means if an animation is used it will be considered "active" until the animation completes.  While a modal is active it will receive the "escape" events.  This is to match up with how ReactNative & iOS seem to do this - at least that's what I vaguely recall.  This could be incorrect or an undesired behavior, I'm not sure.

If we don't want that behavior and instead want a modal to be inactive as soon as it's requested but before it actually closes we can do that, but it'd be a slightly larger change.

Is that what you're requesting?

imnotjames:
Sorry :) I forgot about this PR & following up on it.  Question for you as a response to the requested change.

necolas (approved):
I can't remember what those comments were about tbh hah

Issue #1835, opened 2020-12-05
href does not work on TouchableWithoutFeedback with accessibilityRole=link


RichardLindhout:
**The problem**

I want to add a href a TouchableWithoutFeedback (this is used in the ui library I use react-native-paper) and I want the ctrl+click experience with urls on the web.
![Schermafbeelding 2020-12-05 om 19 29 25](https://user-images.githubusercontent.com/6492229/101260445-3578d000-3730-11eb-87db-d88c6f96ee67.png)

However the href does not get added to the dom, but it does with TouchableHighlight

**How to reproduce**
https://codesandbox.io/s/floral-fire-n7zlu?file=/src/App.js

**Expected behavior**
Href get's also applied with TouchableWithoutFeedback
**Environment (include versions)**

* React Native for Web (version): * (latest)
* React (version): 16.13.1
* Browser: Chrome

necolas:
Those props should be set on the element that touchable without feedback is wrapping.

See the notes in the docs: https://reactnative.dev/docs/touchablewithoutfeedback

>  Importantly, `TouchableWithoutFeedback` works by cloning its child and applying *responder props* to it. It is therefore required that any intermediary components *pass through those props* to the underlying React Native component.

So that component simply forwards props to the element it wraps, mostly responder props and some accessibility props. Nothing else. It will eventually be deprecated in favor of `Pressable`.

RichardLindhout:
Ok thanks for the info! I have a PR open for the library to use Pressable https://github.com/callstack/react-native-paper/pull/2281. So will continue the discussion with the maintainers of react-native-paper.

PR #1836, opened 2020-12-05
pass href attributes and mouse events to TouchableWithoutFeedback

RichardLindhout:
Fixes https://github.com/necolas/react-native-web/issues/1835
+ the onMouse props are great since I use preload queries and components on mouseOver the link before it get clicked to speed up data loading.

### DIFF
--- a/packages/react-native-web/src/exports/TouchableWithoutFeedback/index.js
+++ b/packages/react-native-web/src/exports/TouchableWithoutFeedback/index.js
@@ -60,7 +60,21 @@ const forwardPropsList = {
   onBlur: true,
   onFocus: true,
   onLayout: true,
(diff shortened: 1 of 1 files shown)


RichardLindhout:
Schermafbeelding 2020-12-05 om 19 40 24
Some background context: I'm working on a cross-platform navigation abstraction on top of react-native-navigation and react-router-dom with support for preloading so I would really like to also have the mouse events since some UI libraries use the TouchableWithoutFeedback on the web

RichardLindhout:
Closed since it should be provided to the underlying element according to the docs as Nicolas said here:
https://github.com/necolas/react-native-web/issues/1835

PR #1837, opened 2020-12-05
Add better PermissionsAndroid mock

sesm:
For our usecase (web version of Storybook) we needed a mock of PermissionsAndroid.
Turns out, it's easy to create a high quality mock with this steps:
- Copy original FB source
- Remove native imports
- For every function leave only the fallback code that was originally under `Platform.OS !== 'android'` and delete the rest


necolas:
Best not to import Android-specific modules in web code paths. You can use platform file extensions to write Android-only code rather than bloat the web bundle with useless code.

sesm:
@necolas let me explain our usecase: we are building a mobile app and we use react-native-web only for storybook, in order to make it accessible for non-developers.  For this usecase we don't want to add any extra code to the app, we would rather add more mocks to rn-web to make storybook not crash in non-mobile environment.

If bundle size is an issue for you, would you accept a PR that mocks PermissionsAndroid to something like this?
```
const promiseMock = () => Promise.resolve(false);
export default {
  PERMISSIONS: {},
  RESULTS: {},
  checkPermission: promiseMock,
  check: promiseMock,
  requestPermission : promiseMock,
  request: promiseMock, 
  requestMultiple: promiseMock,
}
```

necolas:
Ok

sesm:
Updated my PR branch with a smaller mock. This PR is not updating because it's closed, and I can't reopen it.

sesm:
Thanks!

PR #1838, opened 2020-12-06
[fix]: styled-components dynamic benchmark

ElForastero:
It throws an error during dynamic styles benchmark if function is passed into `styles` attribute. It should be a styles map.


ElForastero:
@necolas Will you fix it by yourself in one of the future releases?

Becase it cases an error.

"Styled Components" -> "Update dynamic styles"

https://necolas.github.io/react-native-web/benchmarks/

image

PR #1839, opened 2020-12-06
[feat]: styled-components/native benchmark

ElForastero:
I've implemented a `styled-components/native` benchmark.

Useful to ensure that this approach, despite it's more convenient than using `StyleSheet`, brings a real overhead to web.

image

### DIFF
--- a/packages/benchmarks/src/implementations/styled-components-native/Box.js
+++ b/packages/benchmarks/src/implementations/styled-components-native/Box.js
@@ -1,0 +1,31 @@
+import styled from 'styled-components/native';
+import View from './View';
+
+const getColor = color => {
+  switch (color) {
+    case 0:
+      return '#14171A';
+    case 1:
+      return '#AAB8C2';
+    case 2:
+      return '#E6ECF0';
+    case 3:
+      return '#FFAD1F';
+    case 4:
+      return '#F45D22';
+    case 5:
+      return '#E0245E';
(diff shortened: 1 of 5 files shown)


necolas:
Yeah it's a lot worst. I don't want to add this case to the benchmarks, but feel free to point people at this PR if you need to. Thanks

Sharcoux:
@ElForastero I'd like to run your benchmark with `rn-css` instead of `styled-components/native`, but I'm not sure on how I can do that.
The code should be the same, just replacing `styled-component/native` to `rn-css`

ElForastero:
> @ElForastero I'd like to run your benchmark with `rn-css` instead of `styled-components/native`, but I'm not sure on how I can do that.
> The code should be the same, just replacing `styled-component/native` to `rn-css`

Just stick to the styled-components benchmark. I've copied the entire directory and changed imports and some constructions. That's all :)

Sharcoux:
I guess that I'm supposed to run `yarn install` and then `yarn benchmarks`, but I get a bunch of `Can't resolve 'react-native' in 'react-native-web/packages/benchmarks/src/implementations/react-native-web'` 

I'm not confortable with yarn workspace, maybe I'm missing something.

necolas:
Read the docs https://github.com/necolas/react-native-web/blob/master/.github/CONTRIBUTING.md

PR #1843, opened 2020-12-10
added support for image source object

rick-nice:
Here is fixed PR #1470 for latest react-native-web. Left almost everything as it is.

Please tell, if something is not ok. This is my first contribute experience =)


necolas:
Closing this in favor of #2442 by @kidroca. Thanks

PR #1855, opened 2020-12-23
Add role="switch" in Switch component

intergalacticspacehighway:
Issue - #1856 
Putting role="switch" will make screenreader read it as a Switch rather than a checkbox. (Tested on VoiceOver).

https://www.w3.org/TR/wai-aria-1.1/#switch


necolas:
Thanks!

PR #1858, opened 2020-12-26
fixes - 1857 - ModalContent programmatic focus in useEffect

### LINKED ISSUE #1857: Cannot set programmatic focus to an inner element of a Modal in useEffect.
**The problem**
Cannot set programmatic focus (using .focus()) in an element inside a modal in ModalContent's useEffect hook. 

**How to reproduce**
Demo - https://codesandbox.io/s/react-native-modal-inner-elements-focus-issue-4wwpp
Notes:
- Check ModalContent in App.js. First useEffect doesn't work another works as expected (Have added comments).

Simplified test case: 

Steps to reproduce:
Reproduce here - https://codesandbox.io/s/react-native-modal-inner-elements-focus-issue-4wwpp
1.  Click on Open modal button. Focus won't be set to the required element on useEffect.
2. Run another useEffect which is in setTimeout. This one works

**Expected behavior**
- One should be able to set a focus to any focusable element inside a modal on initial useEffect.

I don't think so, since Modal was introduced in 0.14.

* React Native for Web (version): 0.14.10
* React (version): 16.13.1
* Browser: Chrome (87.0.4280.88 (Official Build) (x86_64))

**Cause**
- ModalAnimation sets display none when isRendering is true. Elements which aren't visible cannot be focused.
**Fix**
- Using opacity instead of display or visibility fixes the issue

intergalacticspacehighway:
Fixes #1857

### DIFF
--- a/packages/react-native-web/src/exports/Modal/ModalAnimation.js
+++ b/packages/react-native-web/src/exports/Modal/ModalAnimation.js
@@ -120,7 +120,7 @@ const styles = StyleSheet.create({
     }
   },
   hidden: {
-    display: 'none'
+    opacity: 0
   }
 });
 


necolas:
@imnotjames do you know if changing from `display:none` to `opacity:0` is going to cause problems? I remember you spending a bit of time working on how the modal is shown/hidden to match the expectations from RN

imnotjames:
I don't think it will cause issues with the RN expectation but it may cause problems with screen readers?

However, given that we stop rendering the element entirely once it's no longer visible.. this should be fine?

PR #1867, opened 2021-01-05
Remove phenomic example from README (unmaintained)

MoOx:
That's totally useless, this project is not maintained.


necolas:
Thanks. I've removed it in a rework of the docs that will be landing soon

PR #1870, opened 2021-01-09
[fix] Update docs use of accessibilityRole "heading"->"header"

danielvinson:
It appears that the accessibility guide refers to the wrong name for this accessibilityRole.  The [react-native docs](https://reactnative.dev/docs/0.60/accessibility#accessibilityrole-android-ios) use "header".

Please let me know if I am mistaken!


necolas:
See https://github.com/necolas/react-native-web/pull/1350

PR #1874, opened 2021-01-13
Fix getting the sheet of the styleResolver

mathieudutour:
Without a getter, an outdated version of `sheet` is returned after `getStyleSheet` is called
as shown in this small reproducible example:

```js
function createObject() {
  let ref
  let i = 1
  function init() {
    ref = {
      id: i++
    }
  }
  init()
  return {
    ref,
    getContent() {
      const content = ref.id
      init()
      return content
    }
  }
}
var object = createObject()
object.ref // {id: 1}
object.getContent() // 1

object.ref // {id: 1}
object.getContent() // 2
```


necolas:
Please can you add a regression test?

mathieudutour:
done :+1:

necolas:
Thanks!

Issue #1881, opened 2021-01-22
classList is not forwarded to components


wSedlacek:
**The problem**
The `classList` prop is not forwarded to components despite being listed in the `forwardedProps`.

**How to reproduce**
Apply a `classList` to `Text`, `View`, etc.
Simplified test case:
```ts
<View classList="something" />
```

**Expected behavior**
Given that `classList` is listed as `true` in the forwardProps I would expect it not to be overwritten but instead merged with additional classes that are applied.

**Environment (include versions). Did this work in previous versions? Not sure? This is the first time I am trying to use it.

* React Native for Web (version): 0.14.10
* React (version): 16.13.1
* Browser: Chrome

**Additional context**
I believe it is important to be able to apply additional classes to components in the event that a class created by React Native Web has a limitation, such as not supporting `@media` queries.

The issue is caused by overwriting the value of `classList` in `supportedProps` rather than merging it.
https://github.com/necolas/react-native-web/blob/6739ab95cfdfc1bcb6ffc6989a1afb6ab373521e/packages/react-native-web/src/exports/View/index.js#L132

necolas:
This has been answered so many times before. You can search the issues for previous discussions

wSedlacek:
Thank you for the direction! Looks like I will be using `dataSet` or `setNativeProps`.

PR #1882, opened 2021-01-22
[fix] Fordward classList to DOM elements

wSedlacek:
This changes it so that the `classList` generated in `View`, `Text` and `TextInput` are appended to any existing classList from the props.

Close: #1881


necolas:
Thanks but we don't want this to happen.

laszhiai:
why

steida:
Because RNfW is supposed to be a better abstraction than browser DOM. You can hack it, but the proper way is to not use classes because code should be platform-independent.

Issue #1785, opened 2020-10-27
[Add] Native Modules and NativeUIModules support


muskeinsingh:
**Is your feature request related to a problem? Please describe.**
React Native comes out with a first class support for Native Modules and Native UI Modules and that attempts to solve a LOT of use cases and bridge(pun intended) the gap between Native and the React world. I have a crude implementation of how these could work with the current architecture of RN(registering singleton instances of those modules pre-loading) within the web world too. 

**Describe a solution you'd like**
Basically, a way to expose NativeModules and NativeUIModules capability to web as well. 

**Describe alternatives you've considered**

**Additional context**
Not sure if this is being already worked on or if this doesn't align well with the future roadmap of the library but just in case it does, do let me know and I can help carve out a rusty PR with the same.

necolas:
Some of that stuff might be changing in RN with Fabric. @EvanBacon has worked on a universal module format for Expo; he might have some input.

muskeinsingh:
@EvanBacon can you share your insights here? could be useful to help decide the way forward.

necolas:
Didn't you already do this at Flipkart? You could share a PR of that work to start a discussion

muskeinsingh:
sure, I can share that. will ready a PR 👍

rchavezj:
I'm very interested in native modules used for this hybrid library. So at this moment does react-native-web support ios-swift and java-android inside react-native?

rchavezj:
I'm not completely sure what "Mocked. Missing ability to load native modules." means seeing the status of this feature has '(✓)'

necolas:
It means the app won't crash from importing the `NativeModules` export. You shouldn't expect native code to run on the web

necolas:
Any follow up here @muskeinsingh?

necolas:
I reached out over DM to Flipkart eng management about this and other things, but they didn't follow through. Closing due to lack of follow-up.

PR #1901, opened 2021-02-08
feat: enabling ReactPackage as an export from React Native fo…

muskeinsingh:
…r Web to allow developers to register instances of their native modules to then be available for use across the app under the Native Module import

this change is, however, far from complete and just introduces one layer of change to get an understanding from the community as to how they view this and then take it forward from there. similar work can be done to get instances for components and run requireNativeComponent as well.

initial draft for [adding native module support to rnw](https://github.com/necolas/react-native-web/issues/1785)

### DIFF
--- a/packages/babel-plugin-react-native-web/src/moduleMap.js
+++ b/packages/babel-plugin-react-native-web/src/moduleMap.js
@@ -37,6 +37,7 @@ module.exports = {
   Platform: true,
   Pressable: true,
   ProgressBar: true,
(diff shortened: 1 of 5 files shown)


muskeinsingh:
@necolas would appreciate your feedback here.

muskeinsingh:
closing as this is incomplete and in hindsight, I think supporting lazy loading of packages should definitely be a part of the solution. will work and share updates!

PR #1902, opened 2021-02-08
feat: adding husky to use precommit, prerelease and postrelease hooks

muskeinsingh:
`git commit` throws a warning on the repo currently that running `precommit` and other hooks directly under scripts would be deprecated soon. this change is to remove that warning and still maintain backward compatibility.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,10 +17,7 @@
     "jest": "jest --config ./scripts/jest/config.js",
     "lint": "yarn lint:report --fix",
     "lint:report": "eslint packages scripts",
-    "precommit": "lint-staged",
-    "prerelease": "yarn test && yarn compile && yarn compile:commonjs",
     "release": "node ./scripts/release/publish.js",
-    "postrelease": "yarn benchmarks:release && yarn docs:release",
     "test": "yarn flow && yarn lint:report && yarn jest --runInBand"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -62,6 +59,13 @@
     "react-dom": "^16.10.2",
     "react-test-renderer": "^16.10.2"
   },
+  "husky": {
+    "hooks": {
+      "prerelease": "yarn test && yarn compile && yarn compile:commonjs",
```
necolas, line 64: This is not a hook

package.json
```diff
@@ -62,6 +59,13 @@
...
+      "precommit": "lint-staged",
+      "postrelease": "yarn benchmarks:release && yarn docs:release"
```
necolas, line 66: This is not a hook

necolas (commented):
Hmm, pulling this PR and running `yarn` seems to have broken my local husky install. The pre-commit hook doesn't run at all anymore, including after I checked out master and reinstalled the packages.

Edit: had to run `git config --unset core.hookspath`

package.json
muskeinsingh: resolved

package.json
muskeinsingh: resolved

necolas:
Closing as I fixed this while updating deps locally

PR #1915, opened 2021-02-23
Fix SectionList not considering viewOffset param on scrollToLocation

horatiuvlad:


horatiuvlad:
Seems like it was fixed with #1984

PR #1937, opened 2021-03-13
Pass target unmodified

petrbela:
I'm not sure why the target attribute is handled in this special way but I think it's causing confusion since it's different than the normal browser behavior. Moreover, if you pass `hrefAttrs={{ target: '_blank' }}`, it doesn't pass the `if` and isn't set at all. Since the browser allows any string (which would use an existing window if opened repeatedly with the same target name), or the special `_self`, `_parent`, `_root`, and `_blank`, I think it's best to just pass it unmodified.


petrbela:
Related, would it make sense to support [all attributes](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/a#attributes)? I could make a separate PR for that.

PR #1938, opened 2021-03-15
Add accessibilityCurrent prop

intergalacticspacehighway:
This PR adds accessibilityCurrent attribute that is similar to [aria-current in ARIA spec](https://www.w3.org/TR/wai-aria-1.2/#aria-current)


necolas:
Thanks!

Issue #1935, opened 2021-03-12
`setNativeProps` ignores previous style changes


piaskowyk:
**The problem**

Update `react-native-web` from `0.12.3` to `0.13.0` broke `setNativeProps` function, and next call `setNativeProps` override previously applayed style. 

For example. I have component:
```js
<View
  ref={viewRef}
  style={{ width: 100, height: 100, background: "#333" }}
/>
```
I want to change the style of the component. 
Operation:
```js
viewRef.setNativeProps({ style: { width: 200 } });
viewRef.setNativeProps({ style: { height: 200 } })
```
Result in version `0.12.3`:
`view style = { width: 200, height: 200 }`

Result in version `0.13.0` (and newer):
`view style = { width: 100, height: 200 }`

**How to reproduce**

Just run both apps

Example in version `0.12.3`: https://codesandbox.io/s/rnw-12-y11uy
Example in version `0.13.0`: https://codesandbox.io/s/rnw-13-sbuto

Steps to reproduce:
1. Run both examples
2. Click a couple of times on the button
3. Observe differences

**Expected behavior**

I expect behavior like was in version `0.12.3`

**Environment (include versions). Did this work in previous versions?**

* React Native for Web (version): 0.13.0
* React (version): 17.01
* Browser: 17.01

**Additional context**

In `react-native-reanimated` we use `setNativeProps` to apply changed style during an animation. But after upgrade `react-native-web`, we have strange issues on the web version if one animation is longer than another component back to the initial style. 

Example issue:

- https://github.com/software-mansion/react-native-reanimated/issues/1804 
- https://github.com/software-mansion/react-native-reanimated/issues/1450

I made a little investigation and confirmed that `setNativeProps` doesn't change the initial styles of component: https://github.com/necolas/react-native-web/blob/c47bec7b93d6a3b7c31bbc8bb2e4acd117b79bfc/packages/react-native-web/src/modules/usePlatformMethods/index.js#L29-L40

I can prepare PR to apply changes to the initial style, but I'm not sure how this can affect other components.

Thanks for any help!

necolas:
Sure, you can make a PR. Just be aware that the behavior of setNativeProps is undefined in RN and complicated to implement on web.

necolas:
This patch is available in 0.15.2

necolas:
c0abdbfaaead4b6fd999cb767cfa1de1cc046564

nandorojo:
Thank you both for the quick fix here.

[...]

PR #1939, opened 2021-03-15
Fix setNativeProps

piaskowyk:
`setNativeProps` method was ignoring previously changed style.
Detailed description in issue: https://github.com/necolas/react-native-web/issues/1935

Fixes: https://github.com/necolas/react-native-web/issues/1935


nandorojo:
Thank you @piaskowyk!

PR #1940, opened 2021-03-15
Update Flow to 0.146; Implement types-first via codemod

comp615:
**Problem**
The version of flow the project is currently using is from Oct 2019. Updating can provide better or more accurate flow types. Originally, I set out to find a way to directly use flow types from RNW in projects, however, this portion of the work should still have value.

**Solution**
Because this is obviously a massive change, I wanted to highlight the process:
- Updated flow package dep to latest
- Added `munge_underscores` which will not require exported types for anything "internal" using the underscore prefix convention
- Ran the codemod [specified here ](https://flow.org/en/docs/cli/annotate-exports/). Which tries to auto-annotate all public types from files
- Tweaked the type of View using the types we added at Twitter. This makes the ref type an HTMLElement with native methods mixed in. **This should add type-safety for refs.**
- Manually cleaned up or fixed places the codemod couldn't. Specifically added types for default export forwardRefs.
- Updated the `vendor` modules that were broken from react-native master branch. React-native already upgraded so the flow types have since been added to those.

NOTES:
- I did not manually review every type produced by the codemod. Some could probably be phrased better, but it compiles and shouldn't affect behavior.
- There may be functional changes to the vendor modules grabbed from React-Native. I didn't need to change any calls to those modules so I assumed it's OK, but testing should be focused there.
- I tried to avoid all possible functional changes, there were a few in useEvent for instance where I needed to null check before calling a function, which seemed valid based on types and functionality.


package.json
```diff
@@ -21,7 +21,7 @@
     "jest": "jest --config ./scripts/jest/config.js",
     "lint": "yarn lint:report --fix",
     "lint:report": "eslint packages scripts",
-    "prerelease": "yarn test && yarn compile && yarn compile:commonjs",
```
necolas, line 24: Was this accidentally removed?

packages/react-native-web/src/exports/AppRegistry/AppContainer.js
```diff
@@ -9,6 +9,7 @@
  */
 
 import type { ComponentType, Context } from 'react';
+import type { Node } from 'React';
```
necolas, line 12: This should be using lowercase `react` and combined with the import above

packages/react-native-web/src/exports/CheckBox/index.js
```diff
@@ -26,49 +26,51 @@ type CheckBoxProps = {
   value?: boolean
 };
 
-const CheckBox = forwardRef<CheckBoxProps, *>((props, forwardedRef) => {
-  const { color, disabled, onChange, onValueChange, style, value, ...other } = props;
+const CheckBox: React$AbstractComponent<CheckBoxProps, mixed> = forwardRef<CheckBoxProps, *>(
```
necolas, line 29: Why does this component use `mixed` instead of an explicit type?

packages/react-native-web/src/exports/DeviceInfo/index.js
```diff
@@ -34,7 +34,7 @@ const DeviceInfo = {
     }
   },
 
-  get locale() {
+  get locale(): any | void {
```
necolas, line 37: Shouldn't `any` be `string` in this case?

packages/react-native-web/src/exports/DeviceInfo/index.js
```diff
@@ -44,11 +44,11 @@ const DeviceInfo = {
     }
   },
 
-  get totalMemory() {
+  get totalMemory(): any | void {
```
necolas, line 47: Should be `number` not `any`

packages/react-native-web/src/exports/DeviceInfo/index.js
```diff
@@ -44,11 +44,11 @@ const DeviceInfo = {
...
     return canUseDOM ? window.navigator.deviceMemory : undefined;
   },
 
-  get userAgent() {
+  get userAgent(): any | string {
```
necolas, line 51: No `any`

packages/react-native-web/src/exports/Image/index.js
```diff
@@ -19,7 +19,15 @@ import PixelRatio from '../PixelRatio';
...
 import TextAncestorContext from '../Text/TextAncestorContext';
 import View from '../View';
-import React, { forwardRef, useContext, useEffect, useRef, useState } from 'react';
+import React, {
+  forwardRef,
+  useContext,
+  useEffect,
+  useRef,
+  useState,
+  type AbstractComponent,
```
necolas, line 28: Move type imports to a separate import alongside other import on L11

packages/react-native-web/src/exports/KeyboardAvoidingView/index.js
```diff
@@ -10,6 +10,7 @@
 
 import type { LayoutEvent, LayoutValue } from '../../types';
 import type { ViewProps } from '../View';
+import type { Node } from 'React';
```
necolas, line 13: lowercase `react`

packages/react-native-web/src/exports/Modal/ModalAnimation.js
```diff
@@ -32,7 +32,7 @@ export type ModalAnimationProps = {|
   visible?: ?boolean
 |};
 
-function ModalAnimation(props: ModalAnimationProps) {
+function ModalAnimation(props: ModalAnimationProps): any | null {
```
necolas, line 35: Why `any | null`?

packages/react-native-web/src/exports/Modal/ModalContent.js
```diff
@@ -8,7 +8,13 @@
  * @flow
  */
 
-import React, { forwardRef, useMemo, useEffect } from 'react';
+import React, {
+  forwardRef,
+  useMemo,
+  useEffect,
+  type AbstractComponent,
```
necolas, line 15: Separate types import

packages/react-native-web/src/exports/Modal/ModalFocusTrap.js
```diff
@@ -8,6 +8,7 @@
  * @flow
  */
 
+import type { Node as $IMPORTED_TYPE$_Node } from 'React';
```
necolas, line 11: what is going on here?

packages/react-native-web/src/exports/Modal/ModalPortal.js
```diff
@@ -16,7 +16,7 @@ export type ModalPortalProps = {|
   children: any
 |};
 
-function ModalPortal(props: ModalPortalProps) {
+function ModalPortal(props: ModalPortalProps): any | null {
```
necolas, line 19: `any | null` is a pointless type

packages/react-native-web/src/exports/Modal/index.js
```diff
@@ -8,7 +8,15 @@
...
 
-import React, { forwardRef, useCallback, useMemo, useEffect, useState } from 'react';
+import React, {
+  forwardRef,
+  useCallback,
+  useMemo,
+  useEffect,
+  useState,
+  type ElementRef,
+  type AbstractComponent
```
necolas, line 18: Separate `type` import

packages/react-native-web/src/exports/Picker/index.js
```diff
@@ -14,65 +14,68 @@ import createElement from '../createElement';
...
-    style,
-    testID,
-    /* eslint-disable */
-    itemStyle,
-    mode,
-    prompt,
-    /* eslint-enable */
-    ...other
-  } = props;
+const Picker: React$AbstractComponent<PickerProps, mixed> = forwardRef<PickerProps, *>(
```
necolas, line 33: Why `mixed` here? What's the value of not using `mixed` on all the other components?

packages/react-native-web/src/exports/ProgressBar/index.js
```diff
@@ -12,7 +12,13 @@ import type { ViewProps } from '../View';
 
 import StyleSheet from '../StyleSheet';
 import View from '../View';
-import React, { forwardRef, useEffect, useRef } from 'react';
+import React, {
+  forwardRef,
+  useEffect,
+  useRef,
+  type AbstractComponent,
```
necolas, line 19: types import

packages/react-native-web/src/exports/RefreshControl/index.js
```diff
@@ -10,6 +10,7 @@
 
 import type { ColorValue } from '../../types';
 import type { ViewProps } from '../View';
+import type { Node } from 'React';
```
necolas, line 13: `react`

packages/react-native-web/src/exports/ScrollView/index.js
```diff
@@ -17,21 +17,21 @@ import ScrollResponder from '../../modules/ScrollResponder';
 import ScrollViewBase from './ScrollViewBase';
 import StyleSheet from '../StyleSheet';
 import View from '../View';
-import React from 'react';
+import * as React from 'react';
 
-type ScrollViewProps = {
-  ...ViewProps,
+type ScrollViewProps = {|
+  ...React.ElementConfig<typeof ScrollViewBase>,
```
necolas, line 23: Types should be imported separately rather than assuming they are available on the `React` object

packages/react-native-web/src/exports/StyleSheet/StyleSheet.js
```diff
@@ -18,12 +18,12 @@ const absoluteFillObject = {
...
   bottom: 0
 };
-const absoluteFill = ReactNativePropRegistry.register(absoluteFillObject);
+const absoluteFill: number = ReactNativePropRegistry.register(absoluteFillObject);
 
 const StyleSheet = {
   absoluteFill,
   absoluteFillObject,
-  compose(style1: any, style2: any) {
+  compose(style1: any, style2: any): any | Array<any> {
```
necolas, line 26: Once you have `any` the second option doesn't matter

packages/react-native-web/src/exports/StyleSheet/initialRules.js
```diff
@@ -13,9 +13,9 @@ const resets = [
   'body{margin:0;}',
   // minimal form pseudo-element reset
   'button::-moz-focus-inner,input::-moz-focus-inner{border:0;padding:0;}',
-  'input::-webkit-inner-spin-button,input::-webkit-outer-spin-button,' +
+  ('input::-webkit-inner-spin-button,input::-webkit-outer-spin-button,' +
```
necolas, line 16: I think we can remove this change

packages/react-native-web/src/exports/StyleSheet/normalizeValueWithProperty.js
```diff
@@ -23,7 +23,10 @@ const colorProps = {
   textShadowColor: true
 };
 
-export default function normalizeValueWithProperty(value: any, property?: ?string) {
+export default function normalizeValueWithProperty(
+  value: any,
+  property?: ?string
+): any | void | string {
```
necolas, line 29: I don't get these type signatures

packages/react-native-web/src/exports/Text/index.js
```diff
@@ -39,116 +39,118 @@ const forwardPropsList = {
...
-    onSelectionChangeShouldSetResponder,
-    onSelectionChangeShouldSetResponderCapture,
-    onStartShouldSetResponder,
-    onStartShouldSetResponderCapture
-  });
-
-  function handleClick(e) {
-    if (onClick != null) {
-      onClick(e);
+const Text: React$AbstractComponent<TextProps, mixed> = forwardRef<TextProps, *>(
```
necolas, line 42: `Text` should probably be typed more like `View`.

And rather than changing the indentation of all these files we should do something like RN https://github.com/facebook/react-native/blob/9be3356e02845a3a8194226a34faa4cb6f7295f1/Libraries/Components/ActivityIndicator/ActivityIndicator.js#L188-L192

packages/react-native-web/src/exports/Touchable/index.js
```diff
@@ -9,6 +9,7 @@
  */
 
 'use strict';
+import type { Node } from 'React';
```
necolas, line 12: `react`

packages/react-native-web/src/exports/View/index.js
```diff
@@ -39,9 +39,40 @@ const forwardPropsList = {
   pointerEvents: true
 };
 
+type RNW$MeasureLayoutCallback = (
```
necolas, line 42: Why do these types all have a RNW prefix?

packages/react-native-web/src/exports/View/index.js
```diff
@@ -148,6 +179,6 @@ const styles = StyleSheet.create({
   }
 });
 
-export type { ViewProps };
+export type { ViewProps, RNW$NativeMethodsMixin };
```
necolas, line 182: Why is this type exported? RN doesn't have a NativeMethodsMixin anymore either

packages/react-native-web/src/exports/YellowBox/index.js
```diff
@@ -8,10 +8,11 @@
  * @flow
  */
 
+import type { Node } from 'React';
```
necolas, line 11: `react`

packages/react-native-web/src/modules/AccessibilityUtil/isDisabled.js
```diff
@@ -7,7 +7,7 @@
  * @flow
  */
 
-const isDisabled = (props: Object) =>
+const isDisabled = (props: Object): any =>
```
necolas, line 10: any -> boolean

packages/react-native-web/src/modules/AccessibilityUtil/propsToAccessibilityComponent.js
```diff
@@ -32,7 +32,7 @@ const roleComponents = {
 
 const emptyObject = {};
 
-const propsToAccessibilityComponent = (props: Object = emptyObject) => {
+const propsToAccessibilityComponent = (props: Object = emptyObject): any | void | string => {
```
necolas, line 35: weird type

packages/react-native-web/src/modules/AccessibilityUtil/propsToAriaRole.js
```diff
@@ -22,7 +22,7 @@ const accessibilityRoleToWebRole = {
   text: null
 };
 
-const propsToAriaRole = ({ accessibilityRole }: Object) => {
+const propsToAriaRole = ({ accessibilityRole }: Object): any | void => {
```
necolas, line 25: any -> string

packages/react-native-web/src/modules/UnimplementedView/index.js
```diff
@@ -7,6 +7,7 @@
  * @flow
  */
 
+import type { Node } from 'React';
```
necolas, line 10: `react`

packages/react-native-web/src/modules/useHover/index.js
```diff
@@ -81,8 +81,8 @@ export default function useHover(targetRef: any, config: HoverEventsConfig): voi
           onHoverChange(false);
         }
         // Remove the listeners once finished.
-        addMoveListener(target, null);
-        addLeaveListener(target, null);
+        addMoveListener && addMoveListener(target, null);
```
necolas, line 84: We shouldnt need to guard these calls, and guards should always use if != null statements instead

packages/react-native-web/src/modules/useResponderEvents/utils.js
```diff
@@ -78,7 +78,7 @@ export function getResponderPaths(
 /**
  * Walk the paths and find the first common ancestor
  */
-export function getLowestCommonAncestor(pathA: Array<any>, pathB: Array<any>) {
+export function getLowestCommonAncestor(pathA: Array<any>, pathB: Array<any>): any | null {
```
necolas, line 81: bad type

packages/react-native-web/src/vendor/hash/index.js
```diff
@@ -11,6 +11,8 @@
  * @param {string} str ASCII only
  * @param {number} seed Positive integer only
  * @return {number} 32-bit positive integer hash
+ *
+ * @flow
```
necolas, line 15: Is it necessary to flow type this file?

packages/react-native-web/src/vendor/react-native/Animated/AnimatedEvent.js
```diff
@@ -4,7 +4,7 @@
  * This source code is licensed under the MIT license found in the
  * LICENSE file in the root directory of this source tree.
  *
- * @noflow
+ * @flow
```
necolas, line 7: Were the Animated files updated from RN or just typed the old ones?

packages/react-native-web/src/vendor/react-native/VirtualizedList/index.js
```diff
@@ -1865,4 +2091,4 @@ const styles = StyleSheet.create({
   },
 });
 
-export default VirtualizedList;
+module.exports = VirtualizedList;
```
necolas, line 2094: Need to keep ESM exports

packages/react-native-web/src/vendor/react-native/emitter/_EventEmitter.js
```diff
@@ -0,0 +1,197 @@
...
+        // its listener matches the listener in method parameters
+        if (subscription && subscription.listener === listener) {
+          subscription.remove();
+        }
+      }
+    }
+  }
+}
+
+module.exports = EventEmitter;
```
necolas, line 197: ESM export

packages/react-native-web/src/vendor/react-native/emitter/_EventSubscription.js
```diff
@@ -0,0 +1,44 @@
...
+
+  /**
+   * Removes this subscription from the subscriber that controls it.
+   */
+  remove(): void {
+    this.subscriber.removeSubscription(this);
+  }
+}
+
+module.exports = _EventSubscription;
```
necolas, line 44: ESM

packages/react-native-web/src/vendor/react-native/emitter/_EventSubscriptionVendor.js
```diff
@@ -93,9 +98,11 @@ class EventSubscriptionVendor {
...
-  getSubscriptionsForType(eventType: string): ?[EventSubscription] {
+  getSubscriptionsForType<K: $Keys<EventDefinitions>>(
+    eventType: K,
+  ): ?Array<EventSubscription<EventDefinitions, K>> {
     return this._subscriptionsForType[eventType];
   }
 }
 
-export default EventSubscriptionVendor;
+module.exports = EventSubscriptionVendor;
```
necolas, line 108: ESM

packages/react-native-web/src/vendor/react-native/emitter/__flowtests__/EventEmitter-flowtest.js
```diff
@@ -0,0 +1,127 @@
+/**
```
necolas, line 1: What is this file for?

necolas (changes requested):
Thanks for looking into this.

I do think this change needs to be broken down into smaller pieces.

1. We could land the changes in `vendor` with the files using "noflow"
2. We could upgrade flow, remove "noflow" from the `vendor` files, and update the other types.

I think the LOC changes will come down a lot too once the abstract type mess in the components is moved below the component definition as is done in RN. Then the indentations of the files won't be changed.

What's after this? How do you anticipate getting the flow types available from the npm package?

package.json
comp615: No, it was updated to remove the third clause. `yarn compile` already runs all such `compile:` commands. So it seems like a vestige of yore

packages/react-native-web/src/exports/CheckBox/index.js
comp615: Mixed is the equivalent of *. Updated these to use explicit types.

packages/react-native-web/src/vendor/react-native/emitter/__flowtests__/EventEmitter-flowtest.js
comp615: Copy pasta, will remove tests.

packages/react-native-web/src/exports/DeviceInfo/index.js
comp615: Yeah these any types are from the codemod, and it turns out flow doesn't have navigator fully typed

packages/react-native-web/src/exports/Modal/ModalAnimation.js
comp615: `createElement` isn't typed so it's returning any

packages/react-native-web/src/exports/Modal/ModalFocusTrap.js
comp615: Dodging use of HTML "Node" already existing

packages/react-native-web/src/exports/Modal/ModalPortal.js
comp615: Need to look at this one more, this may need some react-dom types imported as createPortal is currently any

packages/react-native-web/src/exports/Text/index.js
comp615: Fixed forwardRef. But leaving this open for indentation discussion as I just left those to lint

packages/react-native-web/src/exports/View/index.js
comp615: Removed prefix and moved around to match RNW style more than Twitter style

packages/react-native-web/src/modules/useHover/index.js
comp615: This was something I tried to note, but my relative lack of hook experience may show here. It does actually look from a JS perspective like it could be null (`useEvent` appears to nullify the listener and return it on cleanup, so technically `useEvent` can return null). I know in practice that's unlikely to happen though

packages/react-native-web/src/exports/Modal/ModalAnimation.js
comp615: (I'm going to try to throw a very loosey / any type on createElement)

packages/react-native-web/src/exports/StyleSheet/initialRules.js
comp615: Because it's exported, it wants this or we can ignore it. Something with the multi-line thing likely

packages/react-native-web/src/modules/AccessibilityUtil/isDisabled.js
comp615: Also removed deprecated prop

packages/react-native-web/src/vendor/hash/index.js
comp615: It helped flow another export in another file I believe (always returns string)

packages/react-native-web/src/vendor/react-native/Animated/AnimatedEvent.js
comp615: Typed the old ones I believe

packages/react-native-web/src/exports/Modal/ModalPortal.js
comp615: Imported ReactDOM types

comp615:
> I do think this change needs to be broken down into smaller pieces.
I spent a few hours today trying to break this up, and there's small bits I was able to do (i.e. moving around flow-typed), but the majority of the work is kind of messily intermingled. For instance, I can't pull in the new vendor stuff from RN without updating flow because it seems to use some new flow annotations, so we could @noflow that, but then I'm not as confident it drops in right)...and then updating flow reveals errors in the current setup (for instance Picker imports types that don't exist) or it gets mad about importing any types. So then you can disable those rules.

It's possible I just need to scrap it all and start over in chunks now somehow. So maybe take another stab next weekend, but it's starting to suck time.

> What's after this? How do you anticipate getting the flow types available from the npm package?

I tried to call this out in the summary. I don't necessarily think this gets any closer or further to being able to export types. In my play, I found that using a host project with a newer flow was throwing the same errors, but it's somewhat orthogonal (the copy-sources approach basically relies on the host projects flow processor).

All that is to say...what's even the point of types if we can't export them :(

PR #1941, opened 2021-03-16
Add support for "menuitemcheckbox" and "menuitemradio" roles

VinceMalone:
### Issue

I noticed `accessibilityRole="menuitem"` provides _button_ like behavior — e.g. becomes focusable, is clicked when focused and <kbd>space</kbd> is pressed, ignores click events when disabled — but the same treatment isn't applied when the role is set to `menuitemcheckbox` or `menuitemradio`. Given the example in a [W3C wiki for ARIA menus](https://www.w3.org/WAI/GL/wiki/Using_ARIA_menus), the `menuitemcheckbox` and `menuitemradio` roles behave similarly to `menuitem`; i.e. as a _button_.

### Solution

Anywhere `menuitem` is given special treatment, do the same for `menuitemcheckbox` and `menuitemradio`.

### DIFF
--- a/packages/react-native-web/src/modules/AccessibilityUtil/buttonLikeRoles.js
+++ b/packages/react-native-web/src/modules/AccessibilityUtil/buttonLikeRoles.js
@@ -13,7 +13,13 @@ const buttonLikeRoles: { [string]: boolean } = {
   // ARIA menuitem responds to Enter/Space like a button. Spec requires AT to
   // ignore ARIA roles of any children.
   // https://www.w3.org/WAI/GL/wiki/Using_ARIA_menus
-  menuitem: true
+  menuitem: true,
+  menuitemcheckbox: true,
+  menuitemradio: true
 };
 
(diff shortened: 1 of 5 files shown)


necolas:
I think we might be better off removing the special treatment for `menuitem`. These roles are not meant to be in the tab flow unless the element type demands it. The recommended accessibility pattern ("roving tabindex") is that tab enters the menu at the active item, and items are navigated using arrow keys which moves the tabindex. In RN the active item would need the `focusable` prop, and after the special treatment for `menuitem` is removed you'd also have to manage `onClick` and `onKeyDown` to apply the accessible interaction behavior.

necolas:
What I proposed above would be a breaking change. @comp615 let me know your thoughts as it applies to Twitter

VinceMalone:
That's fair — I agree that ultimately removing the special treatment of `menuitem` is ideal. Chatted with @comp615 too and the removal doesn't appear to be too disruptive on our end either 👍

Issue #1899, opened 2021-02-07
Revisit implementation of `accessibilityRole="button"`


necolas:
**Is your feature request related to a problem? Please describe.**

Revisit mapping `accessibilityRole="button"` to `button` element.

The current approach uses ARIA roles to achieve similar results due to [historical problems in Firefox with flexbox children](https://github.com/philipwalton/flexbugs#flexbug-9) of a button. But this was fixed in Firefox 63 and the [current ESR version is 78](https://en.wikipedia.org/wiki/Firefox_version_history#Firefox_78.7esr), so we may be able to migrate away from this special case and map "button" to an element like we do for several other roles. (FB web still supports FF versions prior to 63 but no one performs visual tests that far back so I don't think we need to be too concerned about it.)

This shouldn't produce any savings in bytes since the code used to implement ARIA-backed buttons is also used by the unstable `createElement` API and for other ARIA roles that should have similar behavior to buttons (i.e., respond to user interaction).

doumart:
I was gonna file the same issue. I use react testing library and i use accessibility to test the content of my page. To have a real button tag removes a lot of friction and is more near the reality of pure web development.

intergalacticspacehighway:
@necolas What should be the case for `<View accessibilityRole="button" />`? 
Should this renderd as <div role="button" /> or <button />. 
Got this as a failing test case while implementing the above change.

georgii-ivanov:
@intergalacticspacehighway what's the problem with `<View accessibilityRole="button" />` cuz at this point you'll get button tag and you can put another View/Text element into it as HTML allows you to put something else besides text node.

I think that would be a useful improvement to HTML semantics. Also probably useful for SEO (not sure about that tho)

jgornick:
We ran into this issue because our a11y rules require semantic HTML rather than using a11y attributes (e.g. `aria-*`). In our case we needed to produce a `<button>` when using the `<Button>` component. However, we learned that once that goes through `<TouchableOpacity>` to `<View>` to `createElement` and finally `propsToAccessibilityComponent`, we noticed that it wasn't mapping a role of `button` to a `<button>` a11y component. 

So, we started with @intergalacticspacehighway's PR #1944 implementation, but because not everyone shares our same a11y rules and may not want to map roles to semantic HTML, we short-circuited the `<Button>` component rendering to use `createElement('button', ...)`. That way others could still use a `<View>` (`<div>`) with an a11y role of `button`.

[...]

PR #1944, opened 2021-03-17
revisit accessibilityRole=button to button mapping

intergalacticspacehighway:
[Issue link](https://github.com/necolas/react-native-web/issues/1899)

- Tested example and updated test cases.
- Verified reset styles, tabIndex, and disabled props.


louis-burrows (approved):

jerrywithaz:
@necolas @louis-burrows if the merge conflicts in this get fixed, would you all accept this pr? It would great if buttons were actually rendered instead of div's

s77rt:
This may not work as you'd expect, in our case we had `<div>`s rendered inside `<button>` which is semantically incorrect.

PR #1953, opened 2021-03-20
Move flow-typed files into default directory

comp615:
**Problem**
`Flow-typed` is often used to maintain and pull in libdefs from other projects or dependencies. By [default, flow and the flow-typed command look in the flow-typed dir](https://flow.org/en/docs/config/libs/). However currently RNW uses the `types` dir

**Solution**
Standardize a little bit and move some things around. Update to use flow-typed, and exempt this dir from prettier and eslint since they should be vendored files.

Also updated the eslint-flow config, and removed superflous command from prerelease


necolas:
Merge into `flow-and-vendor` in 494dbf3931d1e54763f5b4c3ed228986226ca85a

PR #1955, opened 2021-03-20
Flow (2/N): Update flow to 0.142

comp615:
Goal is to update flow and implements types first, fixing issues as I go. This is an attempt at rewriting #1940 via stacked reviews.

In this review, we assume part 1, and we update the vendored items which needed Flow resolution. This was done by:
- downloading react-native, and copy pasting the files to RNW.
- Updating module.exports to export default
- Updating Imports to reference RNW

This branch should be shippable in conjunction with #1954 

NOTE: I can't figure out how to stack reviews on GH, so I'll try and continue this later if this path looks OK. Rough structure or plan (although I haven't coded this far out so subject to change).
1. Update
2. Vendors
3. Vendors pt. 2
4. ForwardRefs and View Ref Typing
5. Codemod migration + turn option on
6. Update ViewPropTypes to be strict


necolas (approved):
Thanks. Looks good to me. I can see the updated VirtualizedList implementation should avoid issues like #1959 too.

We should probably make a minor release just for this vendor code and types update.

comp615:
Yeah, that seems good as I don't know what implications the vendor changes might have...so could also merge it into a "next" branch. I can keep piling on top of this after that.

necolas:
This was merged in the following branch https://github.com/necolas/react-native-web/tree/flow-and-vendor

You can submit future PRs for the types/vendor work against that branch.

Thank you!

PR #1971, opened 2021-04-01
Flow 0.142 (4/N): Update View ref type; Fix forwardRefs in other components

comp615:
- Defines a `NativeMethodsMixin` type that reflects what the useNativeMethods mixin adds.
- Sets a instance type for View (and a few other foundational components that use the mixin)
- Updates all forwardRef usages to refer to and have instance types set, and to be well-formed exports

In general, that means converting:
```
const component = forwardRef<Props, *>(...
```
to
```
const component: AbstractComponent<Props, ElementRef<View>> = forwardRef(...
```
Basically moving the signature to the export and letting the generic on forwardRef get inferred that way.

NOTE: The changes can look larger than they are because prettier wanted to tab some things

### DIFF
--- a/packages/react-native-web/src/exports/ActivityIndicator/index.js
+++ b/packages/react-native-web/src/exports/ActivityIndicator/index.js
@@ -8,6 +8,7 @@
  * @flow
  */
 
+import type { ElementRef, AbstractComponent } from 'react';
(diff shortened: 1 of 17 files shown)


necolas:
Merged into flow-and-vendor

PR #1972, opened 2021-04-01
Flow 0.142 (5/N): Codemod export-types on src folder

comp615:
Run `flow codemod annotate-exports` on the src folder of RNW and manually audit all these types to make sure they are reasonable.

Many places still have any-types, so let those as is. In a few places, I was able to make flow happy by using `navigator` directly instead of `window.navigator`. I'm not sure why this is a thing, but I believe they are interchangeable and flow has navigator typed but not as window.navigator. So that helped a bit.

Other than the above note, no functional changes...just a wee bit messier at times


necolas:
Merged into flow-and-vendor. Thanks!

PR #1980, opened 2021-04-08
Flow 0.142 (6/7): Enable types-first by default; fix remaining issues

comp615:
Get to our goal of enabling the new Flow architecture, types-first. The newer versions of flow remove the option at all, so instead of setting to true, I just updated flow again and cleaned up the config.

Fixed a few lingering issues missed by code-mods or raised by the new system (it catches a bit more).

NOTE: If you can take a look at useEffect that would be great. I didn't want to touch functionality and I'm not very good with hooks, but both flow and ESLint seem upset about how it's working, so maybe a quick tweak there would make it happier.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -3,7 +3,7 @@
     "react": {
       "pragma": "React",
       "version": "16.6",
-      "flowVersion": "0.142.0" // Flow version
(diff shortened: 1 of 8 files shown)


necolas:
Thanks. Merged into the 0.16 dev branch

PR #1981, opened 2021-04-08
Deduplicate yarn.lock

comp615:
Something interesting we did at Twitter that actually decreased our bundle size, and a little bit speeds up running yarn...dedupe:
```
npm install -g yarn-deduplicate
yarn-deduplicate yarn.lock
```

It takes all the ranges and re-resolves them against other things already in the lock file, so ^1.0.0 and ^1.0.1 won't have two entries when one could satisfy both. Yarn 2 already does this apparently. Anyways, it's potentially equivalent to minor version updates some places, so that's the risk of testing, but likely is mostly on the dev tooling side.

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -135,60 +135,19 @@
   dependencies:
     "@babel/highlight" "^7.10.4"
 
-"@babel/code-frame@^7.0.0":
-  version "7.0.0"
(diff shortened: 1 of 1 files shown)


necolas:
Done in 0.16-dev

Issue #1684, opened 2020-07-24
Exporting types


necolas:
I'm looking for input from people who use typed JavaScript with this library.

1. If you use Flow or TypeScript, how do you type `react-native-web` exports?
2. If you work across platforms, how do you deal with platform-specific differences in the APIs or components like `View`?
3. Does this library need to export Flow and TypeScript types?
4. Is there a way to generate TypeScript interfaces from Flow types so you don't need to maintain both?

EvanBacon:
Here is a useful dialog about TypeScript in RNW: https://github.com/necolas/react-native-web/issues/832

There is no DefinitelyTyped package for [`react-native-web`](https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/types/react-native-web). I'm not entirely sure how one would work (efficiently). I've seen people propose copying all `react-native` types and then extending them, and somehow aliasing `types/react-native` to `types/react-native-web`. This of course wouldn't work with other out-of-tree solutions so I would discourage the approach.

Personally I would opt for universal packages like ["react animated"](https://github.com/facebook/react-native/pull/28892) or "react text" (probably not with these names since they're all taken). These universal packages would implement iOS, Android, and web in the same place. Said approach would also fix some versioning issues and help reduce the custom bundler configuration required for universal React apps. 

I will note that RN is now moving towards a monorepo structure (creating packages like [react-native/normalize-color](https://github.com/facebook/react-native/tree/master/packages/normalize-color) color and [react-native/polyfills](https://github.com/facebook/react-native/tree/master/packages/polyfills)) which could make modifications like this more feasible. 

At Expo, we've refactored all of our universal React packages from flow to TypeScript (as opposed to adding external type definitions files `.d.ts`) because we find this is easier to work with from a DX perspective.

I personally think that having better types will help bring more visibility to important react-native-web features (e.g. `href` on `Text`). Due to the high demand in the Expo/React community for TypeScript support on web, I may look at creating some abstract packages which reexport react-native(-web) modules with proper types.

paularmstrong:
In Build Tracker (typescript, web-only), I pulled the react-native types from definitely-typed, then removed all of the iOS/Android specific stuff and added web-specific bits. Ideally I would take the few minutes to contribute this back, but I don't feel comfortable enough with my knowledge on Typescript to really be authoritative that any of it was done correctly.

[...]

PR #1982, opened 2021-04-08
Generate Flow files into dist

comp615:
Ok! So after some testing, I was able to get this mostly going. The key it turns out is to tell flow that it needs to key off the module field instead of the main field in package.json.

Luckily, it has a new option for this, however I had trouble figuring this out, so I wanted to try and feature it prominently in the docs. I think this is a great path forward as it will naturally stay in sync going forward. There are some quirks that we can improve internally (i.e. ViewProps is not a closed type, so it allows passing props that are not valid...but it's a start)


packages/docs/src/pages/docs/getting-started/multi-platform.md
```diff
@@ -65,10 +65,12 @@ module.exports = {
 
 ### Flow
 
-[Flow](https://flow.org) can be configured to understand the aliased module. You may also need to include a custom libdef ([example](https://gist.github.com/paularmstrong/f60b40d16fc83e1e8e532d483336f9bb)) in your config.
+[Flow](https://flow.org) can be configured to understand the aliased module and pull types from Reac Native Web's source code. To do this, you can alias the name as below. You will also need to tell flow to use the `module` field of RNW's `package.json` or it may not find the flow types.
 
 ```yml
 [options]
+module.system.node.main_field=module
+module.system.node.main_field=main
```
necolas, line 73: Are both these lines meant to be here, or just the first?

necolas: This isn't very well documented but I'm assuming flow picks the first recognized entry if there are several: https://flow.org/en/docs/config/options/#toc-module-system-node-main-field-string

necolas:
Merged into the 0.16-dev branch.

This should also resolve #1684

comp615:
Yes correct, it goes down the list till it finds a matching one. Not sure
from documents if it always puts main in, so included anyways.

PR #1985, opened 2021-04-10
Flow polish to more accurately reflect current behavior

comp615:
Working through integrating this at Twitter, I found several places the types were mismatched with how they function, or could be improved:
- Typed AppRegistry
- Dimension is always set
- Fix platformMethods callback that I messed up (it's not a return object, it's actually each as its own argument)
- Override some RN Event stuff for web (so target becomes an HTML element)
- Add statics to Image type (This isn't the most beautiful)
- Change one responder type to void | boolean, it's weird, but if you dig down the the React Native code, that type is the same, and there's some TODOs and confusion about why this is void/boolean too. So just matching to that.

NOTE: One area that I have yet to explore is some weirdness around Styles. I don't think we're getting the appropriate Style keys all the way up the chain yet. I'd like to be unable to put invalid style keys into Styles.create, but this may require exporting the types of styles or some heavy generic working. I'll likely have to seek help to get this right, but it's like a fun puzzle.


necolas:
Thanks. Merged

PR #1989, opened 2021-04-16
fix: make correct documentation site link

tranhiepqna:


necolas:
The link is correct

gonzalch:
@necolas I don't think the link is correct, if I navigate to https://necolas.github.com/react-native-web/ I get this 404 error page -
image

I think changing this to https://necolas.github.io/react-native-web/docs/ would be more useful.

Found this PR and made this comment only because I was about to open a similar one.

necolas:
Oh my bad, I read the PR backwards

necolas:
Fixed now. Thanks

PR #1996, opened 2021-04-22
Update linking example: use hrefAttrs instead of target

wishawa:
The `target` prop no longer works. The correct way to set `target="_blank"` is to set `hrefAttrs={{target: '_blank'}}`.


necolas:
Thanks

PR #1997, opened 2021-04-22
Update flow View/Pan Responder interaction

comp615:
**Problem**
Had a dangling change off my previous branch. In PanResponder, we create handlers with a certain signature. There's a comment in there specifically about onGrant, where RN isn't sure if it should be void or boolean. This creates a flow issue if you just loop the result back to View (because boolean isn't a return)

**Solution**
Make View's props have the same return signature as the PanResponders.

NOTE: It could be neat to directly just reference PanResponder here, but then the argument type needs to be PressEvent, and there's some disagreement about PressEvent between RN and RNW


necolas:
The internal issue was filed in 2015 and hasn't been touched since. I don't think the boolean return is used anywhere and it isn't documented in the API docs. The new RN renderer might make this pattern unnecessary in the future, as `defaultPrevented` on the event could communicate the same information to Android (and web).

> It could be neat to directly just reference PanResponder here

Since `PanResponder` is built on top of the Responder system, it's the Pan types that would be changed. However, for best compat with RN maybe we just stick with the weird `void | boolean`.

comp615:
Yeah as you point out, the other option is to change the return type inside the RN vendor stuff.
Tacked on one more change Lenny helped with. Making GenericStyle read only allows this to work:

```
const dynamicProps: {| accessibilityLabel: string |} = { accessibilityLabel: 'foo' };
return (<View style={dynamicProps} />);
```

In short, it let the type be up-cast from a more specific explicit type

necolas:
Thanks. These changes are in 0.16.1

PR #2021, opened 2021-05-13
Accessibility Ref List Bug Fixes

andrewhayward:
There are various accessibility props that can be either a string (ID Ref), or an array of strings (ID Ref List). There are various minor bugs relating to these props.

The relevant props are:
 * `accessibilityControls`
 * `accessibilityDescribedBy`
 * `accessibilityFlowTo`
 * `accessibilityLabelledBy`
 * `accessibilityOwns`

Docs
----
The accessibility docs only mention `string` as a valid type for each of these props. This patch updates the docs to instead use `string | Array<string>`.

Types
-----
Most of the props were correctly typed as `idRefList`, but `accessibilityDescribedBy` was not. This patch updates the relevant type to be `idRefList`.

Creating DOM Props
------------------
None of the props appropriately handled arrays of strings as a value. So while an array of strings could be passed in (for the most part), an valid DOM value was not generated, instead relying on default JavaScript array behaviour. This patch fixes this, generating string-separated values where an array is passed in.

Tests
-----
Tests have been added for the various props, to confirm that an array of values will generate a space-separated DOM value, rather than the previous comma-separated default that was being generated before.

### DIFF
--- a/packages/docs/src/pages/docs/concepts/accessibility.md
+++ b/packages/docs/src/pages/docs/concepts/accessibility.md
@@ -53,7 +53,7 @@ Equivalent to [aria-colindex](https://www.w3.org/TR/wai-aria-1.2/#aria-colindex)
 Equivalent to [aria-colspan](https://www.w3.org/TR/wai-aria-1.2/#aria-colspan).
 {% endcall %}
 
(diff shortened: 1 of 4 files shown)


necolas:
Thanks! Good patch. This is available in 0.16.3

PR #2024, opened 2021-05-17
fix: ScrollView style when refreshControl is present

aryo:
Currently passing in a refreshControl to a ScrollView causes the ScrollView's style to be used in the actual refresh control, but not the actual ScrollView itself.


necolas:
I don't understand what this patch is meant to fix. It only rearranges some code and removes the baseStyle

aryo:
@necolas the baseStyle applied to the ScrollView itself is hard overriding the style potentially passed in from props.

As is, the ScrollView styling kind of works by coincidence because props.style is being applied to the refresh control, which contains the ScrollView. It should be applied to the ScrollViewClass itself.

Issue #427, opened 2017-04-14
Investigate implementing yoga 'aspectRatio'


necolas:
aspectRatio: https://facebook.github.io/yoga/docs/aspect-ratio/

https://codepen.io/aghassemi/pen/RpdaKM

jordwalke:
I don't happen to know anything about the future direction. Are you looking to reimplement the behavior?

necolas:
I think it would be relatively easy to implement. However, if the only way to trigger "RTL mode" in React Native is to use `I18nManager`, and that _also_ flips `right` to `left`, then it doesn't seem very useful to support `start/end`. And Yoga doesn't flip values, right (`textAlign:left`)?

MoOx:
Would love to see this available in RNW. Anything planned? Can I help?

necolas:
Feel free to investigate. I haven't got any plans. I don't know if building this into StyleSheet for web is viable; maybe needs an `AspectRatio` component a bit like RN needs `LinearGradient` component but RNW supports `backgroundImage: linear-gradient(...)`.

MoOx:
Indeed, that could be tricky to implement in StyleSheet.

I tried to make a component based on the absolute stretch + padding-bottom trick, but that don't work correctly on Firefox with flexbox (this require display block to get this working - at least for now - see https://stackoverflow.com/a/39069462/988941)

I may try to share this one day or another (but my current code is in reasonml so I would have to convert I guess).

necolas:
I think you could use JS rather than CSS to calculate the dimensions once you're in a component

necolas:
Here is the `AspectRatio` component I wrote for Twitter Lite. (It doesn't handle everything the Yoga aspect ratio does. Only width-based)

```js
import { any, func, number } from 'prop-types';
import { createElement, StyleSheet, View, ViewPropTypes } from 'react-native';
import React, { Component } from 'react';

class AspectRatio extends Component {
  static displayName = 'AspectRatio';

  static propTypes = {
    children: any,
    onLayout: func,
    ratio: number.isRequired,
    style: ViewPropTypes.style
  };

  render() {
    const { children, onLayout, ratio, style } = this.props;
    const percentage = 100 / ratio;

    return (
      <View onLayout={onLayout} style={[styles.root, style]}>
        {createElement('div', {
          style: [styles.ratio, { paddingBottom: `${percentage}%` }]
        })}
        {createElement('div', {
          children,
          style: styles.content
        })}
      </View>
    );
[...]

PR #2038, opened 2021-06-01
[fix] Convert aspectRatio value to string

petrbela:
Turns out React expects the aspectRatio value to be a [string](https://github.com/frenic/csstype/blob/master/index.d.ts#L259), otherwise it ignores it.

This should make #427 finally work.


necolas:
Thanks!

Issue #2039, opened 2021-06-05
Cleanup DOM reference from  TouchableMixin


bhanu-allaparthi:
**The problem**
We are using react-native web package as part of our app development and we observed some memory leaks in the Touchable component mainly due to the following elements

_touchableNode is retaining DOM 

https://github.com/necolas/react-native-web/blob/e810f1fd2b41293cb1efe04e332fb6f8d4bcca65/packages/react-native-web/src/exports/Touchable/index.js#L341

this.state.touchable.responderID is retaining DOM:

https://github.com/necolas/react-native-web/blob/e810f1fd2b41293cb1efe04e332fb6f8d4bcca65/packages/react-native-web/src/exports/Touchable/index.js#L396

get locationX is creating a closure based leak around the touch.target property

https://github.com/necolas/react-native-web/blob/e810f1fd2b41293cb1efe04e332fb6f8d4bcca65/packages/react-native-web/src/exports/Touchable/index.js#L703

![image](https://user-images.githubusercontent.com/74990338/120875767-695b7380-c562-11eb-849c-4d93a4baa0ae.png)

**How to reproduce**

https://codesandbox.io/s/react-native-web-touchable-fire-twice-forked-w86cr

Simplified test case: 

Steps to reproduce:
1. Navigate to this link https://w86cr.csb.app/ on the browser and then enable mobile mode.
2. Click on the two touchable components A and B alternatively for 15 times

**Expected behavior**
The DOM elements has to cleaned up after unmounting the component.
After explicitly setting the above elements to null. There were no memory leaks due to them

```
componentWillUnmount: function () {
    if (this._touchableNode && this._touchableNode.addEventListener) {
      this._touchableNode.removeEventListener('blur', this._touchableBlurListener);
    }
    this.touchableDelayTimeout && clearTimeout(this.touchableDelayTimeout);
    this.longPressDelayTimeout && clearTimeout(this.longPressDelayTimeout);
    this.pressOutDelayTimeout && clearTimeout(this.pressOutDelayTimeout);
    this.pressInLocation = null;
    this.state.touchable.responderID = null;
    this._touchableNode = null
  },
```

**Environment (include versions). Did this work in previous versions?** 

I don't think so.

* React Native for Web (version): TBC
* React (version): TBC
* Browser: TBC

necolas:
Thanks. I'll review a PR to fix this

PR #2047, opened 2021-06-07
Explicitly clearing the dom nodes to prevent memory leaks

bhanu-allaparthi:
This PR will address the memory leak concerns in the touchable component. 

Related issues.
https://github.com/necolas/react-native-web/issues/2039


necolas:
Thanks! This will be included in a new patch release later today

PR #2049, opened 2021-06-08
[fix] Allow focusable={false} to set tabIndex="-1"

### LINKED ISSUE #2046: Allow to set tabIndex -1 to implement Roving tabindex pattern
Roving tabindex is the basic accessiblity pattern for grouped focusable items.

RNfW should provide API for that so users don't have to hack it with setNativeProps in useEffect which can be too late.

There is no alternative for that.

**Additional context**
Here is an example https://codesandbox.io/s/gallant-germain-ujqnm?file=/src/App.js

necolas:
Close #2046

cc @steida


steida:
LGTM. Thank you for your attention to detail.

Issue #2051, opened 2021-06-09
onPress invoked by keyenter leaks


steida:
**The problem**
Key navigation with RNfW often leads to leaks.
`Warning: Can't perform a React state update on an unmounted component.`

**How to reproduce**
Just tab to the button and press enter.
Simplified test case: https://codesandbox.io/s/pedantic-mirzakhani-vwhjx?file=/src/App.js

**Expected behavior**
No leaks.

**Environment (include versions). Did this work in previous versions?**
No.

* React Native for Web (version): Last few versions.
* React (version): Always the last.
* Browser: Chrome

Here it fails with touchable opacity, but Pressable fails too.

steida:
It leaks less but it still leaks. 

![Screenshot 2021-06-17 at 19 38 23](https://user-images.githubusercontent.com/66249/122447039-9ae12f80-cfa3-11eb-964e-e35da415d42f.png)

PR #2053, opened 2021-06-09
[fix] Ignore transition after component unmount

Federkun:
fix https://github.com/necolas/react-native-web/issues/2051

Prevent `setOpacityTo` to run if the component unmount. When using the press event, and `onPress` result in the component unmounting, onPressEnd will still run after it, resulting in react warning about a memory leak.


necolas:
Thanks but this isn't a fix for the problem

PR #2055, opened 2021-06-10
[fix] Change logic of PressResponder for keybaords

### LINKED ISSUE #2051: onPress invoked by keyenter leaks

necolas:
This patch removes the keyboard `onClick` shim from `unstable_createElement` and adds it to the `PressResponder`.

Previously, the logic in `createElement` was calling `onClick` during `onKeyDown`. This could result in React state update warnings if the host element was no longer in the DOM during `keyup`.

Fix #2051


steida:
LGTM

necolas:
2a930b7e3f7f58638e457e1f15b608c6f6c010e8

PR #2061, opened 2021-06-22
Removing persistent `clickHandler` in favour of checking for usefulness

andrewhayward:
I noticed that `Text` always has an `onClick` handler, whether it's needed or not. This can be confusing for some screen reader users, because their screen reader will announce an element as "clickable" if it has a click event handler attached. To reduce that confusion, this patch suggests using `React.useMemo` to only generate a click handler as needed.

Make of it what you will - just a thought.

### DIFF
--- a/packages/react-native-web/src/exports/Text/__tests__/index-test.js
+++ b/packages/react-native-web/src/exports/Text/__tests__/index-test.js
@@ -155,6 +155,22 @@ describe('components/Text', () => {
     });
   });
 
+  describe('prop "onClick"', () => {
+    test('is called', () => {
+      const onClick = jest.fn();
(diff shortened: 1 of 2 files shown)


necolas:
Thanks, good catch.

I think my preferred fix would be to wrap the existing `handleClick` function in a `useCallback` and then check if `onClick`/`onPress` is defined before setting it to the props.

andrewhayward:
I've adjusted things, as per preference, and the (now memo-ised) callback is only applied if `onClick`/`onPress` is present. At least, I think that's what you were getting at!

necolas:
Thanks. This is available in 0.17.1

PR #2071, opened 2021-06-25
Pass onLayout when a child is sticky

denghui1010:
every cell in VirtualizedList will be layout, but when a cell is sticky, the cell will be wrapped by a new View who has not pass the onLayout prop


quantizor:
is this patch still needed?

PR #1709, opened 2020-08-20
Make keyboardType work as expected for numeric input types

RichardLindhout:
Fixes: https://github.com/necolas/react-native-web/issues/1705

PR #2108, opened 2021-08-21
Fix keyboard number type

petrbela:
Numeric/decimal keyboardType should also set type=number to allow only numeric keys on desktop and mobile.

### DIFF
--- a/packages/react-native-web/src/exports/TextInput/__tests__/index-test.js
+++ b/packages/react-native-web/src/exports/TextInput/__tests__/index-test.js
@@ -171,18 +171,21 @@ describe('components/TextInput', () => {
       const { container } = render(<TextInput keyboardType="decimal-pad" />);
       const input = findInput(container);
       expect(input.inputMode).toEqual('decimal');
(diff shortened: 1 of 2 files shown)


petrbela:
Related to numeric input, could RNW also support `min`, `max` and `step`? I know they don't exist in RN but they'd simply be ignore there. I can add it to this PR or create a separate one.

necolas:
See https://github.com/necolas/react-native-web/pull/1709

petrbela:
@necolas Fair enough. Would you be okay with adding web-only `type`, `min`, `max`, and `step` props to enable explicitly enforcing specific browser behavior when the developer chooses to do so?

necolas:
I don't think so, as that would puncture the API and allow completely different results to be produced on web.

You can either implement that functionality cross platform in JS, or use the `unstable_createElement` API to create a custom web element using those props, and decide where to fork/recombine that with your RN code in your product.

necolas:
But this is a relevant issue for future xplat react APIs, as it surfaces more limitations of RN's `TextInput` API

PR #2121, opened 2021-09-04
Fixes missing 'n' characrer

kahoca:

PR #2137, opened 2021-09-23
fix: Patch class names in documentation

sawa-zen:
## What
Fixed two typos in the documentation.
- Change `rn-` to `r-`.
- Correct class name for `margin-top`.

## Ref
- https://github.com/necolas/react-native-web/pull/2121


sawa-zen:
@necolas 
As a side note, the documentation for the `React GUI` is similarly misrepresented.
https://ixrvy.sse.codesandbox.io/docs/styling/#implementation-details

Issue #1517, opened 2020-01-03
Styles should work with ShadowDOM roots.


xdumaine:
**Is your feature request related to a problem? Please describe.**

When using styles inside a shadow root, styles are not applied, because react-native-web attaches styles to the document root

**Describe a solution you'd like**

Automatically attach styles to the shadow root instead of the document root if one is present. Resolve issues of cascading rules where a shadow root breaks up parents from children.

**Describe alternatives you've considered**

Avoid having styles compiled into a stylesheet with classes altogether, and have all styles inlined on components.

**Additional context**

The primary context here is that my react app is rendered inside a custom element, which can be included on anyone else's page as a widget. In order to get style isolation, I must render my app inside a shadow root (natively, or using `react-shadow`). This gives me proper style isolation from the outer (uncontrolled) application, but it also causes me to lose all of my styles from react-native-web.

danperkins:
@necolas I was able to get a solution working where I add a style node as a child whenever mounting a new shadow DOM node.  After that I added this functionality to `createStyleResolver`:
```
var sheets = {};
var savedRules = [];

// Call this when a new shadow DOM node is added passing in the reference to the style sheet
export function addSheet(id, sheet) {
  var orderedSheet = createOrderedCSSStyleSheet(sheet);
  sheets[id] = orderedSheet;
  savedRules.forEach((x) => orderedSheet.insert(x.style, x.group));
  return orderedSheet;
}

// Use this instead of sheet.insert
function insert(style, group) {
  Object.values(sheets).forEach(function(sheet) {
    sheet.insert(style,group);
  });
  savedRules.push({style: style, group: group});
}
```

I saw that `NativeMethodsMixin` has an approach for providing a custom style resolver when resolving styles but I couldn't figure out a way to add a custom resolver (or override the default resolver) in the normal `createElement` flow.  If there is an easy way to do this could you point me to it?  If not and you have an idea about how the infrastructure could be extended to support this let me know and I would be happy to submit a PR

xdumaine:
@necolas any chance you could give us some brief thoughts or guidance on this one?

necolas:
See #2208 for a build that should support ShadowDOM use

Issue #1765, opened 2020-10-12
Using react-native-web with ReactDOM.createPortal


timothylui:
**Is your feature request related to a problem? Please describe.**
Hi, I'm creating an electron app with multiple windows and we're currently using window.open and then using ReactDOM.createPortal to set the document body.

This works great if you're using straight react, but when we're trying to use react-native it is missing the styles. 

This also happens when trying to do the same thing in the browser, so isn't related to electron.

**Describe a solution you'd like**

As ReactDOM.createPortal creates a section of the dom, a way to set up the page so that react-native-web components render correctly. Alternatively, if this works already, a working example would be great.

**Describe alternatives you've considered**

Writing it in pure react but that means we can't reuse the components

necolas:
Shouldn't you be using `render` rather than `createPortal`?

Multiple document roots isn't supported at the moment

wjps:
I've been fighting with the same problem so would also be interested in the best way to get in working. Would something like `react-gateway` work?

I'm not sure how I'd use `render`  to solve the problem though.

nandorojo:
I've been looking for something like `react-gateway` too. Now that RNW supports modals, I think a modal-based solution would work for popovers. I came across https://github.com/SteffeyDev/react-native-popover-view/issues/82 today, so I'll see if that works with the new RNW.

wjps:
Ok, I've managed to get things working by doing 
```
import App from '../App';
import { AppRegistry } from 'react-native-web';

AppRegistry.registerComponent('App', () => App);

then rendering the css into the new window with 
```

necolas:
Multiple roots should now be supported in 0.18

PR #2138, opened 2021-09-24
feat: shadow root support

jon-w1:
### Background
Importing `react-native-web` kicks off `StyleSheet` resolution on a single `style` element which gets added to the `head` of the HTML document. However, when rendering an application within a Shadow DOM root (e.g. when encapsulating within an HTML custom element), any stylesheets from the main `document` are ignored. 

### Issue Link
https://github.com/necolas/react-native-web/issues/1517

### In This Update
- Updated `renderApplication` to check if the `rootTag`'s `RootNode` is of `ShadowRoot` type. If so, an additional `style` tag is added to the shadow root
- Modified `createStyleResolver` to insert any new rules into both the document-level stylesheet as well as any additional shadow root stylesheets

This is my first PR for RNW so let me know if there's anything else I should consider!

### DIFF
--- a/packages/react-native-web/src/exports/AppRegistry/renderApplication.js
+++ b/packages/react-native-web/src/exports/AppRegistry/renderApplication.js
@@ -31,6 +31,10 @@ export default function renderApplication<Props: Object>(
 
   invariant(rootTag, 'Expect to have a valid rootTag, instead got ', rootTag);
 
+  if (rootTag.getRootNode() instanceof ShadowRoot) {
+    styleResolver.addShadowSheet(rootTag);
+  }
+
   renderFn(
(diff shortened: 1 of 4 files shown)


necolas:
I think there was a similar issue with rendering into sourceless iframes. Perhaps this patch could be adapted to include support for that use case too?

necolas:
And it's possible this overlaps with the issue for using multiple windows (i.e., roots) in an electron environment #1765

necolas:
@jon-w1 following up on the comments above. Thanks

zarubond:
> And it's possible this overlaps with the issue for using multiple windows (i.e., roots) in an electron environment #1765

Here I am creating a fix for multiple windows/iframes https://github.com/zarubond/react-native-web/tree/zarubond/multi_root

PR #2149, opened 2021-10-24
module fbjs/lib/ExecutionEnvironment does not have named export canUseDOM

nifgraup:


Cavitt:
Thanks! This fixes react-native-web loading via browser modules.

Issue #2162, opened 2021-11-08
Focus jumps between TextInputs


maevic:
**The problem**

Adding the property `selectTextOnFocus` to multiple `TextInput`s can cause the focus to jump between them indefinitely.

**How to reproduce**

Simplified test case: 
https://codesandbox.io/s/nifty-kowalevski-tbmev

Video:
![screen](https://user-images.githubusercontent.com/15337940/140756942-8ba68edd-ee88-45d3-aa23-035de0c9bda2.gif)

Steps to reproduce:
1. Click in first input
2. Click another window, so the browser looses the focus
3. Click in second input
4. The focus should now jump between them

**Expected behavior**

The last clicked input should get the focus.

**Environment (include versions). Did this work in previous versions?**

* React Native for Web (version): 0.17.5
* React (version): 17.0.2
* Browser: Chromium version 95.0.4638.54

maevic:
Yes, I am on Linux. If I recall correctly the issue occurred on Windows as well, but I'm not entirely sure about that. Anyway, thanks for your help.

PR #2163, opened 2021-11-08
[fix] Focus jumps between TextInputs

maevic:
Adding the property selectTextOnFocus to multiple TextInputs can cause the focus to jump between them indefinitely. This PR fixes the corresponding issue #2162.

### DIFF
--- a/packages/react-native-web/src/exports/TextInput/index.js
+++ b/packages/react-native-web/src/exports/TextInput/index.js
@@ -243,7 +243,8 @@ const TextInput: React.AbstractComponent<
       }
       if (selectTextOnFocus) {
         // Safari requires selection to occur in a setTimeout
-        setTimeout(() => {
+        clearTimeout(TextInput.focusTimeout);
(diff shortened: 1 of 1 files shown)


packages/react-native-web/src/exports/TextInput/index.js
```diff
@@ -243,7 +243,8 @@ const TextInput: React.AbstractComponent<
       }
       if (selectTextOnFocus) {
         // Safari requires selection to occur in a setTimeout
-        setTimeout(() => {
+        clearTimeout(TextInput.focusTimeout);
+        TextInput.focusTimeout = setTimeout(() => {
```
necolas, line 247: The timeout should be managed as a ref or callback instead of this approach

maevic: If we used refs we'd only have access to this instance's timeout. But we need to clear a timeout created by another TextInput instance.

Issue #1758, opened 2020-09-29
Switch disabled track and thumb color


reck753:
**Is your feature request related to a problem? Please describe.**
No.

**Describe a solution you'd like**
I saw that the disabled colors are hard-coded so I would only change them with variables passed from props. Eg. I would add two more props disabledTrackColor and disabledThumbColor. I feel like that styling should be adjustable, because not everyone is happy with those colors.

**Describe alternatives you've considered**

**Additional context**
Eg. I have a switch with red track and white thumb and I would like it to have some light red color for track and some light grey color for thumb when disabled, instead of those that are hard-coded.

bouaoud:
same here

Moe-kanan:
+1

makkarMeenu:
@reck753 [This](https://github.com/necolas/react-native-web/pull/2164) PR resolves the above issue in a different way.
You can change trackColor and thumbColor  conditionally according to disabled condition. 
I think, we can't introduce new props as react-native also doesn't provide `disabledTrackColor` and `disabledThumbColor`.

reck753:
> @reck753 [This](https://github.com/necolas/react-native-web/pull/2164) PR resolves the above issue in a different way. You can change trackColor and thumbColor conditionally according to disabled condition. I think, we can't introduce new props as react-native also doesn't provide `disabledTrackColor` and `disabledThumbColor`.

Hi, @makkarMeenu!
Thanks for the PR! This would be a good enough solution for me.

PR #2164, opened 2021-11-09
fix: removed hardcoded value for switch in disabled state

makkarMeenu:
Fixed https://github.com/necolas/react-native-web/issues/1758
- Removed hardcoded values for disabled track colors and thumb colors.
- Now we will be able to change track colors and thumb colors in `disabled` state (same as react-native does)

### DIFF
--- a/packages/examples/pages/switch/index.js
+++ b/packages/examples/pages/switch/index.js
@@ -24,6 +24,14 @@ export default function SwitchPage() {
         <Switch disabled={true} value={false} />
         <Divider />
         <Switch disabled={true} value={true} />
+        <Divider />
+        <Switch activeThumbColor="#fff" activeTrackColor="#E0245E" disabled={true} value={true} />
+        <Divider />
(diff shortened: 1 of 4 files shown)


necolas:
Thanks. Please could you also include an example of a custom disabled color being used in the 'examples' app?

makkarMeenu:
> Thanks. Please could you also include an example of a custom disabled color being used in the 'examples' app?

Hey, just added a new [commit](https://github.com/necolas/react-native-web/pull/2164/commits/6261cd7b231ec2039de02c78b7ac110d4615fc2c) for all possible examples for disabled state in switch component.

necolas:
This change needs unit tests, and it would be good to simplify the logic

makkarMeenu:
> This change needs unit tests, and it would be good to simplify the logic

Added snapshot test cases [here](https://github.com/necolas/react-native-web/pull/2164/commits/ba5cdf22935b392f47ac51d817ed2da4fc9f1afe) for the above changes  . 
I don't see a room for simplifying the logic, any suggestions are welcomed.

PR #2190, opened 2022-01-03
ResponderTouchHistoryStore into instanciated class

zarubond:
Refactor ResponderTouchHistoryStore from singleton to class instantiated in ResponderSystem. The eventual the plan is to have multi-window support in RNW which requires having an instance of this Store for each window.  PR code does change the internal logic. 

This PR is a part of greater effort to enable support for multiple browser windows in RNW.  Whole change is available in the draft PR https://github.com/necolas/react-native-web/pull/2185 .

### DIFF
--- a/packages/react-native-web/src/modules/useResponderEvents/ResponderSystem.js
+++ b/packages/react-native-web/src/modules/useResponderEvents/ResponderSystem.js
@@ -232,6 +232,7 @@ let currentResponder: ResponderInstance = {
   node: null,
   idPath: null
 };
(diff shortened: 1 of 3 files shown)


zarubond:
@necolas If you would have time, could you please check the PR, or give me any indication about future of https://github.com/necolas/react-native-web/pull/2185

necolas:
Yes I'm interested in supporting multiple windows/iframes. Thanks, this PR looks OK in principle. I'm currently focused on other updates and will have to come back to this later though

necolas:
Thanks, this is now in master.

PR #2195, opened 2022-01-10
fix(Text): Support ellipsizeMode="clip"

evancharlton:
The `Text` component supports different options for `ellipsizeMode`,
rather than assuming a trailing-edge '...'.

This change adds support for `ellipsizeMode="clip"` -- but only in one
special case: `numberOfLines={1}`. This is because that's all that is
supported by the browser vendors at the moment, so it's the best that
can be done by pure CSS.

Additionally, ths expands the handling of `numerOfLines` to accommodate
the value `0`, as react-native specifies that a value of `0` is [the
same as not specifying anything][rn docs].

[rn docs]: https://reactnative.dev/docs/text#numberoflines

### DIFF
--- a/packages/examples/pages/text/index.js
+++ b/packages/examples/pages/text/index.js
@@ -383,21 +383,37 @@ export default function TextPage() {
               <Text> words</Text>
             </View>
           </View>
-          <Text numberOfLines={1} style={{ marginBottom: 20 }}>
-            Maximum of one line, no matter how much I write here. If I keep writing, it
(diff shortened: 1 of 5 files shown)


evancharlton:
I realize that this fixes two bugs in one change; I'm more than happy to split them into separate PRs if you'd prefer :)

necolas:
Going to close this because a number of changes have been made since this PR. Feel free to create a new PR on top of master. Thanks

Issue #1636, opened 2020-06-10
Allow Image to not set alt property so browsers can provide enhanced functionality


comp615:
**Is your feature request related to a problem? Please describe.**
Currently, when a site uses an <Image> element, they _may_ provide an accessibilityLabel property, commonly referred to as alt text. When RNW interprets the image, it will create two elements: an outer View and an inner-hidden img.

When set, the accessibility label is applied to the View and as `alt` to the img.
When not set, nothing is applied to the View, but the img gets an empty alt set (alt='').

Empty alt implies that the image is decorative only in nature, an example of an appropriate alt='' image might be an avatar on Twitter, because the visual information is duplicative with the username shown next to it (usually).

Separately, Chrome provides interesting functionality I just learned about to auto-caption images that are unlabeled. https://support.google.com/chrome/answer/9311597?hl=en. This potentially makes sites, especially those with user generated content, much more accessible to users. This functionality works by searching for img tags without alt, and using ML to label them for the user.

However, because the hidden img tag _always_ has an alt set (albeit empty), Chrome will not label or detect any such images.

I'm proposing to reconsider a little bit how accessibility labels are handled in Image, primarily to enable this functionality for users.

**Describe a solution you'd like**
Right now, the View accessibilityLabel duplicates the hidden img label. Instead, we could generate DOM ids, and have View be labelled-by img. This creates one source of truth. (optional)

From there, I'd propose only setting alt on img, if accessiblityLabel is explicitly set (including '' or null which would set it as today). If accessibilityLabel = undefined, no alt prop would be set on img.

https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/Image/index.js#L259

**Describe alternatives you've considered**
Another option would be to completely separate the img label and allow that to be manually controlled. One potential perk of doing it this way, is that I think the reason for setting '' as the default alt is to prevent clutter of unlabeled images from screen reader users or devs who don't leverage altText. So this would provide an opt-in path, but I don't think is aligned with RN spec.

**Additional context**
Currently with the suggested pointer label approach, Chrome will not use the ML generated caption for the outer View. I will separately ask them to consider this based on the outcome of this discussion (https://bugs.chromium.org/p/chromium/issues/detail?id=1093364). Until then, we'll provide a hidden text of "image" so that users navigating with j/k understand there is an image/graphic there.

necolas:
Closing as this feature is being tracked in #1786

Issue #2171, opened 2021-12-02
Image doesn't work when cross-origin.


ngokevin:
**The problem**

Image component fails with cross-origin URL over the network. I was able to patch it.

**Solution**

Set `image.crossOrigin = 'anonymous'` in `modules/ImageLoader` and `hiddenImage.crossOrigin = 'anonymous'` in `Image/index`.

Then display an `img` tag, not the background image view. Perhaps this could be a `prop`, but background image URLs fail to load over CORS.

With my patch-package, my image shows.

```
diff --git a/node_modules/react-native-web/dist/exports/Image/.index.js.swp b/node_modules/react-native-web/dist/exports/Image/.index.js.swp
new file mode 100644
index 0000000..e652063
Binary files /dev/null and b/node_modules/react-native-web/dist/exports/Image/.index.js.swp differ
diff --git a/node_modules/react-native-web/dist/exports/Image/index.js b/node_modules/react-native-web/dist/exports/Image/index.js
index 5c25ebf..4c7bdb6 100644
--- a/node_modules/react-native-web/dist/exports/Image/index.js
+++ b/node_modules/react-native-web/dist/exports/Image/index.js
@@ -225,6 +225,7 @@ var Image = /*#__PURE__*/React.forwardRef(function (props, ref) {
   var hiddenImage = displayImageUri ? createElement('img', {
     alt: accessibilityLabel || '',
     classList: [classes.accessibilityImage],
+    crossOrigin: 'anonymous',
     draggable: draggable || false,
     ref: hiddenImageRef,
     src: displayImageUri
@@ -309,14 +310,6 @@ var Image = /*#__PURE__*/React.forwardRef(function (props, ref) {
     pointerEvents: pointerEvents,
     ref: ref,
     style: [styles.root, hasTextAncestor && styles.inline, imageSizeStyle, flatStyle]
-  }), /*#__PURE__*/React.createElement(View, {
-    style: [styles.image, resizeModeStyles[resizeMode], {
-      backgroundImage: backgroundImage,
-      filter: filter
-    }, backgroundSize != null && {
-      backgroundSize: backgroundSize
-    }],
-    suppressHydrationWarning: true
   }), hiddenImage, createTintColorSVG(tintColor, filterRef.current));
 });
 Image.displayName = 'Image'; // $FlowIgnore: This is the correct type, but casting makes it unhappy since the variables aren't defined yet
[...]

Issue #1786, opened 2020-10-27
Image: modern API and implementation


necolas:
**Is your feature request related to a problem? Please describe.**

The `Image` component is old and somewhat constrained by React Native design choices. This issue is to discuss how we should rewrite `Image` to support modern web features. 

**Describe a solution you'd like**

The new `Image` should support various features:

1. Support browser ML generated `alt` text (#1636)
2. Support native lazy loading (#1542)
3. Display image in Windows High Contrast mode (#1298)
4. Support native DPI image switching
5. Available for SSR

Some of these features are in opposition to certain React Native features:

1. HTTP properties in `source` definitions
2. `onProgress` etc callbacks
3. The `repeat` value of `resizeMode`.

I have an early implementation of a base `Image` API that looks like this:

```
<Image
  alternativeText=''
  crossOrigin='anonymous'
  decoding='async'
  draggable='false'
  focusable='false'
  loading='lazy'
  nativeID=''
  onError={() => {}}
  onLoad={() => {}}
  source={[ {uri, scale}, {uri, scale} ]}
  style={}
  testID=''
/>
```

Any dynamic loading via XHR/fetch would have to be layered upon that, as it would be if using a native `img` element. There would be no support for `size` and other APIs that are basically media queries related to viewport dimensions.

**Additional context**

Please comment with addition context and use cases.

comp615:
We have two uses for the image queryCache I wanted to share here

1. When dataSaver is on, we try and get the largest cached variant already loaded to save data.
  ```
  // If data saver is on, then attempt to use the highest quality image we've already loaded to save data,
    // Exclude tiny variant (pre-data) saver images, so that we have a minimum quality
    const dataSaverCachedSource =
      dataSaver &&
      loadingVariant &&
      Math.max(loadingVariant.width, loadingVariant.height) >= MIN_RESOLUTION_CACHED_DATA_SAVER_THRESHOLD
        ? loadingVariant.uri
        : null;
    const src = dataSaverCachedSource || _selectSource(photo);
```

2. We use `defaultSource` to show the largest loaded image while the full-size variant loads:
```
<Image
        defaultSource={loadingVariant && loadingVariant.uri}
        draggable
        onContextMenu={onContextMenu}
        // If there's a fallback image showing, that's good enough and we can swallow the error
        onError={loadingVariant ? undefined : onError}
        onLoad={onLoad}
       // These just set internal state and shows a little loading spinner so people know we are upscaling the image
[...]

Issue #1542, opened 2020-02-24
Support <img loading="lazy" /> attribute


MoOx:
**Is your feature request related to a problem? Please describe.**

Recently Chrome added a native way to lazy load images, it would be nice to be able to leverage it.

**Describe a solution you'd like**

We could simply add support for this attributes that accepts 3 values "auto" (default), "lazy", "eager". We don't have to implement anything beside passing this to the dom `<img`.

That said, I am not sure what is the best way to implement this for img only as Image rely on View+propsToAriaRole.

_Would you be open for a PR to add this as escape hatch into `filterSupportedProps`?_

**Describe alternatives you've considered**

I think it's best to not really think about that since it's a recent feature, which make sense the way it is imo.

**Additional context**

It's especially interesting for SSR to get good perf notes for a good SEO :D

MoOx:
One thing I didn't think about is that Image actually render img tag just for convenience & the actual rendering is using css `background-image`... I guess in it's current state it's not possible to resolve this issue then...

necolas:
RN's `Image` is probably the API with the biggest differences from web images. Ultimately I suspect it will need to be redesigned significantly to better support web targets, e.g., `Image` shouldn't support any `resizeMode` types that cannot be implemented with `object-fit` and the ability to load images with HTTP headers also probably needs to be moved into a separate API from the rendering of images.

MoOx:
Yeah I will probably try to make another API with more restriction to handle images via img on the web & Image in RN for the moment.
Should we close this issue then or keep it open as a reminder for futur improvements? Not sure what is best move here

Sharcoux:
Well... I'd say that [this PR](https://github.com/necolas/react-native-web/pull/1686) could help handle this case easily.

The purpose is the same: easily integrate web specificities in a react-native codebase.

Sharcoux:
To solve this issue, I think that I'm able to propose a PR that would make `img` the new underlying component for Image, but for that, it would be quite helpful to rely on createRNElement PR as it would do half of the job. The second half can be to map resizeMode to object-fit, source to src and source.uri to src. We could then pass by the other props to the img element.

Without the createRNElement PR, we have to copy paste the View code to have an underlying <img /> element instead of a div.

Another interesting point is that createRNElement could help solving the issue temporary by creating an img element more easily.

So @necolas , why is this off topic, please?

necolas:
Closing as this feature is being tracked in #1786

PR #2207, opened 2022-01-27
Image updates: remove BG-style, add crossOrigin functionality, allow nullable alt text

comp615:
**Problem**
As noted in #1786 there are a few ways that images could serve customers better. In particular, this diff explicitly aims to fix: #1636 and #2171. (Happy to break this up to focus on each individually but was working on an integration test w/ Twitter)

**Solution**
While a full rewrite may be necessary to solve all those problems, I think a number of these issues stem from the usage of a pseudo-img-element on top of a background-image design. So if we can remove that, it may help bring focus to the remaining issues. In short, we can provide a sort of interim solution.

The crossOrigin image functionality represents a class of general problem: we cannot use all the native image functionality because we are constrained by ultimately using background image (e.g. see #1542). With privacy sandbox and site isolation coming, CORS authenticated images are necessary for us, this works OK for a normal image, but background-images have no equivalent so it's impossible to accomplish.

Getting rid of the background image also simplifies the DOM and likely fixes things like Windows High Contrast and the ML accessibility issues as well.

**CORS / Cross Origin**
While the RN API is to use body and headers on the source object, I do not believe this model works with the web. Because we are subject to CORS, and have server-only cookies. We necessarily need to use flags instead of explicit headers.

To solve this, I decided to add a crossOrigin mode flag matching the img element to the source definition.

This change necessitated tweaking ImageLoader for the same reason. This widens the API to take a source object not just a URI on the load method. I believe RN API also misses the mark here. If an image can be loaded with headers...how are you supposed to prefetch and object with headers? I think we should generally assume that all ImageLoader methods should take any valid source type instead of just a string URI.

**Object Fit**
My interpretation of the existing code is that visually, we could only get the proper resizing behavior by using background-size. This meant that for accessibility and usability (i.e. saving an image), we previously had to put an invisible "Real" img tag over the background-image. Based on that, it seems like using object-fit gives us the best of both worlds and helps simplify the code (we no longer need onLayout callbacks and so on).

The tradeoff is that [browser support](https://caniuse.com/object-fit) does not match RNW's current [stated support](https://necolas.github.io/react-native-web/docs/browser-compatibility/). In particular old Edge and IE. We could get around this by using an [Object fit Polyfill](https://github.com/constancecchen/object-fit-polyfill). So that's a disucssion.

BREAKING!: Additionally, because it's not replicable via objectFit...this drops support for the `repeat` resizeMode.

**Tests**
I tried to verify this works in our core image cases on Twitter, but we do not exercise the full functionality of this (i.e. ImageBackground or certain image props like blur or tint).

It appeared that the image related examples looked the same before and after, and that the snapshot changes were reasonable.


packages/react-native-web/src/exports/Image/index.js
```diff
@@ -194,54 +157,36 @@ const Image: React.AbstractComponent<ImageProps, React.ElementRef<typeof View>>
...
-      if (resizeMode === 'center' || resizeMode === 'repeat' || onLayout) {
-        const { layout } = e.nativeEvent;
-        onLayout && onLayout(e);
-        updateLayout(layout);
-      }
-    }
-
     // Image loading
+    // NOTE: in order to prevent costly reloads when objects aren't equal, we only check the resolved URI here
+    // however, what we really want is to deepEqual test the source var on all props that might affect loading.
```
kidroca, line 177: I've got the same problem for my PR: #2442 where it's possible to have `source.headers` 

The way I handle it is to extract `source` to local state, and only change it when the input source have deeply changed

https://github.com/necolas/react-native-web/blob/abbfe8f8eb81d859b274f87822b30f9bfa580031/packages/react-native-web/src/exports/Image/index.js#L369-L380

This will in turn trigger the loading effect hook

To make this work I've changed the `resolveAssetUri` to `resolveSource` and always work with the object shape of the source. This is to make the rest of the logic simpler and not having to guess whether the source is `number` or anything else 
But this results in some conflicts with your PR 
Can you take a look and give you opinion on how to move forward - maybe we can move the hook / change detection logic to your PR

comp615: On my phone while traveling so not diving into the code too much. But in general, I think it would be great to patch on top of this. I strongly believe a broader image rewrite is merited, and this solves several tasks on that front.

The only issue is that I'm not sure if/when this will end up in master or the other image work will land, so it may take a bit of finagling and ordering to get all the image related patches and work together.

PR #2208, opened 2022-01-28
StyleSheet with styleq runtime

### LINKED ISSUE #1517: Styles should work with ShadowDOM roots.

### LINKED ISSUE #2007: Shorthand style properties overwriting the more specific properties
First i'd like to say I really appreciate this project, thanks everyone ❤️ 

**The problem**

Less specific shorthand style properties are being applied instead of the more specific styles if there is another element also with one of the more specific styles.

This is due to the common styles being pulled out as class names for de-duplication.

**How to reproduce**

Simplified test case: https://codesandbox.io/s/dazzling-gould-ihiig

Steps to reproduce:
1. Open the link
2. The second View has horizontal paddings of 8px instead of 40px

Note that moving the first View below the second causes the issue to go away.

**Expected behavior**

I expect the horizontal padding to be 40px as this matches the behaviour in React Native.

* React Native for Web (version): 0.16.1
* React (version): 17.0.1
* Browser: Chrome

This appears to be an issue in previous versions as well

**Additional context**

I did some digging and this appears to be caused by the way that styles are de-duplicated. The first View get a class name applied with a `padding-left` & `padding-right` of `40px`. 

The second View then applies the same class and removes the horizontal padding in order to de-duplicate the styles. However this just leaves the `padding: 8px` in the style prop which then takes precedence over our de-duplicated padding class name. See here:

Screenshot 2021-05-01 at 15 36 15

I believe this is caused by `_resolveStyle` here https://github.com/necolas/react-native-web/blob/8fdbbc8d021ba3cb6a75f8b523a36da3c1be7007/packages/react-native-web/src/exports/StyleSheet/createStyleResolver.js#L147

I'm also willing to help out with a PR if someone can give me some pointers on the expected behaviour for how these styles should be applied.

Thanks 🙏

### LINKED ISSUE #2196: React 18 support
Any plans/roadmap to support react 18 concurrent features?

Looks like style system need changes
https://github.com/reactwg/react-18/discussions/110

### LINKED ISSUE #2210: Mount text tree performance tests broken
**The problem**

Benchmark "Mount text tree" on this page: https://necolas.github.io/react-native-web/benchmarks/ is broken for some libraries

**How to reproduce**

Simplified test case:  N/A

Steps to reproduce:
1. Go to https://necolas.github.io/react-native-web/benchmarks/
2. Select "Mount text tree" in benchmark. 
3. Then select `styled-components` in library (also see for other libraries except `react-native-web`)
4. React App crashes
5. In console, the following error is shown:
```
Error: Minified React error #130; visit https://reactjs.org/docs/error-decoder.html?invariant=130&args[]=undefined&args[]=
```

**Expected behavior**

The performance tests application should not crash

* React Native for Web (version): N/A
* React (version): N/A
* Browser: Firefox 96

necolas:
## StyleSheet refactor

Build: `npm i react-native-web@0.0.0-b096cd2d`

* Improves React Native compatibility by making `StyleSheet.create` the identify function.
* Improves RTL support, correctly styling nested LTR/RTL contexts.
* Supports React 18 by inserting styles on eval, not render. SSR must extract all styles from the app at build time (i.e., no streaming of styles). This can be done without a compiler by using `StyleSheet.getSheet()` after evaling all styled components.
  * #2196
* Supports use with multiple windows (i.e., iframes) and shadow roots.
  * #2138
  * #1517
* Supports 3rd party compilers and extraction to static CSS.
* Fixes static and dynamic short/longform deduplication
  * #2007
* Reduces browser support: Safari 10.1+, Chromium Edge, no IE, no legacy Android browsers.
* Removes automatic vendor-prefixing of inline styles (for better perf).
* Removes legacy `I18nManager.swapLeftAndRightInRTL()` and `I18nManager.doLeftAndRightSwapInRTL`.
* Removes focus-visible polyfill as modern browsers no longer show focus rings for mouse interactions.

Fix #2138
Fix #2196
Fix #2210
Fix #2007
Fix #1517

## Todo

- [x] Revisit focus-visible polyfill.
- [x] Fix SSR and add reliable SSR unit tests
- [x] Add link style reset to View
- [ ] Revisit RTL polyfill format to better suit build-time compilers.
- [x] Move multi-root StyleSheet logic to `render`.
- [ ] Update localization docs.
- [x] Add deprecation warning to setNativeProps.


natew:
Just a small piece of input (and not sure if it's already there) but having the object => stylesheet logic in a separate npm package would be nice to have, for tamagui at least.

necolas:
> having the object => stylesheet logic in a separate npm package would be nice to have

What's the "object => stylesheet logic"? And why would it be helpful in a separate package?

packages/benchmarks/src/implementations/styleq/Dot.js
```diff
@@ -0,0 +1,27 @@
+import React from 'react';
+import { styleq } from 'styleq';
+
+const Dot = ({ size, x, y, children, color }) => {
+  const [className, inlineStyle] = styleq([
```
natew, line 5: This basically is the function that would be useful. Tamagui at build time would use this to pre-generate the style object or className, so would want access to it.

necolas: It's on npm. But `StyleSheet` wraps this function, so you could call `StyleSheet(styles)` to get basically the same result plus logic for RTL polyfill. The output of the RNW compiler could be adjusted to match what a build-time compiler would be able to produce for RTL polyfilling, but I don't know if you're supporting that.

necolas: The RN API is not great for extensions, but it could be useful to have something like `StyleSheet/runtime` that only exports the code needed to merge the "css" objects (assuming a compiler doesn't leave `StyleSheet.flatten()` calls in the bundle)

necolas:
> Is this going to be a requirement with React 18?

Inserting during render can be problematic for initial renders because React 18 will yield back to the browser many times before updating the DOM. If styles are inserted during a render call, the browser might be forced to update layout during the pause. The `useInsertionEffect` hook is meant to mitigate this, but it's only available in React 18 and complicates the styling API significantly – essentially, the styling API either needs to be turned into a hook or needs to return a function to be called by the insertion hook. Instead, I'm hoping we can just insert everything during eval to avoid the render perf issue caused by React 18. Eval insertions are basically batched and lazy-loaded evals shouldn't be taking place during perf-sensitive user actions either. Ultimately, the "insert on create" approach plays better with build-time extraction strategies too, which is what React 18 wants. The React Native API design makes it very difficult to extract all styles, due to the existence of `StyleSheet.flatten()` and its use in `Animated`. But a web-first app could ban these APIs, allowing all styles to be extracted to a stylesheet

paularmstrong:
At quick glance, this will fully break runtime theming like [react-native-stylex](https://github.com/retyui/react-native-stylex) and similar approaches (eg, Twitter), since the theme comes through React context and needs to be built at runtime. 

Are you imagining that the only supported way to do theming will be to apply theme-based style values via the `style` attribute on `View` & `Text`?

necolas:
AFAIK, Twitter's runtime theming works [like this](https://gist.github.com/necolas/5b421ca860ed98eabc5fd2b9bc6d1136) to create new objects when the global theme changes. Each render the object getter returns the theme-specific style object. So that should still work fine for runtime, and isn't a regression in terms of ability to extract styles. But future approaches to theming would be better off relying on CSS variables for runtime value changes, rather than creating new style objects.

paularmstrong:
Mostly looking good, except SSR.

> SSR must extract all styles from the app at build time (i.e., no streaming of styles). This can be done without a compiler by using StyleSheet.getSheet() after evaling all styled components.

Do you have an example of using `StyleSheet.getSheet()`? I've run it after eval but it seems to just return an empty content: `{ id: 'react-native-stylesheet', textContent: '' }`

Alternatively, are you leaving it up to the community to figure out how to deal with this at build time?

necolas:
New build `0.0.0-88ea3411` should fix SSR. I've added dedicated node tests to more reliably test SRR features of StyleSheet and AppRegistry. LMK if still having problems.

> Will this be deprecating (or removing?) getStyleElement() returned from AppRegistry.getApplication()?

I was thinking we'll keep `AppRegistry.getApplication()` as is to make upgrading easy. It just won't return the styles that are a function of rendering the App, instead it should return all the styles from eval-ed modules.

paularmstrong:
All appears to be working except for an issue with box-shadows:

* 0.17.0: https://codesandbox.io/s/clever-dhawan-ct3d7?file=/src/Card.js
* 0.0.0-88ea3411: https://codesandbox.io/s/epic-marco-s367d?file=/src/Card.js

It seems that possibly it's the `shadowOffset` and `shadowOpacity` are not being applied.

necolas:
> All appears to be working except for an issue with box-shadows

Good spot. There's a bug producing a different box shadow for each of the RN properties and it isn't caught by the existing unit tests.

necolas:
@paularmstrong Should be fixed in `0.0.0-b096cd2d`: https://codesandbox.io/s/epic-pascal-t86z3

CooCooCaCha:
This is exciting! Does this styling system also work for react-native?

gut4:
Is it possible to extract styles from StyleSheet into ```*.css``` file(s)?
It will be useful with REMIX: https://remix.run/docs/en/v1/guides/styling

jgornick:
> Supports 3rd party compilers and extraction to static CSS.

@necolas (or someone else), can you expand on this, specifically **extraction to static CSS**? What makes these changes easier to extract `StyleSheet` usage to CSS?

Thank you!

necolas:
See #2248 for the 0.18 integration branch

raviolpula:
Hi @necolas, Regarding this comment 
`Reduces browser support: Safari 10.1+, Chromium Edge, no IE, no legacy Android browsers. `

We just updated from 0.17.7 to 0.21.1 and our app targeting Tizen 2017 broke. One breaking change is usage of `getRootNode()`. We fixed that with polyfill. But observing more style/layout issues. We would like to know more insights on what are the breaking changes for old browsers?

Thank you very much.

PR #2209, opened 2022-01-28
benchmarks: add Griffel

layershifter:
This PR adds [Griffel](https://github.com/microsoft/griffel/tree/main/packages/react) to CSS-in-JS competition.

P.S. @necolas you suggested, here it comes 🐱


necolas:
I meant you could add to the benchmarks locally if you wanted to do your own testing. Sorry for the confusion. Going forward the only external library I'll be testing against is the baseline of css-modules.

PR #2216, opened 2022-02-02
exposed type prop

MD-REHMAN:
Exposing the type prop. Allowing a way to set Input type (to _number, date, time_ or ....)

### DIFF
--- a/packages/docs/src/pages/docs/components/text-input.md
+++ b/packages/docs/src/pages/docs/components/text-input.md
@@ -139,7 +139,6 @@ Callback that is called when the text input's selection changes.
 Callback that is called when the keyboard's submit button is pressed. When `multiline={true}`, this is only called if `blurOnSubmit={true}`.
 {% endcall %}
 
-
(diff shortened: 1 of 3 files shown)


MD-REHMAN:
Hi @NE-SmallTown, 
I understand it can be a lot to go through every PRs. So let me simplify this for you.

## What this PR is doing?
Users can pass `type` prop on TextInput, which is analogous to Input's type attributes. Which allow the user to access different types of Inputs.

## Why it's required?
- Currently not able to use various types of input available on the web (like date, number, color and so on).
- https://stackoverflow.com/questions/70855116/how-do-you-do-input-type-number-in-native-base/71121683#71121683

## What does this PR do?
- Simply expose the `type` prop to the user. And it'll only change TextInput behaviour when `type` is passed, not changing any of the existing flow.
- Also docs and typing for the same.


I hope this simplifies everything. And do let me know if the PR still requires any changes.

necolas:
React Native doesn't support these types of input, and those it does support are available by using the `keyboardType` prop.

PR #2237, opened 2022-02-25
Git Ignore .idea Folder

DavidRieman:
Git ignore local `.idea` folders which are created by IDEA-based IDEs like WebStorm when opening the whole `react-native-web` folder as a project.

Recommend to merge this to avoid anyone accidentally trying to git add the `.idea` folder which is a user-specific area that should not be under source control.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,4 @@
+.idea
 .next
 build
 coverage


necolas:
You should add the ignore (and any other editor-specific files) to your machine's global git ignore instead. Thanks

DavidRieman:
Huh.  Not a bad idea to do as well, but is there an alternative to this [standard practice](https://github.com/facebook/react-native/blob/main/.gitignore#L57) in also helping out potential contributors automatically on this front?

PR #2241, opened 2022-02-25
Updated VirtualizedList vendor files (with FlatList, SectionList)

DavidRieman:
Per discussion at https://github.com/necolas/react-native-web/discussions/2236, this is a direct upgrade of a handful of select `react-native-web` `/src/vendor/react-native` components to reflect all the accumulated bug fixes and improvements made to their `react-native` roots since they were last updated some time last year.  I have tested these changes with some new `yarn examples` pages which are up for review separately.  (Most importantly, temporarily upgrading those samples to have hundreds of items used to produce major virtualization bugs which we can now see are solved with the latest versions of these files.)  The `yarn tests` also pass the same tests as before.

The upgrade process was simply done by 'diffing' the latest code from the react-native repo against the mainline here.  (Fortunately, I found WebStorm has a handy "Compare with Clipboard" feature which puts it into merge resolution mode.)  At first I was selective, but in the end with my final passes, I accepted the vast majority of changes from react-native for these, and even put the `import` order here into parity to make further diffs (and consequently, taking future bug fixes from react-native) much easier going forward.

In a couple small cases, I had to make up a type since we don't have the equivalent classes.  The import order parity should help identify any cases like that as well, as they'll be a type declaration where react-native used a type import.  Where I could, I tried to follow existing precedent for similar types in other react-native-web files.

There is one _dev-only_ feature which was new to VirtualizedList, which I intentionally avoided trying to add for now.  Trying to add the feature would simply add too much scope to a single PR.  Instead, I logged #2239 and cited that ticket at the disabled code.  I left the code just commented out there in order to call out the missing feature, and also to reduce scope of any future diffs that may happen to occur before that ticket can be tackled properly.

Issue #2068, opened 2021-06-24
Stylesheet.create incompatible return type


badsyntax:
**The problem**

The return type of RNW `StyleSheet.create` is incompatible with RN `StyleSheet.create`. This causes problems in css libraries where RN is aliased to RNW via https://www.npmjs.com/package/babel-plugin-react-native-web.

RNW implementation: 

https://github.com/necolas/react-native-web/blob/13f7b9e2f45af2761bfdccd763c511992c0be030/packages/react-native-web/src/exports/StyleSheet/StyleSheet.js#L47-L57

RN implementation:

https://github.com/facebook/react-native/blob/36318b6776507ba31070f6c8a3493609238a1a98/Libraries/StyleSheet/StyleSheet.js#L359-L371

**How to reproduce**

```js
// RNW
StyleSheet.create({
  foo: {
    color: 'blue'
  }
})
// results in: { foo: 89 }


// RN
StyleSheet.create({
  foo: {
    color: 'blue'
  }
})
// results in: { foo: { color: 'blue' } }
```

Simplified test case: Not sure if this is useful: https://codesandbox.io/s/async-bush-kfpgo?file=/src/App.js

**Expected behavior**

I expect the return type to match RN.

**Environment (include versions). Did this work in previous versions?**

* React Native for Web (version): 0.17.0
* React (version): 16.14.0
* Browser: n/a

**Additional context**

- I believe this was changed in RN with this commit: https://github.com/facebook/react-native/commit/a8e3c7f5780516eb0297830632862484ad032c10
- https://github.com/emotion-js/emotion/issues/1543#issuecomment-867583957

necolas:
This has been discussed before and for now we keep the original RN return type as web optimisations are built around it

badsyntax:
Okie doke. I took a quick look at the required changes and I agree it seems like high effort to refactor things to align with react-native, which is unfortunate. 

Here's some related issues where this change was previously discussed:

- https://github.com/necolas/react-native-web/issues/1593
- https://github.com/necolas/react-native-web/pull/1466
- https://github.com/necolas/react-native-web/issues/1259 (Possibly related, not sure)

PR #2248, opened 2022-03-02
0.18 preview release

necolas:
# 0.18 preview

* Install: `npm i react-native-web@0.0.0-de1d3a7c`
* Try: [Canary sandbox](https://codesandbox.io/s/nostalgic-williams-01cguk)
* [0.18 Milestone](https://github.com/necolas/react-native-web/milestone/23)

## Breaking changes

* Browser support has been reduced: Safari 10.1+, Edge (Chromium), no IE, no legacy Android browser.
* `StyleSheet.create()` is now the identify function, returning the input style objects. Previously, it replaced styles with numeric ids. This improves compatibility with React Native, but accessing the returned style objects at runtime is still *not recommended* as it can prevent static extraction to CSS. (#2068)
* React 18 optimizations mean that styles are now inserted on module eval, not component render. Server-rendering can continue using `AppRegistry` to inline styles, but all styles in eval-ed modules will be included, rather than only the styles produced by the initial render. The `StyleSheet.getSheet()` API can also be used to retrieve the current style sheet as a CSS string. (#2196)
* `I18nManager` is now a mock API included only for compatibility with React Native. The ability to swap between left/right values for RTL writing directions has been removed entirely. See below for new localization APIs.
* Inline styles are no longer automatically vendor-prefixed. This improves inline style performance and discourages use of inline styles. This should have little effect in practice if your app is predominantly using `StyleSheet.create()`.
* The deprecated `accessible`, `accessibilityState`, and `accessibilityValue` props have been removed.
* The `:focus-visible` polyfill has been removed, as by default modern browsers no longer show focus rings for pointer interactions.
* VirtualizedList, FlatList, and SectionList vendor components have been updated to match latest React Native. (#2241) Thanks to @DavidRieman.
* The jest preset and snapshot serializer has been removed.
* The `unstable_createElement` API has changed. Children are no longer accepted as the 3rd argument, they must be included in the props as the 2nd argument. The 3rd argument is now an options object with a field to set the expected writing direction (`{ writingDirection }`).

## New features

* Localized LTR/RTL layouts can now be nesting to arbitrary depths simply by setting the `dir` or `lang` prop on a component.
* Localized LTR/RTL layouts can be switched dynamically at runtime.
* A new `useLocaleContext` API is exported to query a component's locale and writing direction for fine-grained control over localized layout.
* Rendering into multiple windows and shadow roots in now supported. (#1517, #2138)
* Extraction to static CSS and consumption of external CSS is now possible. The StyleSheet runtime can consume any styles that match an expected format. (See below for more details.)

## Fixes

* Fixes static and dynamic shortform / longform deduplication. (#2007)
* Display the focus ring on `CheckBox` and `Switch` elements. (#2242)

### StyleSheet runtime

`StyleSheet` has been rewritten to better separate the runtime from the compiler. The performance of the new merging runtime is on par with using css-modules. The core runtime is a standalone package called styleQ; refer to the [styleQ documentation](https://github.com/necolas/styleq) for more details.

image

The performance of inline styles has also been significantly improved, although `StyleSheet` still has to perform extra work to transform non-standard React Native styles and polyfill logical styles.

image

The `StyleSheet` import is now a standalone module. `StyleSheet` itself is a function that can be called to resolve styles to DOM `className` and `style` values, e.g.,

```js
const [ className, inlineStyle ] = StyleSheet([ styles.root, prop.styles ])
```

### StyleSheet extraction

One of the benefits behind the new architecture is that `StyleSheet` can now resolve compiled styles that conform to the expected format, whether they are produced by the built-in compiler or an external source.

Style compilers must produce annotated objects that are treated as class name maps, and classes are deduped based on whether they share the same key, e.g.,

```js
{
  $$css: true,
  display: 'display-class',
  opacity: 'opacity-class'
}
```

This is the first step towards allowing 3rd party tools to extract styles to static CSS files, and optionally dropping the client-side compiler, without needing to integrate deeply with the `StyleSheet` runtime or having to modify props/prop values in the source code, i.e., only the `StyleSheet.create()` call needs transforming by the compiler.

The existence of the `StyleSheet.flatten()` API can be a problem for build-time compiling of some styles. Components that flatten styles expect to be able to work with the source styles, and use the source values at runtime. A compiler that aims to extract all styles and have no runtime transform would need to disallow the use of `StyleSheet.flatten()` entirely (unfortunately it's also used by the Animated API internals).


marklawlor:
There appears to be an issue with `Animated` when mixing preprocessed styles and inline-styles.

https://codesandbox.io/s/crazy-chatelet-iq8uq4?file=/src/App.js

Error: `styleq: maxWidth typeof 500 is not "string". `

You can see on the root `View` that it has added the inline styles as classNames.

necolas:
See this note on problems for static extraction:

> The existence of the StyleSheet.flatten() API can be a problem for build-time compiling of some styles. Components that flatten styles expect to be able to work with the source styles, and use the source values at runtime. A compiler that aims to extract all styles and have no runtime transform would need to disallow the use of StyleSheet.flatten() entirely (unfortunately it's also used by the Animated API internals).

Possibly StyleSheet.flatten could be modified to keep compiled and dynamic styles separate. But at the moment no one is using static extraction of styles.

Unfortunately, the RN API isn't being designed for Web, and that both limits a lot of what we can do, as well as adds a lot of overhead in the web shim.

jmeistrich:
@marklawlor @necolas I attempted a fix to the `Animated` issue and it does seem to work. It basically pulls any styles with `$$css` out of the `AnimatedProps` and merges it back in when rendering the animated component. My approach is a bit hacky but maybe the concept makes sense?

https://github.com/jmeistrich/react-native-web/commit/0266076d0cbe2c15417eb4ada78bf4922093cf72

marcuzgabriel:
Using the latest react-native-web with react / react-dom version 18 or above causes error as ReactDOM.render is no longer supported. The error is reproducable by taking your sandbox example and changing the react and react-dom versions to 18.0.0 or above. The error is then logged in the console.

I do not see any issues in general but. Is this something that will be addressed?

I can create an issue / observation if this is the wrong forum

Skærmbillede 2022-05-18 kl  15 48 37

necolas:
> Using the latest react-native-web with react / react-dom version 18 or above causes error as ReactDOM.render is no longer supported.

It's a warning, not an error. I don't intend to switch to `createRoot` in this release, as React Native won't be supporting it for a while yet, and various APIs need changing (e.g., unmountComponentAtNode)

necolas:
@jmeistrich that patch doesn't look like it will work with more than one style. As I mentioned above, fundamentally this is a problem related to having `StyleSheet.flatten` as an API. Any code path that uses `StyleSheet.flatten` during render (like Animated does internally) will likely break if styles have been processed and extracted at build time. Even components like `Image` use flatten to provide compatibility with how React Native handles various styles on images (e.g., shadows, `tintColor`, etc., get turned into filters)...at least in those cases I can just drop those compatibility features. But fixing the issue with `Animated` would involve rewriting `Animated` to use Web Animations and dropping compatibility with the Animated APIs that cannot be implemented on web. @vincentriemer explored this at Meta, but I don't think either of us have time to work on it for React Native for Web.

necolas:
This branch was merged into master and 0.18 is now available on npm

necolas:
> There appears to be an issue with Animated when mixing preprocessed styles and inline-styles.

@marklawlor I may have found a workaround for this issue while looking into https://github.com/necolas/react-native-web/issues/2387. If you follow that issue I'll likely have a patch up and publish a test version for you to try in your e2e setup

PR #2254, opened 2022-03-08
[change] Remove backgroundClip prefix workaround

SukkaW:
The PR removes the workaround for `background-clip` when prefixing.

The fix has been submitted to `inline-style-prefixer` as well, see https://github.com/robinweser/inline-style-prefixer/pull/221

### DIFF
--- a/packages/react-native-web/src/modules/prefixStyles/static.js
+++ b/packages/react-native-web/src/modules/prefixStyles/static.js
@@ -1,4 +1,3 @@
-import backgroundClip from 'inline-style-prefixer/lib/plugins/backgroundClip';
 import crossFade from 'inline-style-prefixer/lib/plugins/crossFade';
 import cursor from 'inline-style-prefixer/lib/plugins/cursor';
 import filter from 'inline-style-prefixer/lib/plugins/filter';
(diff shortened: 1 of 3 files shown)


SukkaW:
@necolas Should I change the target branch to `0.18-dev`?

SukkaW:
@necolas ping?

necolas:
The upstream package hasn't fixed the issue so there's nothing to here at the moment

SukkaW:
> The upstream package hasn't fixed the issue so there's nothing to here at the moment

The change I made here has more sense than the original workaround: it uses `inline-style-prefixer`'s `prefixMap` feature.

packages/react-native-web/src/modules/prefixStyles/static.js
```diff
@@ -1,4 +1,3 @@
-import backgroundClip from 'inline-style-prefixer/lib/plugins/backgroundClip';
```
necolas, line 1: This file is generated from `./scripts` and shouldn't be manually altered

SukkaW:
https://github.com/robinweser/inline-style-prefixer/pull/221 has been merged by the upstream. Just wait for the upstream package bumping. In the meantime, I will close the PR (since it is quite outdated and has too many conflicts) and will open a new one.

PR #2262, opened 2022-03-22
Update accessibility.md

paour:
`accessibilityRole` is a string rather than boolean

Ideally it would be a list of literal values, but there are too many to iterate in this document.


necolas:
Fixed on master

Issue #1744, opened 2020-09-18
Open in new tab


Sharcoux:
Opening link in a new tab is something quite common on web nowadays. This use case should be handled by react-native-web.

Currently, the only solution is [described here](https://github.com/necolas/react-native-web/issues/162).

Can I propose [this solution](https://github.com/necolas/react-native-web/pull/1743) as an improvement to react-native-web?

imnotjames:
Duplicate of https://github.com/necolas/react-native-web/issues/1544 ?

Issue #1544, opened 2020-02-26
Open in new tab


jfrolich:
It would be great for `Linking` to provide an option to open in a new window/tab. (Or is there already a way to do this?)

steida:
It's impossible to open a new tab without a link with the target.

jfrolich:
Sure? I think it's possible with `window.open`?

https://stackoverflow.com/questions/14132122/open-url-in-new-window-with-javascript

steida:
Window, not tab. Remember it's restricted because of popups.

jfrolich:
> Window, not tab. Remember it's restricted because of popups.

You are right. I formulated it wrong, it should be open in new tab.

steida:
It's impossible. The user has to click to a link. Browsers require that.

jfrolich:
> It's impossible. The user has to click to a link. Browsers require that.

`window.open` second argument is target, if I use that with `"_blank"` it opens in a new tab (at least in Chrome)

Sharcoux:
Could [this](https://github.com/necolas/react-native-web/pull/1743) solve this issue?

JijoBose:
Can Linking open an External Browser when running **react-native-web** inside a **Mobile App** in its **WebView**?
For eg: A **react-native-web** PWA running inside a **React Native** App's **webview**

Code
```
  const openLink = async url => {
    const supported = await Linking.canOpenURL(url)
    if (supported) {
      await Linking.openURL(url)
    }
  }

```
The above code however redirects to a page within the WebView and not from an external Browser (eg, open with Chrome, etc)
Any workarounds for this issue?

avanwinkle:
Resurrecting this thread to ask why the implementation hard-coded `_blank` as the target, instead of the other proposal https://github.com/necolas/react-native-web/pull/1743 which allowed a parameter?

I ask because there are scenarios where the desired behavior is *not* to open in a new tab, but that is currently impossible due to the hard-coded `_blank`.

PR #2277, opened 2022-04-12
Allow explicit target for Linking.openURL(), default to "_blank"

avanwinkle:
This PR extends on the work started by @Sharcoux in PR #1743 and relates to issues #162, #1544, and #1744.

The desire of those tickets was to open `Linking.openURL(url)` in a new tab, and the solution merged by @necolas [in this commit](https://github.com/necolas/react-native-web/commit/b7b383098aac42609166da38808a65497b5469a5) hard-codes the `"_blank"` target to force a new tab. This is great for people who want new tabs, but inconvenient for people who want to use the same tab.

This PR reworks the method to accept a second argument, `Linking.openURL(url, target)` for an explicit target. Any value passed will be used as the target, _including_ `undefined`. The ability to pass undefined allows users to maintain the same browser tab, while arbitrary values allow... well, anything.

For backwards compatibility, if `Linking.openURL(url)` is called **without** a `target`, the default value of `"_blank"` will be passed. Therefore this change should have zero impact on any code already written, only users henceforth who explicitly pass a `target` will get the new behavior.

Also, tests added!

### DIFF
--- a/packages/react-native-web/src/exports/Linking/index.js
+++ b/packages/react-native-web/src/exports/Linking/index.js
@@ -66,12 +66,16 @@ class Linking {
 
   /**
    * Try to open the given url in a secure fashion. The method returns a Promise object.
+   * If a target is passed (including undefined) that target will be used, otherwise '_blank'.
    * If the url opens, the promise is resolved. If not, the promise is rejected.
-   * Dispatches the `onOpen` event if `url` is opened successfully
+   * Dispatches the `onOpen` event if `url` is opened successfully.
    */
-  openURL(url: string): Promise<Object | void> {
+  openURL(url: string, target?: string): Promise<Object | void> {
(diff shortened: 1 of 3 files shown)


packages/react-native-web/src/exports/Linking/index.js
```diff
@@ -85,13 +89,13 @@ class Linking {
...
 
-const open = (url) => {
+const open = (url, string) => {
   if (canUseDOM) {
     const urlToOpen = new URL(url, window.location).toString();
     if (urlToOpen.indexOf('tel:') === 0) {
       window.location = urlToOpen;
     } else {
-      window.open(urlToOpen, '_blank', 'noopener');
+      window.open(urlToOpen, string, 'noopener');
```
Maker-Mark, line 98: Finally! Thanks for adding this :)

PR #2285, opened 2022-05-11
fix: use null prototypes to prevent prototype pollution

matias-la:
I found a potential [prototype pollution vulnerability](https://raw.githubusercontent.com/HoLyVieR/prototype-pollution-nsec18/master/paper/JavaScript_prototype_pollution_attack_in_NodeJS.pdf) while reading the code of the package. I'm not sure if it has any security impact, but I think it's better to fix it anyway.

The issue could be triggered if some cache property is named `__proto__`. In that case `cache[property][value] = object` would be creating properties in the global Object prototype. This could cause lots of unexpected things in the application.

### DIFF
--- a/packages/react-native-web/src/exports/StyleSheet/compile.js
+++ b/packages/react-native-web/src/exports/StyleSheet/compile.js
@@ -29,7 +29,7 @@ const cache = {
   get(property, value) {
     if (
       cache[property] != null &&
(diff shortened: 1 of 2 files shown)


necolas:
Closing as StyleSheet has been refactored several times and doesn't use this code anymore.

PR #2299, opened 2022-06-08
fix: Fix Touchable with href is still clickable if disabled

### LINKED ISSUE #2298: Touchable with href can still be clicked when disabled is true
**The problem**

Hello, I recently found out the `props.disable` of Touchable will not prevent it from clicking if `props.href` is provided, although it prevents the onPress event to be fired, but still will be navigated to the distination.

**How to reproduce**

Simplified test case: https://codesandbox.io/s/cocky-smoke-pl5kog

Steps to reproduce:
1. render a `TouchableOpacity` or `TouchableHighlight` with `props.href="/"` and `props.disabled={true}`
2. click on the touchable

**Expected behavior**

Should not be navigated to "/" since the touchable is disabled

* React Native for Web (version): TBC
* React (version): TBC
* Browser: TBC

rnike:
fix #2298

According to the implementation of `View`

https://github.com/necolas/react-native-web/blob/dc9ecfbc84c7e6cbf0c36c537ac1b0a86b7e43c1/packages/react-native-web/src/exports/View/index.js#L106-L107

`div` will be changed to `a` if `props.href` is provided, the `a` will be clickable no matter if `props.disabled` is true or not.

`TouchableHighlight` and `TouchableOpacity` are both using `View` as a base component

https://github.com/necolas/react-native-web/blob/dc9ecfbc84c7e6cbf0c36c537ac1b0a86b7e43c1/packages/react-native-web/src/exports/TouchableHighlight/index.js#L164-L165
https://github.com/necolas/react-native-web/blob/dc9ecfbc84c7e6cbf0c36c537ac1b0a86b7e43c1/packages/react-native-web/src/exports/TouchableOpacity/index.js#L123-L124

This PR will add `pointerEvents="none"` to the component if `props.disabled` is `true` in `TouchableHighlight` and `TouchableOpacity` to make them actually disabled.

Tested on the example app and also our app.

### DIFF
--- a/packages/react-native-web/src/exports/Button/__tests__/__snapshots__/index-test.js.snap
+++ b/packages/react-native-web/src/exports/Button/__tests__/__snapshots__/index-test.js.snap
@@ -32,7 +32,7 @@ exports[`components/Button prop "color" 1`] = `
 exports[`components/Button prop "disabled" 1`] = `
 <div
   aria-disabled="true"
-  class="css-view-1dbjc4n r-backgroundColor-11mpjr4 r-borderRadius-1jkafct r-transitionProperty-1i6wzkk r-userSelect-lrvibr"
+  class="css-view-1dbjc4n r-backgroundColor-11mpjr4 r-borderRadius-1jkafct r-pointerEvents-633pao r-transitionProperty-1i6wzkk r-userSelect-lrvibr"
   role="button"
   style="transition-duration: 0s;"
   tabindex="-1"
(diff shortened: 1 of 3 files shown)


rnike:
`yarn jest` complete successfully

necolas:
The issue is because the href is still followed by the browser, so ideally there'd be no href set when the element is disabled. However, that does either require each Touchable/Pressable component to put some conditional logic around `href`, or for View and Text to understand the concept of being disabled so that they can avoid setting `href`.

If we go with the `pointerEvents` solution, is the intent of your patch that the `pointerEvents` value could be overridden? I would have assumed we want it enforced, in which case we should place `pointerEvents={disabled ? 'none' : null}` after the `{...rest}` spread.

rnike:
@necolas I'm letting `pointerEvents` be overridable in purpose because forcing it to none if disabled is going to change these components' behaviour. For example, someone who is currently using `TouchableOpacity` with `pointerEvents` will find it working otherwise after having this upgrade. But thinking it through again, this seems to be an over-design because after all `pointerEvents` is not available for Touchables, I'm going to change it back and thanks for the review.

necolas:
This patch is included in the latest canary release: react-native-web@0.0.0-de1d3a7c

PR #2315, opened 2022-06-27
[fix] Modal should pass through View props

### LINKED ISSUE #2314: Modals do not pass through accessibility* props
**The problem**

When using a `Modal`, I have a requirement to include at least a `aria-labelledBy` prop that references the title of the dialog. However, no accessibility* props appear to be carried over to the `role="dialog"` element that the `<Modal />` produces. More info on dialog role: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/dialog_role

**How to reproduce**

Simplified test case: https://codesandbox.io/s/react-native-web-modal-does-not-include-a11y-props-h26vs6?file=/src/App.js

Steps to reproduce:
1. Add `<Modal accessibilityLabelledBy="some-id">...</Modal>`
2. Render

**Expected behavior**

`aria-*` props should be added as provided by `accessibility*` props

* React Native for Web (version): ^0.17.0, ^0.18.1
* React (version): 17, 18
* Browser: any

necolas:
Fix #2314

### DIFF
--- a/packages/react-native-web/src/exports/Modal/ModalContent.js
+++ b/packages/react-native-web/src/exports/Modal/ModalContent.js
@@ -8,24 +8,27 @@
  * @flow
  */
 
+import type { ViewProps } from '../View';
+
 import * as React from 'react';
 import View from '../View';
 import StyleSheet from '../StyleSheet';
 import ExecutionEnvironment from 'fbjs/lib/ExecutionEnvironment';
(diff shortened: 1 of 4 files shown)


paularmstrong:
lgtm!

necolas:
Fixed by https://github.com/necolas/react-native-web/commit/d9fd6a942f81c80b7ed14cd802b04b7584b5ae6f

Issue #2296, opened 2022-06-05
Pressable suppression of contextmenu on Windows


necolas:
**The problem**

Long-press with a touch on a Pressable does not suppress the contextmenu on Windows. On iOS and Android, the contextmenu is triggered while the pointer is down, after a certain length of time. However, on Windows, the contextmenu is displayed after pointerup (with the same timestamp as pointerup), whether triggered by a mouse right-click or a touch long-press. 

**How to reproduce**

Simplified test case: https://6r9liu.sse.codesandbox.io/pressable (existing Pressable example)

Steps to reproduce:
1. Open test case on Windows
2. Use Responsive Design mode in Chrome dev tools and enable touch emulation.
3. Long press the Pressable and then release the pointer.
4. The contextmenu will display (with a type of PointerEvent, pointerType of `touch`, button value of `-1`, and the same timestamp as `pointerup`).

**Expected behavior**

No contextmenu should display if listening to longpress interaction. This can probably be fixed simply by calling preventDefault on all contextmenu events with a button value of `-1` and/or `pointerType` of `"touch"` when `onLongPress` is defined.

**Environment (include versions). Did this work in previous versions?**

* React Native for Web (version): 0.17 (all)
* Browser: Browsers on Windows

PR #2316, opened 2022-06-28
[fix] Pressable suppression of contextmenu on windows 

yuxizhe:
Fix https://github.com/necolas/react-native-web/issues/2296
Long-press with a touch on a Pressable does not suppress the contextmenu on Windows. On iOS and Android, the contextmenu is triggered while the pointer is down, after a certain length of time. However, on Windows, the contextmenu is displayed after pointerup (with the same timestamp as pointerup), whether triggered by a mouse right-click or a touch long-press.

before:
Jun-28-2022 15-48-53

after:
Jun-28-2022 15-50-21

### DIFF
--- a/packages/examples/package.json
+++ b/packages/examples/package.json
@@ -11,12 +11,13 @@
     "next": "^12.1.0",
     "react": "^17.0.2",
     "react-dom": "^17.0.2",
-    "react-native-web": "0.18.2"
+    "react-native-web": "0.18.2",
+    "babel-plugin-react-native-web": "0.18.2"
   },
   "devDependencies": {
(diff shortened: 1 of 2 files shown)


yuxizhe:
> I noticed codesandbox ci didn't replace the version for `babel-plugin-react-native-web` in the package.json: https://codesandbox.io/s/examples-fdt05h?file=/package.json

Yes, it seems we have to move `babel-plugin-react-native-web` to `dependencies` instead of `devDependencies`.

this is my test: https://codesandbox.io/s/examples-rbzfou?file=/package.json

PR #2324, opened 2022-07-01
Monorepo & Babel

necolas:
* Reorganize monorepo inc README and LICENSE files
* Use `@babel/runtime` to de-duplicate helpers in compiled output.
* Switch to using npm for monorepo task management, e.g., `npm run build -w react-native-web`


magsout:
@necolas why this change from yarn to npm

Issue #1529, opened 2020-01-25
Compatibility with createRoot for latest concurrent mode


SamyPesse:
Using `react-native-web` relies on `AppRegistry.runApplication` which doesn't use `ReactDOM.createRoot`. It means that using suspense with `react-native-web` is only possible using the previous Suspense API (`<React.unstable_ConcurrentMode>`), since then the React had released [great new concurrent APIs such as `useTransition`, or `useDeferredValue`](https://reactjs.org/docs/concurrent-mode-reference.html).

Supporting concurrent mode with `react-native-web` will require:
- Replace uses of deprecated React API (legacy context https://github.com/necolas/react-native-web/issues/1506, findDOMNode) 
- Use `ReactDOM.createRoot` in [`renderApplication`](https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/AppRegistry/renderApplication.js)

I think a potential solution in the short term could be to use `ReactDOM.createRoot` when an option is passed to `AppRegistry.runApplication` so that it supports both latest and older react versions.

I'd be happy to open a PR for such changes if you agree with such implementation :)

necolas:
I'll update the implementation once a stable release of concurrent mode is available

hosseinmd:
This is possible if we change renderApplication like this
```js
function renderApplication<Props: Object>(
  RootComponent: ComponentType<Props>,
  WrapperComponent?: ?ComponentType<*>,
  callback?: () => void,
  options: {
    initialProps: Props,
    rootTag: any,
    mode: 'Concurrent' | 'Blocking' | 'Legacy',
  },
) {
  const {initialProps, rootTag, mode} = options;

  invariant(rootTag, 'Expect to have a valid rootTag, instead got ', rootTag);

  const App = (
    <AppContainer rootTag={rootTag} WrapperComponent={WrapperComponent}>
      <RootComponent {...initialProps} />
    </AppContainer>
  );

  if (mode === 'Concurrent') {
    ReactDOM.createRoot(rootTag).render(App, callback);
  } else if (mode === 'Blocking') {
    ReactDOM.createBlockingRoot(rootTag).render(App, callback);
  } else {
    ReactDOM.render(App, rootTag, callback);
  }
}
```

hosseinmd:
I want to test react experimental on react-native-web this code help me.
could you release that in an experimental like react to test that by anyone who is curious.

hosseinmd:
[...]

Issue #1506, opened 2019-12-18
Remove all use of legacy context API


necolas:
**The problem**

The legacy context API is being used in a number of components. This API will eventually be deprecated and there are performance improvements to using the current context API.

sergeymild:
I am still seeing that problem
```
Warning: Legacy context API has been detected within a strict-mode tree.

The old API will be supported in all 16.x releases, but applications using it should migrate to the new version.

Please update the following components: CellRenderer

Learn more about this warning here: https://reactjs.org/link/legacy-context
    at CellRenderer (http://localhost:3002/static/js/0.chunk.js:92083:36)
    at div
    at View (http://localhost:3002/static/js/0.chunk.js:76732:24)
    at div
    at View (http://localhost:3002/static/js/0.chunk.js:76732:24)
    at http://localhost:3002/static/js/0.chunk.js:70871:34
    at ScrollView (http://localhost:3002/static/js/0.chunk.js:17993:37)
    at VirtualizedList (http://localhost:3002/static/js/0.chunk.js:90941:34)
    at FlatList (http://localhost:3002/static/js/0.chunk.js:88627:34)
    at div
    at View (http://localhost:3002/static/js/0.chunk.js:76732:24)
    at ProfileScreen (http://localhost:3002/static/js/0.chunk.js:24876:76)
    at StaticContainer (http://localhost:3002/static/js/0.chunk.js:2987:16)
    at EnsureSingleNavigator (http://localhost:3002/static/js/0.chunk.js:2697:3)
    at SceneView (http://localhost:3002/static/js/0.chunk.js:2884:3)
    at div
    at View (http://localhost:3002/static/js/0.chunk.js:76732:24)
    at div
    at View (http://localhost:3002/static/js/0.chunk.js:76732:24)
    at div
    at View (http://localhost:3002/static/js/0.chunk.js:76732:24)
    at CardSheet (http://localhost:3002/static/js/0.chunk.js:12009:3)
    at div
    at View (http://localhost:3002/static/js/0.chunk.js:76732:24)
    at AnimatedComponent (http://localhost:3002/static/js/0.chunk.js:85088:32)
    at AnimatedComponentWrapper
    at Dummy (http://localhost:3002/static/js/0.chunk.js:9907:3)
    at div
    at View (http://localhost:3002/static/js/0.chunk.js:76732:24)
    at AnimatedComponent (http://localhost:3002/static/js/0.chunk.js:85088:32)
[...]

PR #2330, opened 2022-07-06
React 18 concurrent render functions / removal of findDOMNode / test updates

edkimmel:
Looking to resolve the issues pointed out by https://github.com/necolas/react-native-web/issues/1529

The breaking changes come from the deprecation / removal of findNodeHandle, which doesn't appear to be a strictly necessary change to support React 18.  Let me know if I should revert those changes.

- Bump React dependency to 18.

- Remove uses of `findNodeHandle`.  Remove direct uses of `findDOMNode`.  Marked `findNodeHandle` as deprecated for uses outside of this library
  - TouchableMixin has a new required function on the user - `getTouchableNode` - which returns the dom node of the touchable element.
  -  ScrollResponderMixin no longer has the imperative `findNodeHandle` fallback, the dom node must be provided by `getScrollableNode`
- Support for concurrent mode in `AppRegistry`
  - This required swapping to a callback ref on `AppContainer `as the callback parameter to the new render/hydrate calls was removed.
- Using react testing library's render and act functions in tests 
  - Using the legacy render functions caused a React deprecation warning on every test
  - Using the new render functions directly caused this warning: https://github.com/testing-library/react-testing-library/issues/1061

### DIFF
--- a/packages/react-native-web/package.json
+++ b/packages/react-native-web/package.json
@@ -31,8 +31,8 @@
     "styleq": "^0.1.2"
   },
   "peerDependencies": {
-    "react": "^17.0.2 || ^18.0.0-0",
-    "react-dom": "^17.0.2 || ^18.0.0-0"
+    "react": "^18.0.0",
+    "react-dom": "^18.0.0"
   },
   "author": "Nicolas Gallagher",
   "license": "MIT",
--- a/packages/react-native-web/package.json
+++ b/packages/react-native-web/package.json
@@ -3,7 +3,7 @@
     "registry": "https://registry.npmjs.org/"
   },
   "name": "react-native-web",
-  "version": "0.19.0",
+  "version": "0.18.4",
   "description": "React Native for Web",
   "module": "dist/index.js",
   "main": "dist/cjs/index.js",
--- a/packages/react-native-web/package.json
+++ b/packages/react-native-web/package.json
@@ -3,7 +3,7 @@
     "registry": "https://registry.npmjs.org/"
   },
   "name": "react-native-web",
-  "version": "0.18.9",
+  "version": "0.19.0",
   "description": "React Native for Web",
   "module": "dist/index.js",
   "main": "dist/cjs/index.js",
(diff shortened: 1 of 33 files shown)


packages/react-native-web/src/exports/Touchable/index.js
```diff
@@ -73,6 +72,11 @@ const extractSingleTouch = (nativeEvent) => {
  *     return merge(this.touchableGetInitialState(), yourComponentState);
  *   }
  *
+ * - Add a method to get your touchable component's node.
+ *   getTouchableNode: function() {
+ *     return this.touchableRef.current
+ *   }
```
necolas, line 78: Where is this used / implemented?

packages/react-native-web/package.json
```diff
@@ -3,7 +3,7 @@
     "registry": "https://registry.npmjs.org/"
   },
   "name": "react-native-web",
-  "version": "0.18.4",
+  "version": "0.19.0",
```
necolas, line 6: No package version change please

necolas (commented):
Thanks! This looks pretty good

necolas:
The `unmountComponentAtNode` export probably also needs rethinking for the new root API.

packages/react-native-web/src/exports/Touchable/index.js
edkimmel: Guidance from React is to remove `findDOMNode` and use React.Ref instead.  The `Touchable` and `ScrollResponder` mixins don't render themselves, so they become dependent on the concrete component to capture the root to a ref and provide it.  

I didn't see any uses of `Touchable` directly in this repository, so that seems to be only a user-facing breaking change.  Nothing in this repository actually implements it.

`ScrollResponderMixin` was used by `ScrollView`, which already provided `getScrollableNode`.  It appears that it would only be user components using `ScrollResponderMixin` that would have to be sure to implement `getScrollableNode` now.

edkimmel:
Removing `findDOMNode` is a step towards react-native-web working under strict mode.

The other two I can think of are:
1) https://github.com/necolas/react-native-web/issues/1506 - legacy context APIs
2) Uses of `UNSAFE_componentWillMount` and `UNSAFE_componentWillReceiveProps`

packages/react-native-web/src/exports/Touchable/index.js
necolas: I think after this is merged, it will also be a good time to remove those mixins as a legacy feature that will not carry forward with React 18 adoption.

edkimmel:
> Do you know on which update this will be fixed ? Thank you !

@JinbeiStudio This is part of 0.19 https://github.com/necolas/react-native-web/pull/2377

 I don't know when that release is expected to go out.

MalgoBlock:
Are there any plans for this to be released soon? Thank you!

PR #2332, opened 2022-07-06
Fix centerContent on ScrollView

### LINKED ISSUE #2331: `ScrollView`'s `centerContent` prop doesn't work

### Describe the issue

As always, thanks for the incredible work of RNW. Onto the issue...

`<ScrollView centerContent={true} />` doesn't center the content. Here is a comparison of iOS and Web:

### iOS

Screen Shot 2022-07-06 at 5 04 19 PM

### Web

Screen Shot 2022-07-06 at 5 04 35 PM

### Expected behavior

`centerContent` should align content to the center.

### Steps to reproduce

1. Add a `<ScrollView centerContent>`
2. Put a view inside of it
3. Compare native and Web behavior

### Test case

https://snack.expo.dev/@beatgig/rnw-centercontent-broken

### Potential solution

Adjusting `contentContainerStyle` seems to have fixed it:

```tsx
import { View, StyleSheet, ScrollView } from 'react-native';

  return (
    <ScrollView centerContent contentContainerStyle={styles.fix}>
      <View style={{ height: 50, backgroundColor: 'blue' }} />
    </ScrollView>
  );
}

  fix: {
    justifyContent: 'center',
    flexGrow: 1,
  },
});
```

Screen Shot 2022-07-06 at 5 05 12 PM

For additional context: if I didn't add `flexGrow: 1`, then the content wasn't properly scrollable when it overflowed the size of its container.

If that's the correct solution, please let me know and I'm happy to submit a PR. Thanks!

nandorojo:
Closes #2331.

I tried to naïvely add a unit test, but I don't have experience writing them, so I'm clearly missing something.

I attempted this:

```js
  describe('prop "centerContent"', () => {
    test('without', () => {
      const { container } = render(
        <ScrollView style={{ backgroundColor: 'blue' }} />
      );
      expect(container.firstChild).toMatchSnapshot();
    });

    test('with', () => {
      const { container } = render(
        <ScrollView centerContent style={{ backgroundColor: 'blue' }} />
      );

      expect(container.firstChild).toMatchSnapshot();
      expect(container.firstChild).toHaveStyle('justify-content', 'center');
      expect(container.firstChild).toHaveStyle('flex-grow', '1');
    });
  });
```

However, it seems that `toHaveStyle` doesn't exist. I came across that API from searching online for `jest-dom` + `@testing-library/react`, but I wasn't really sure where to go from there.

I've tested it in my app and it's working well for me. Happy to add/update a unit test if you have any guidance for updating my broken test.

### DIFF
--- a/packages/react-native-web/src/exports/ScrollView/index.js
+++ b/packages/react-native-web/src/exports/ScrollView/index.js
@@ -31,7 +31,8 @@ type ScrollViewProps = {
   refreshControl?: any,
   scrollEnabled?: boolean,
   scrollEventThrottle?: number,
-  stickyHeaderIndices?: Array<number>
+  stickyHeaderIndices?: Array<number>,
+  centerContent?: boolean
 };
 
 const emptyObject = {};
@@ -136,6 +137,7 @@ const ScrollView = ((createReactClass({
       forwardedRef,
       keyboardDismissMode,
       onScroll,
+      centerContent,
       /* eslint-enable */
       ...other
     } = this.props;
@@ -189,10 +191,11 @@ const ScrollView = ((createReactClass({
         children={children}
         collapsable={false}
         ref={this._setInnerViewRef}
-        style={StyleSheet.compose(
+        style={[
           horizontal && styles.contentContainerHorizontal,
+          centerContent && styles.contentContainerCenterContent,
           contentContainerStyle
-        )}
+        ]}
       />
     );
 
@@ -328,6 +331,10 @@ const styles = StyleSheet.create({
   },
   contentContainerHorizontal: {
     flexDirection: 'row'
+  },
+  contentContainerCenterContent: {
+    justifyContent: 'center',
+    flexGrow: 1
   },
   stickyHeader: {
     position: 'sticky',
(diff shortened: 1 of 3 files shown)


packages/react-native-web/src/exports/ScrollView/index.js
```diff
@@ -189,10 +191,11 @@ const ScrollView = ((createReactClass({
         children={children}
         collapsable={false}
         ref={this._setInnerViewRef}
-        style={StyleSheet.compose(
+        style={[
```
nandorojo, line 194: Every example of `StyleSheet.compose` throughout the repo had only two styles passed to it. Since this one now has three, I turned it into an array. If that's incorrect, feel free to let me know.

burakgormek:
IMO for the test, toMatchSnapshot is enough. It will add classes to div and will compare output to match snapshot. No needed to check with toHaveStyle.

But you have to run test to create initial snapshot.

nandorojo:
> IMO for the test, toMatchSnapshot is enough. It will add classes to div and will compare output to match snapshot. No needed to check with toHaveStyle.
> 
> But you have to run test to create initial snapshot.

Thanks for the feedback. Didn’t realize it was that trivial. I pushed the test without `toHaveStyle`.

PR #2335, opened 2022-07-08
Allow passing className and id props without exposing them

natew:
Just a thought, open to alternative routes. 

These small changes would allow me to totally remove the last patch I have in place for react-native-web vis-vis Tamagui.

The ID one could likely just go, I remember needing it for some strange reason but looking now seems like form elements all take nativeID already.


necolas:
`nativeID` should always to produce `id`.

`className` is supported via the `style` API as described in the [0.18 release notes](https://github.com/necolas/react-native-web/releases/tag/0.18.0)  and in more detail in the [styleQ docs](https://github.com/necolas/styleq#notes-for-compiler-authors)

necolas:
Closing as no follow up and mechanism to achieve the same results has been described

Issue #2333, opened 2022-07-06
Remove fbjs dependency from 'react-native-web'


necolas:
### Is there an existing request?

- [X] I have searched for this request

### Describe the feature request

- [ ] Replace 'fbjs/lib/ExecutionEnvironment' with [inlined](https://unpkg.com/browse/fbjs@3.0.4/lib/ExecutionEnvironment.js) 'modules/canUseDOM'.
- [ ] Replace 'fbjs/lib/invariant' with in-place errors.
- [ ] Replace 'fbjs/lib/performanceNow' with `performance.now`.

PR #2336, opened 2022-07-08
fix: replaces fbjs canUseDom with inline module

crathor:
Fixes part of #2333 

This PR removes the `fbjs canUseDom` and replaces it with a local module `canUseDom`


necolas:
Thanks!

PR #2337, opened 2022-07-08
fix: replace fbjs performanceNow

kyletsang:
Addresses part of #2333.  Replaces fbjs's `performanceNow` with `global.performance.now`

### DIFF
--- a/packages/react-native-web/src/vendor/react-native/JSEventLoopWatchdog/index.js
+++ b/packages/react-native-web/src/vendor/react-native/JSEventLoopWatchdog/index.js
@@ -38,7 +38,7 @@ const JSEventLoopWatchdog = {
     totalStallTime = 0;
     stallCount = 0;
     longestStall = 0;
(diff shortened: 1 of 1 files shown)


packages/react-native-web/src/vendor/react-native/JSEventLoopWatchdog/index.js
```diff
@@ -39,7 +38,7 @@ const JSEventLoopWatchdog = {
     totalStallTime = 0;
     stallCount = 0;
     longestStall = 0;
-    lastInterval = performanceNow();
+    lastInterval = global.performance.now();
```
necolas, line 41: I'm not sure if `global` is reliable on web vs `window`.

kyletsang: I've replaced with `window` now. Thanks for catching that

necolas:
Thanks!

Issue #1807, opened 2020-11-11
Copy and Pasting text comes out backwards when using FlatList inverted


puneetlath:
**The problem**
When using the `inverted` prop with `FlatList` text is pasted in backwards order. For example if the list (after inversion) is: 
- 1
- 2
- 3
- 4

Copy and pasting it will result in: 
- 4
- 3
- 2
- 1

**How to reproduce**
Simplified test case: https://codesandbox.io/s/cold-wave-grzpj?file=/src/App.js

Steps to reproduce:
1. Create a list using `FlatList`
2. Apply the `inverted` prop
3. Note the order of some list items and copy them
4. Paste them and note that the order is backwards from what was displayed

**Expected behavior**
I would expect that when pasted the list would retain the order that is shown on the screen. 

**Environment (include versions). Did this work in previous versions?**

* React Native for Web (version): 0.13.18
* React (version): 16.14.0
* Browser: Chrome, Safari, Firefox

**Additional context**
The problem seems to have to do with the fact that when the `inverted` prop is applied, the html elements are still in the original order and the inversion is done via css `flex-direction: column-reversed`.

necolas:
VirtualizedList / FlatList comes directly from React Native, so you should raise this issue in that repo. But I don't know that there is any practical alternative to the existing approach to inversion, where the HTML order is preserved. There are also some flexbox issues in React Native that might be involved - https://github.com/facebook/react-native/pull/23632#issuecomment-528876788

mallenexpensify:
If anyone is able to fix this, we're willing to compensate. Issue posted https://github.com/Expensify/Expensify.cash/issues/1341

mallenexpensify:
We're still looking for help to "Fix Inverted prop on the FlatList component with a RN Patch". The current link is https://github.com/Expensify/Expensify.cash/issues/3381,the job amount is $8000. Any help/tips are greatly appreciated.

PR #2345, opened 2022-07-14
Fix copy and paste order bug created by inverted FlatList

AndrewGable:
This original PR was created by @azimgd [here](https://github.com/necolas/react-native-web/pull/2151) but never taken out of draft. In addition it contains additional fixes for Safari as mentioned here: https://github.com/Expensify/react-native-web/pull/3

> ## Expected Result:
> - When copying text in `<FlatList inverted={true} />` that when you highlight lines of text, including across multiple messages, that the text is highlighted in order from top to bottom or bottom to top.
> - When pasting text that has been copied from `Flatlist`, that it shows up in the correct order.
> 
> ## Actual Result:
> - Copying is sporadic and 'jumps around' when you highlight text over multiple messages.
> - Pasting - When pasting copied text from multiple e.cash messages, it's out of order
> 
> Highlighting glitch             |  Copy order 
> :-------------------------:|:-------------------------:
>  | 
> 
> This has been reviewed and tested by multiple developers. Some references are:
> - https://github.com/Expensify/App/issue/3381
> - https://github.com/Expensify/App/pull/5561
> - https://github.com/necolas/react-native-web/issues/1807
> - https://github.com/facebook/react-native/issues/30383
> 
> ## Implementation
> I have removed flatlist inversion (by inverted=true prop) which used [transform: translate] css hack and instead adding flatlist array items in reverse order [unshift instead of push]; It also reverses ScrollView event response received from browser [e.nativeEvent.configOffset *= -1];
> 
> ## Sidenote
> I was not able to fix this issue by applying this code directly in React Native repo. The reason is that RN column-reverse won't anchor scroll position to the bottom on Mobile device as it does on Web
> https://github.com/facebook/yoga/issues/866#issuecomment-797548962

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "private": true,
-  "version": "0.17.3",
+  "version": "0.17.4",
   "name": "monorepo",
   "scripts": {
     "clean": "del-cli ./packages/*/dist",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "private": true,
-  "version": "0.17.4",
+  "version": "0.17.5",
   "name": "monorepo",
   "scripts": {
     "clean": "del-cli ./packages/*/dist",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "private": true,
-  "version": "0.17.2",
+  "version": "0.17.3",
   "name": "monorepo",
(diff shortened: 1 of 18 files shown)


AndrewGable:
Looks like this branch is too out of date, I will update it in a new PR.

PR #2347, opened 2022-07-19
Update react peer dependency in react-native-web

brentvatne:
This helps react-native-web play nicely with npm 7+ and react-native@0.69 / react@18. We are working on releasing Expo SDK 46 which uses react-native@0.69.1 and react@18.0.0, and without this change users will have to use `--legacy-peer-deps` to initialize their projects: `EXPO_BETA=1 npx create-expo-app -- --legacy-peer-deps`


necolas:
Try 0.18.7

PR #2374, opened 2022-08-24
[fix] Allow Pressable to be programmatically focused when disabled

### LINKED ISSUE #2357: Setting a Pressable as "disabled" from enter key press locks its internal `focused` state as `true`

Might be a side effect of the fix for https://github.com/necolas/react-native-web/issues/2127

### Describe the issue

If a Pressable's `onPress` function causes it to become `disabled`, and this is triggered by tabbing in with the keyboard and triggering the `onPress` function from the keyboard using the Enter key, then the Pressable's internal `focused` state becomes locked as `true` even after the actual HTML element has blurred.

For example, here, the first Pressable was disabled and blurred using the Enter key and the second Pressable is the focused element, but the first Pressable thinks it is focused still.

image

### Expected behavior

A Pressable's `focused` state should not be `true` if it is not focused, regardless of how it came to be not focused, and it should never be both disabled and focused.

### Steps to reproduce

- Create a `Pressable` that shows its internal `focused` state visibly
- Set its `disabled` state to `true` in an `onPress` handler
- Select it using the keyboard tab key and trigger the onPress handler with the Enter key

### Test case

https://codesandbox.io/s/quirky-williamson-v3r94g
https://snack.expo.dev/@alansl/amused-cheese

### Additional comments

There's also a possibly-related issue described in comments in the linked example, where it's possible to programmatically focus a `disabled` Pressable just before `disabled` is removed, and then when `disabled` is removed, the element is focused but the focused state is stuck as `false`. Which issue occurs depends on whether mouse or keyboard events are used and timing of when the focus switch happens. Suggests the issue hinges on the timing of rerenders triggered by different event types.

In the provided example, it's _almost_ possible to work around this issue with careful timing - programmatically blurring the element after setting disabled to `false` in the target but before setting it to `true` in the trigger element. But this creates some strange timing issues when mouse clicks are used. Also, in real-life examples, this may not be an option: for example if the elements' `disabled` state is controlled by one `selected` state, it may not be possible to set a moment when both are enabled.

### Suggested fix

It looks like all focus / blur handling is suppressed as soon as `disabled` is applied, and the timing of this relative to event handler timing varies by event type. Suggestion:

 - When `disabled` flips from false to true, if `focused` is still true, it shouldn't be and the element is losing focus; so call `onBlur` (including setting internal state) while it still can.
 - When `disabled` flips from true to false, if the element has become actually focused in the DOM but `focused` is false, it is legitimately gaining focus but the event handlers were skipped due to event / re-render timing; so call `onFocus` (including setting internal state)

necolas:
Fix #2357

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -14922,7 +14922,7 @@
       }
     },
     "packages/babel-plugin-react-native-web": {
-      "version": "0.18.6",
+      "version": "0.18.7",
       "license": "MIT",
       "devDependencies": {
         "babel-plugin-tester": "^10.1.0"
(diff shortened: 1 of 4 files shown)


necolas:
codesandbox for this patch with the test case from the original bug: https://codesandbox.io/s/react-native-web-examples-forked-yb5p26

AlanSl:
Great, thanks, I've tried this out locally in my real-world component that had the initial problem, and this fixes it, means I can remove the workaround I'd previously used. And of course it works perfectly in the sandbox reproduction too.

LGTM and thank you!

PR #2375, opened 2022-08-24
[fix] CheckBox supports readOnly on native control

### LINKED ISSUE #2359: CheckBox does not apply readonly when using accessibilityReadOnly

### Describe the issue

`aria-readonly` / `readonly` attributes for checkboxes

### Expected behavior

using `<CheckBox accessibilityReadOnly />` should apply `readonly` attribute to the output `<input type="Checkbox" reaonly />` when using `, but it currently does not.

### Steps to reproduce

`<CheckBox accessibilityReadOnly />` 

See test case

### Test case

https://codesandbox.io/s/icy-fire-fr2ct4

### Additional comments

I'm uncertain if there are other accessible props that are also not translating through and/or if this is only related to `CheckBox`. It does appear to work correctly on `TextInput`, as shown in the test case link.

There may also be needs to have the colors either the same as a `disabled` checkbox, or muted in some other way.
Screen Shot 2022-08-09 at 10 02 03

necolas:
Fix #2359

### DIFF
--- a/packages/react-native-web-examples/pages/checkbox/index.js
+++ b/packages/react-native-web-examples/pages/checkbox/index.js
@@ -24,6 +24,8 @@ export default function CheckboxPage() {
         <CheckBox disabled style={styles.item} value={false} />
         <Divider />
         <CheckBox disabled style={styles.item} value={true} />
+        <Divider />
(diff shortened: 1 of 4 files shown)


necolas:
93b1d043068b701d08c0b06fc9127b71810ba3ff

PR #2377, opened 2022-08-24
0.19 dev

necolas:
Canary: `npm install react-native-web@0.0.0-47671ca4`

Best compatibility: `react-native@>=0.71`

## React 18 support

* [change] React 18 `createRoot` support.
* [change] Update `Animated` to support React 18, including new `AnimatedColor` node.
* [change] Update `ScrollView` to Class component for React 18 support.

## New features

* [add] Map `'button'` and `'paragraph'` role to equivalent semantic HTML elements.
* [add] Add support for `aria-*` props.
* [add] Add support for `role` prop.
* [add] Add support for `id` prop.
* [add] Add support for `tabIndex` prop (only values `0` and `-1`).
* [add] Support for PointerEvent props.
* [add] Add support for `pointerEvents` style.
* [add] Add support for `userSelect` style.
* [add] Add polyfilled support for W3C logical styles (e.g., `insetInline`.)
* [add] Add `<Image>` support for `tintColor` prop.
* [add] Add `<TextInput>` support for `enterKeyHint` prop.
* [add] Add `<TextInput>` support for `inputMode` prop.
* [add] Add `<TextInput>` support for `readOnly` prop.
* [add] Add `<TextInput>` support for `rows` prop.

## Deprecations

* [change] Deprecate `Touchable*` components. Use `Pressable`.
* [change] Deprecate `Button` component. Use `Pressable`.
* [change] Deprecate `StyleSheet.compose()`. Use style array syntax.
* [change] Deprecate `accessibility*` props. Use `aria-*` and `role` props.
* [change] Deprecate `nativeId` prop. Use `id` prop.
* [change] Deprecate `focusable` props. Use `tabIndex` prop.
* [change] Deprecate `pointerEvents` prop. Use `pointerEvents` style.
* [change] Deprecate `selectable` prop. Use `userSelect` style.
* [change] Deprecate `fontVariant` style array values. Use space-separated string values.
* [change] Deprecate `textAlignVertical` style. Use `verticalAlign` style.
* [change] Deprecate `transform` style array values. Use space-separated string functions.
* [change] Deprecate non-standard logical styles, e.g., 'marginStart'.
* [change] Deprecate `<Image>` style `resizeMode`. Use `resizeMode` prop.
* [change] Deprecate `<Image>` style `tintColor`. Use `tintColor` prop.
* [change] Deprecate `<TextInput>` prop `keyboardType`. Use `inputMode` prop.
* [change] Deprecate `<TextInput>` prop `returnKeyType`. Use `enterKeyHint` prop.
* [change] Deprecate `<TextInput>` prop `editable`. Use `readOnly` prop.
* [change] Deprecate `<TextInput>` prop `numberOfLines`. Use `rows` prop.

## Breaking changes

* [change] Remove previously deprecated Android/iOS platform-specific exports.
* [change] Remove previously deprecated `setNativeProps` API.
* [change] Remove previously deprecated `AppState.removeListener` API.
* [change] Text is position relative by default
* [change] Add task queue for InteractionManager

## Fixes

* [fix] `Animated` now works with compiled styles.
* [fix] Text inheritance of textAlign.

Todo:

- [x] https://github.com/necolas/react-native-web/issues/1506
- [x] https://github.com/necolas/react-native-web/issues/2379
- [x] Support `onPointer*` props.
- [x] Update docs
- [x] Remove integrations + remove list of co's using RNW
- [ ] `Image` DOM props
- [x] Notify maintainers of various projects
  - [x] react-native-screens https://github.com/software-mansion/react-native-screens/commits/main
  - [x] expo


packages/react-native-web/src/index.js
```diff
@@ -13,9 +13,7 @@ export { default as Animated } from './exports/Animated';
 export { default as Appearance } from './exports/Appearance';
 export { default as AppRegistry } from './exports/AppRegistry';
 export { default as AppState } from './exports/AppState';
-export { default as BackHandler } from './exports/BackHandler';
```
edkimmel, line 16: I briefly tested the canary build by changing package.json and yarn patching the babel plugin to match the changes in this PR.

I may be missing another step, but what is the expectation for code that previously expected ReactNativeWeb to match ReactNative's exports?  

I can work around first party code by creating platform-specific files, but libraries such as react-native-screens will fail until updated for the same reason.

necolas: RNW hasn't matched the RN exports for years. This is just removing the last of the platform-specific exports.

You can open an issue against their project and ask them to fix it. Looking at their code using BackHandler, it might not have crashed before but they have no special handling for web so functionally it probably wasn't cross-platform anyway. They also use it in only 2 places so it should be straightforward to fix.

necolas:
`react-native-web@0.0.0-c26cb9b08` is the latest canary with a lot of additional changes, including support for PointerEvent props and CSS logical properties. 0.19 will be the release that is compatible with the W3C props/styles changes that will be part of React Native 0.71.

I've been juggling a dozen things recently and will need more help than in previous releases to canary test changes and find bugs.

The canary currently includes many deprecation warnings. I'm also looking for feedback on that. I'm open to a couple of different approaches:

**Either**: 0.19 is released with deprecation warnings in console; 0.20 goes out with all deprecated props/styles removed and no warnings.

**Or**: 0.19 is released with no warnings in console; 0.20 is immediately released with all the warnings in console output; 0.21 removes all the deprecated APIs.

cc @EvanBacon for Expo perspective on how disruptive different approaches might be for their SDK users.

dereknelson:
great work, really exciting stuff! out of curiosity, why are the touchables* being deprecated?

rolandpoulter:
Trying to use`react-native-web@0.0.0-c26cb9b08` along with `react-native-paper` but it does not work seemingly because of the this commit: [16d3c23](https://github.com/necolas/react-native-web/pull/2377/commits/16d3c2396b819353096d98f77013d80140c4c5ad)

It fails when `react-native-paper` imports `BackHandler` which is exported from `react-native`. Are there any plans to restore this compatibility?

necolas:
You should create an issue against react-native-paper

rolandpoulter:
I don't think `react-native-paper` is using `react-native` incorrectly, and I thought the goal of this project was to provide the `react-native` API in the web. I'm happy to create an issue on their end but I think this issue could end up breaking other `react-native` dependencies running in the web using `react-native-web`.

I you suggesting I create an issue for the `react-native-paper` developers to stop using `BackHandler` in their Modal components?

necolas:
> Are you suggesting I create an issue for the react-native-paper developers to stop using BackHandler in their Modal components?

Yep. It never did anything on web and shouldn't be imported or used in web components anymore.

rolandpoulter:
`BackHandler` doesn't do anything in iOS either but it is still possible to import in iOS despite it not doing anything in that environment.

See:
https://github.com/callstack/react-native-paper/blob/cb9b028d3f708757ff08cd76b90906b4803bd2bf/src/components/Modal.tsx#L4

I understand the code doesn't do anything in the web, but it also doesn't do anything in iOS but it is implement as a no-op:

https://github.com/facebook/react-native/blob/8bd3edec88148d0ab1f225d2119435681fbbba33/Libraries/Utilities/BackHandler.ios.js

What other reasons is there for removing the BackHandler from the web? I think it sounds like a trade-off between compatibility and maintainability.

necolas:
Please create issues against other projects so they have time to make changes before this release goes out. This PR has been open for nearly 6 months, there have been several canary releases, and ample time for testing.

The release after this (0.20) will likely remove all the above-mentioned deprecated APIs entirely, so that RNWeb can get significantly smaller and faster.

rolandpoulter:
I don't think it will be significantly smaller and faster, infact it will be easier to wrap this library in the short term than to deal with this in every other project.

efoken:
I also need `BackHandler` in many projects. Conditionally importing it would not always work, for example in Next.js it needs to be present otherwise you get errors. Also I can imagine that you will have problems with bundlers. IMO it's the better solution to leave it there just as a function stub like @rolandpoulter suggested.

necolas:
There are already mechanisms in Metro to support per-platform modules. You can move BackHandler into a platform-specific file, i.e., `x.native.js` that you can then import into a cross-platform component if needed.

rolandpoulter:
@necolas can you explain how to do this in Metro? Also I am not using Metro and there are many other bundler tools out there that might not support that as a feature.

necolas:
https://reactnative.dev/docs/platform-specific-code#platform-specific-extensions

efoken:
@necolas you are suggesting to create a platform specific file to use BackHandler? But isn’t it the goal to reduce fragmentation?

rolandpoulter:
@necolas can you please explain how one could use platform specific extentions to install dependencies that use the `BackHandler`?

I don't see how that helps out here?

rolandpoulter:
@necolas I think you're going overboard with this https://github.com/necolas/discussions-and-proposals/blob/reduce-fragmentation/proposals/0000-reduce-fragmentation.md#other-maintenance

efoken:
@rolandpoulter I wrote a small hook for using `BackHandler` across platforms. Simply checking `BackHandler !== undefined` is enough for Web

```tsx
import * as React from "react"
import { BackHandler } from "react-native"

export function useBackHandler(fn: () => boolean) {
  React.useEffect(() => {
    if (BackHandler !== undefined) {
      BackHandler.addEventListener("hardwareBackPress", fn)
      return () => {
        BackHandler.removeEventListener("hardwareBackPress", fn)
      }
    }
    return () => {}
  }, [fn])
}
```

But anyway @necolas I would really prefer having a function stub exported from react-native-web, same as it is used for iOS. It should not be significantly increasing the bundle size of this library. And besides that, when not using something like this hook, people have to always check if `BackHandler` is defined which increases bundle size when used many times 😉

necolas:
Latest / late anticipated canary: 0.0.0-47671ca4

Planning to get 0.19 out this week barring any unexpected issues

burakgormek:
It would be nice align Linking api to RN before 0.19 publish :/

Deprecated Linking listeners still there... and we don't have `remove()` to remove subscription returned by `Linking.addEventListener`.

necolas:
@burakgormek no one reported that issue or submitted a PR. You should do it

PR #2389, opened 2022-08-30
[fix] export useLocaleContext as default export

krosenberg:
**Problem**
babel-plugin-react-native-web rewrites `import { useLocaleContext } from 'react-native'` to `import useLocaleContext from 'react-native-web/dist/exports/useLocaleContext';`

This rewrite seems to assume there is a default export in every `exports/` file. But that's not the case for useLocaleContext, whose [file](https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/useLocaleContext/index.js) looks like this:
```
export { useLocaleContext } from '../../modules/useLocale';
```
**Solution**
Set `useLocaleContext` as the default export in [exports/useLocaleContext/index.js](https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/useLocaleContext/index.js).


necolas:
Thank you. This should be fixed in 0.18.9

PR #2393, opened 2022-09-14
Added Embraer to companies list

ckrodrigues:
I work with a team working for the [Beacon](https://beacon.works) project which is owned by [Embraer](https://embraer.com).
We've been using `react-native-web` in production for about 3 years now.


necolas:
That's great!

I don't want to set the precedent that anyone using RNW should be listed in the README though, so I will probably remove everyone except Twitter and Meta.

steida:
@necolas You can do it as fp-ts, https://github.com/gcanti/fp-ts/issues/1124

necolas:
No thanks, I don't want to maintain a list of companies or imply endorsement of their products.

However, I created a pinned Discussion where people can share their examples - https://github.com/necolas/react-native-web/discussions/2417

PR #2401, opened 2022-10-01
Mention Solito under Integrations

nandorojo:
Solito is a library for bringing together React Native with Next.js' navigation. It has a number of starter cross-platform starter monorepos.

It could be useful to mention it on the docs as it gains popularity as a way of using React Native Web.


necolas:
I'd rather remove all integration links since these kinds of projects are frequently abandoned, and there's nothing special necessary for integration with any framework.

Issue #1254, opened 2019-02-21
infinite FlatList with inverted prop becomes crazy


Sharcoux:
**The problem**

When loading many items in a FlatList with the `inverted` prop, the list gets weird behaviours. The scroll moves on its own. It is also reverted whereas it shouldn't, and to top it all, some elements of the list gets inaccessible... 

**How to reproduce**

[Simplified test case](https://codesandbox.io/s/p5373vlm)

Steps to reproduce:
1. just go to the top of the list to load more messages.
2. repeat the process a few time, and things should start to become weird.

**Expected behavior**

If you remove the `inverted` prop, you get the expected behaviour. `inverted` should just make the items of the list to be displayed from bottom to top and `onEndReached` to be called when reaching the top.

**Environment (include versions). Did this work in previous versions?**

* React Native for Web (version): 0.10.0
* React (version): 16.5.1
* Browser: Chrome 72.0.3626.109

Sharcoux:
```
  verticallyInverted: {
    transform: [{scaleY: -1}],
  },
  horizontallyInverted: {
    transform: [{scaleX: -1}],
  },
```

Ouch! I don't think that this is a great way to handle the `inverted` property!
Replacing `flex-direction: row` by `flex-direction: row-reverse` would probably do a much better job, no?

necolas:
FlatList is a copy of what's in React Native

Sharcoux:
Then how come it behave so differently in the browser?
Using this mirror trick is responsible for the scroll to go in the opposite way, which is very unnatural. Moreover, the onEndReach events firing become completely broken...

Sharcoux:
Ok, I think I can fix it. I'll make a PR.

Sharcoux:
This is the pending PR. I will try to fix this problem directly with RN team.

necolas:
Linking the related issue @Sharcoux created for Yoga https://github.com/facebook/yoga/issues/866

sdcooke:
I had this issue - some very confusing behaviour with an inverted FlatList on web only. For anyone looking for a workaround `disableVirtualization={true}` seems to stop the weird behaviour for me, although it's obviously not an ideal solution.

caesar4321:
@sdcooke  `disableVirtualization={true} ` works for me too. But it is deprecated and will not be able to use on Expo either. What can we do next?

azesmway:
I solved this problem as follows
[...]

PR #2412, opened 2022-10-23
Fix Inverted VirtualizedList's baseline measurement for items' offset (1st of 4 problems that cause #1254)

LucioChavezFuentes:
### This proposal is deprecated, New proposal [here](https://github.com/necolas/react-native-web/pull/2501) ###

I found 4 problems that cause the [Inverted VirtualizedList to go wild (#1254)](https://github.com/necolas/react-native-web/issues/1254). This PR will explain and fix the **1st** of **4** problems found.

1. Inverted VirtualizedList has incorrect baseline measurement for its items’ offset  —> **Problem Explanation and Solution below**.
---
### **Update:** ### 

---
2. $lead_spacer expands scroll artificially when Inverted VirtualizedList is mounting new items —> [Problem Explanation and Solution in this PR](https://github.com/necolas/react-native-web/pull/2413).

#### Inverted FlatList with all problems fixed ####

### Inverted VirtualizedList has incorrect baseline measurement for its items’ offset. (**1st** of **4** problems that cause #1254) ###

FlatList uses a virtual area (VirtualizedList) where it only renders the elements we need to see. In a FlatList with items with different heights (like report chat list), VirtualizedList mounts and measures the ‘y’ offset position of each item and saves it on an object called `_frames`.

Once the offset of an item is saved, VirtualizedList (on each scroll) can safely unmount the measured item if it is out of the virtual area. And mount it back if the offset position is in the current virtual area. So an accurate and complete `_frames` is fundamental for a smooth and fast scroll on FlatList.

Inverted FlatList has inaccurate offset values in the object `_frames` because it measures items’ offsets (`y` position) as baseline measurement at the top (`y = 0`), but the scroll view’s offset baseline measurement is at the bottom.

**Why?**

Inside OnLayout’s react-native-web implementation, `y` offsets measures are done with `getBoundingClientRect()`. This API measures the DOM elements offsets as we see it, with `y = 0` at the top of our viewport (as a starting point of measure).

image8

A normal FlatList looks and measures like this:

Screen Shot 2022-10-13 at 17 26 21

The inverted FlatList uses `{transformY: -1}` that inverts not only our items but also our initial scroll position of the scroll view.   `y = 0` is now at the bottom of our scroll view:

Screen Shot 2022-10-12 at 17 53 57

But Inverted FaltList still measures item's offsets using `getBoundingClientRect()`(with `onLayout`) with `y = 0` at the top:

Screen Shot 2022-10-12 at 17 56 24

As you can see the first items have the greatest offsets and our last items have offsets closer to zero. But on each scroll, FlatList's scroll view expects the opposite. The problem gets worse with longer lists because our first item will have the greatest offset, but it will be based only within the virtualized area, not on our entire list.

### Solution ###

Use items’ `offsetTop` instead of onLayout’s `y`. `offsetTop` gets the correct offset for each item in normal or inverted lists because it ignores parent's transformations (like `transform: [{scaleY: -1}]`). 

[react-native-web/src/vendor/react-native/VirtualizedList/index.js](https://github.com/necolas/react-native-web/pull/2412/files#diff-a250e90659d45d2a93179da0bf6968e6f561a742ed0f74e097b84fb40a13ed50R1320-R1322)
[Screen Shot 2022-10-25 at 8 35 16](https://github.com/necolas/react-native-web/pull/2412/files#diff-a250e90659d45d2a93179da0bf6968e6f561a742ed0f74e097b84fb40a13ed50R1320-R1322)

### DIFF
--- a/packages/react-native-web/src/exports/UIManager/index.js
+++ b/packages/react-native-web/src/exports/UIManager/index.js
@@ -17,7 +17,8 @@ const getRect = (node) => {
   const { x, y, top, left } = getBoundingClientRect(node);
   const width = node.offsetWidth;
   const height = node.offsetHeight;
-  return { x, y, width, height, top, left };
+  const offsetTop = node.offsetTop;
+  return { x, y, width, height, top, left, offsetTop };
 };
 
 const measureLayout = (node, relativeToNativeNode, callback) => {
@@ -25,10 +26,10 @@ const measureLayout = (node, relativeToNativeNode, callback) => {
   if (node && relativeNode) {
     setTimeout(() => {
       const relativeRect = getBoundingClientRect(relativeNode);
-      const { height, left, top, width } = getRect(node);
+      const { height, left, top, width, offsetTop } = getRect(node);
       const x = left - relativeRect.left;
       const y = top - relativeRect.top;
-      callback(x, y, width, height, left, top);
+      callback(x, y, width, height, left, top, offsetTop);
     }, 0);
   }
 };
--- a/packages/react-native-web/src/exports/UIManager/index.js
+++ b/packages/react-native-web/src/exports/UIManager/index.js
@@ -17,8 +17,7 @@ const getRect = (node) => {
   const { x, y, top, left } = getBoundingClientRect(node);
   const width = node.offsetWidth;
   const height = node.offsetHeight;
-  const offsetTop = node.offsetTop;
-  return { x, y, width, height, top, left, offsetTop };
+  return { x, y, width, height, top, left };
 };
 
 const measureLayout = (node, relativeToNativeNode, callback) => {
@@ -26,10 +25,10 @@ const measureLayout = (node, relativeToNativeNode, callback) => {
   if (node && relativeNode) {
     setTimeout(() => {
       const relativeRect = getBoundingClientRect(relativeNode);
-      const { height, left, top, width, offsetTop } = getRect(node);
+      const { height, left, top, width } = getRect(node);
       const x = left - relativeRect.left;
       const y = top - relativeRect.top;
-      callback(x, y, width, height, left, top, offsetTop);
+      callback(x, y, width, height, left, top);
     }, 0);
   }
 };
--- a/packages/react-native-web/src/modules/useElementLayout/index.js
+++ b/packages/react-native-web/src/modules/useElementLayout/index.js
@@ -29,23 +29,20 @@ function getResizeObserver(): ?ResizeObserver {
           if (typeof onLayout === 'function') {
             // We still need to measure the view because browsers don't yet provide
             // border-box dimensions in the entry
-            UIManager.measure(
-              node,
-              (x, y, width, height, left, top, offsetTop) => {
-                const event: LayoutEvent = {
-                  // $FlowFixMe
-                  nativeEvent: {
-                    layout: { x, y, width, height, left, top, offsetTop }
-                  },
-                  timeStamp: Date.now()
-                };
-                Object.defineProperty(event.nativeEvent, 'target', {
-                  enumerable: true,
-                  get: () => entry.target
-                });
-                onLayout(event);
-              }
-            );
+            UIManager.measure(node, (x, y, width, height, left, top) => {
+              const event: LayoutEvent = {
+                // $FlowFixMe
+                nativeEvent: {
+                  layout: { x, y, width, height, left, top }
+                },
+                timeStamp: Date.now()
+              };
+              Object.defineProperty(event.nativeEvent, 'target', {
+                enumerable: true,
+                get: () => entry.target
+              });
+              onLayout(event);
+            });
           }
         });
       });
--- a/packages/react-native-web/src/modules/useElementLayout/index.js
+++ b/packages/react-native-web/src/modules/useElementLayout/index.js
@@ -29,20 +29,23 @@ function getResizeObserver(): ?ResizeObserver {
           if (typeof onLayout === 'function') {
             // We still need to measure the view because browsers don't yet provide
             // border-box dimensions in the entry
-            UIManager.measure(node, (x, y, width, height, left, top) => {
-              const event: LayoutEvent = {
-                // $FlowFixMe
-                nativeEvent: {
-                  layout: { x, y, width, height, left, top }
-                },
-                timeStamp: Date.now()
-              };
-              Object.defineProperty(event.nativeEvent, 'target', {
-                enumerable: true,
-                get: () => entry.target
-              });
-              onLayout(event);
-            });
+            UIManager.measure(
+              node,
+              (x, y, width, height, left, top, offsetTop) => {
+                const event: LayoutEvent = {
+                  // $FlowFixMe
+                  nativeEvent: {
+                    layout: { x, y, width, height, left, top, offsetTop }
+                  },
+                  timeStamp: Date.now()
+                };
+                Object.defineProperty(event.nativeEvent, 'target', {
+                  enumerable: true,
+                  get: () => entry.target
+                });
+                onLayout(event);
+              }
+            );
           }
         });
       });
--- a/packages/react-native-web/src/vendor/react-native/VirtualizedList/index.js
+++ b/packages/react-native-web/src/vendor/react-native/VirtualizedList/index.js
@@ -1319,7 +1319,7 @@ class VirtualizedList extends React.PureComponent<Props, State> {
     const next = {
       // OffsetTop gets the correct offset for each cell for inverted lists beacuse it ignores
       // parent's transformations (like 'transform: [{scaleY: -1}]').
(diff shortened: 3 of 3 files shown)


LucioChavezFuentes:
@necolas Let me know what you think, I'm happy to answer any questions.

Sharcoux:
Never seen a PR so well detailed. Nice job on that.

packages/react-native-web/src/exports/UIManager/index.js
```diff
@@ -17,18 +17,19 @@ const getRect = (node) => {
...
   const relativeNode = relativeToNativeNode || (node && node.parentNode);
   if (node && relativeNode) {
     setTimeout(() => {
       const relativeRect = getBoundingClientRect(relativeNode);
-      const { height, left, top, width } = getRect(node);
+      const { height, left, top, width, offsetTop } = getRect(node);
       const x = left - relativeRect.left;
       const y = top - relativeRect.top;
-      callback(x, y, width, height, left, top);
+      callback(x, y, width, height, left, top, offsetTop);
```
necolas, line 32: This would be changing the API, but I think we can avoid doing that.

packages/react-native-web/src/modules/useElementLayout/index.js
```diff
@@ -29,20 +29,23 @@ function getResizeObserver(): ?ResizeObserver {
...
-              });
-              onLayout(event);
-            });
+            UIManager.measure(
+              node,
+              (x, y, width, height, left, top, offsetTop) => {
+                const event: LayoutEvent = {
+                  // $FlowFixMe
+                  nativeEvent: {
+                    layout: { x, y, width, height, left, top, offsetTop }
```
necolas, line 38: Again, this is changing the expected values from the RN API and might be a problem if users start to think they can depend on `offsetTop` across platforms.

packages/react-native-web/src/vendor/react-native/VirtualizedList/index.js
```diff
@@ -1317,7 +1317,9 @@ class VirtualizedList extends React.PureComponent<Props, State> {
   _onCellLayout(e, cellKey, index) {
     const layout = e.nativeEvent.layout;
     const next = {
-      offset: this._selectOffset(layout),
+      // OffsetTop gets the correct offset for each cell for inverted lists beacuse it ignores
+      // parent's transformations (like 'transform: [{scaleY: -1}]').
+      offset: !this.props.horizontal ? layout.offsetTop : layout.x,
```
necolas, line 1322: Here we could get `offsetTop` directly from `e.target.offsetTop` rather than wiring it through all the other APIs

necolas (commented):
Hi, thanks for taking the time to explain your PRs in detail. Several of them appear to not be web-specific and so I wonder if you've looked into how much of this code should be upstream in react-native? The VirtualizedList code is almost entirely copy-pasted from react-native every few months, and web-specific code is likely to get lost each time that happens. If you can identify only the code that needs to exist in the react-native-web fork, that would be a big help. Thanks

LucioChavezFuentes:
Thank you @necolas, I didn't know about the connection between react-native-web and react-native. I thought components in react-native, like VirtualizedList, were purely written in native code.

I will review my code with these insights and make PRs accordingly.

packages/react-native-web/src/vendor/react-native/VirtualizedList/index.js
```diff
@@ -1317,7 +1317,9 @@ class VirtualizedList extends React.PureComponent<Props, State> {
   _onCellLayout(e, cellKey, index) {
     const layout = e.nativeEvent.layout;
     const next = {
-      offset: this._selectOffset(layout),
+      // OffsetTop gets the correct offset for each cell for inverted lists beacuse it ignores
+      // parent's transformations (like 'transform: [{scaleY: -1}]').
+      offset: !this.props.horizontal ? e.nativeEvent.target.offsetTop : layout.x,
```
LucioChavezFuentes, line 1322: @necolas I have a question regarding your concern about this solution. If I want this to be merged do I have to find a solution that does the same but uses only the RN layout tools and not platform-specific API?

Another question: do you have an estimated date when the new react-native's VirtualizedList code will be applied to react-native-web?

necolas: > If I want this to be merged do I have to find a solution that does the same but uses only the RN layout tools and not platform-specific API?

Ideally yes. But in this case it depends on whether the `y` value in RNW should be implemented differently to match RN. I still don't understand why `offsetTop` is used here but no where else. Is it because the layout values change on web with the scaleY transform, but don't in RN? Previously there have been suggestions that the scaleY approach breaks too many things on web anyway (like scroll wheel use).

This is why I want the canonical VirtualizedList implementation to have to support web, because it's too difficult to manage the necessary implementation across 2 projects when the upstream project doesn't think about web at all. Plus, VirtualizedList is slow on native and web, so RNW maintaining some fork of a slow thing (that most people shouldn't be using in its current state) is not a good use of time either.

> do you have an estimated date when the new react-native's VirtualizedList code will be applied to react-native-web?

No. But if this is blocking you, you could copy-paste it to RNW yourself and that updated will be part of the PR.

LucioChavezFuentes: > I still don't understand why offsetTop is used here but nowhere else. Is it because the layout values change on web with the scaleY transform, but don't in RN?

@necolas Yes, the `y` values of items change with [`transform: [{scaleY: -1}]`](https://github.com/necolas/react-native-web/blob/dee258ae02188befca4192ea502d4fbebc478341/packages/react-native-web/src/vendor/react-native/VirtualizedList/index.js#L2159) in Web. Meanwhile, in RN they don't change.

In RN, it doesn't matter if scaleY is normal or inverted, the first item will receive a `y` value equal to zero and the last item gets a `y` close to our list's length.

As I tried to explain in this PR's description, inside onLayout's Web implementation uses the DOM's [`getBoundingClientRect()`](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect). This API gets the positions of elements strictly relative to the viewport. Elements closer to the top of our screen will get `y` closer to zero and Elements further from the top will get a greater `y`.  When `transform: [{scaleY: -1}]` is applied, we position the first item of our list at the bottom, and Web's onLayout (with`getBoundingClientRect()`) returns the greatest `y` value for our first item. The last item, at the top of our VirtualizedList, will receive a `y` equal to zero. But as more elements are rendered, the new elements at the top of our inverted list will get a `y` value also equal to zero. That will make a mess in the calculation of virtualization.

`offsetTop` gets the `y` relative to the parent. It ignores the parent's transformations ( like `transform: [{scaleY: -1}]`). But in order to work the parent must be a positioned element i.e. `position: relative` or `position: absolute`.

This is an edge case where Web's onLayout is not working as its native counterpart. Basically, `offsetTop` here is patching a Web's onLayout edge case. An edge case caused by its functionality being affected by element transformations.

roryabraham: I agree this change in particular seems like a platform difference, but I think you need to account for horizontal inverted lists as well?

```suggestion
      offset: !horizontalOrDefault(this.props.horizontal) ? e.nativeEvent.target.offsetTop : e.nativeEvent.target.offsetLeft,
```

LucioChavezFuentes:
Hi @necolas , Just gut checking, I am planning to find a solution to make Web onLayout 'y' behave the same as RN's onLayout 'y' in inverted VirtualizedList. Do you think this would be a good path to follow?


In case you agree, do you know how I can test onLayout in multiple scenarios so I can't break something?

JWesorick:
Are these changes being submitted to react-native?

LucioChavezFuentes:
Hi @necolas, I submitted [PR 2](https://github.com/facebook/react-native/pull/35413), [PR 3](https://github.com/facebook/react-native/pull/35414), and [PR 4](https://github.com/facebook/react-native/pull/35415) for upstream react-native. I think this PR should be in react-native-web.

necolas:
@LucioChavezFuentes Thanks! Please close any PRs here that you no longer believe necessary.

necolas:
> Hi @necolas , Just gut checking, I am planning to find a solution to make Web onLayout 'y' behave the same as RN's onLayout 'y' in inverted VirtualizedList. Do you think this would be a good path to follow? In case you agree, do you know how I can test onLayout in multiple scenarios so I can't break something?

I think this is probably a good idea because we're also trying to make these lists npm packages that react-native-web can import, rather than having to maintain forked versions. And there may be other scenarios in which the difference in `y` values causes unexpected issues on web. In general it *should* be ok to make changes to `onLayout`, because it is used so much in APIs like the lists and other RN packages. If unexpected things break in a canary/RC release, we could always explore using patch-package in the future to make surgical changes to the list packages

genie19197:
Great details.

PR #2413, opened 2022-10-23
Fix artificial scroll limit expanding on scroll to un-measured area (2nd of 4 problems that cause #1254)

LucioChavezFuentes:
I found 4 problems that cause the [Inverted VirtualizedList to go wild (#1254)](https://github.com/necolas/react-native-web/issues/1254). This PR will explain and fix the **2nd** of **4** problems found.

---
### **Update:** ### 

---

2. $lead_spacer expands scroll artificially when Inverted VirtualizedList is mounting new items —> **Problem Explanation and Solution below**.

### $lead_spacer expands scroll limit artificially when Inverted VirtualizedList is mounting new items. (2nd of 4 problems that cause #1254) ###

In a FlatList with items with different heights, the scroll is limited to the last measured item. That’s because `VirtualizedList` (the inner `FlatList`'s component) must measure the offset of mounted items before it tries to mount new items. In this way, all items in the virtual area are measured and the `_frames` object is complete.

The scroll limit should be only expanded when the user tries to scroll beyond the scroll limit (unmeasured area) and then new items are mounted and their offsets are measured and saved in `_frames`. When recently mounted items are measured, the scroll limit is expanded to the new latest measured item, then the user will be ready to keep scrolling to unmeasured area and repeat the process to keep expanding the scroll limit until the last item is met.

Sometimes users may “overscroll” to unmeasured areas while previously mounted items haven't been measured yet, and `VirtualizedList` starts skipping items for measuring and measures others that shouldn’t be measured yet, leading to a fragmented map of offsets (e.g. we have offset values for …31, 32, 33, then jumps to 37, 38…).

**Why do users “overscroll”?**

Flaltlist constantly mounts and unmounts items. There is a white space mounted on our list called `$tail_spacer`. It helps us to fill with white space the top outside of our virtual area (because it’s an inverted list, for normal lists it fills the bottom), where items were previously mounted and measured but unmounted afterward, so the scroll limit sticks to our latest measured item offset.

Screen Shot 2022-10-12 at 18 02 58

`$tail_spacer` height depends on how much area we had measured, but it reduces to zero when the latest measured item is mounted and the user pretends to scroll to an unmeasured area looking for more items. When `$tail_spacer`'s height is zero, the scroll is limited and prevents the user from “overscroll”. This gives time to VirtualizedList to properly mount and measure new items to then use their offsets as our new scroll limit.

Sometimes `$tail_spacer` has a height greater than zero when it should be zero, allowing the user to “overscroll”.

**Why sometimes `$tail_spacer` has a height greater than zero when it should be zero?**

On some updates, `$tail_spacer` gets a negative value for its height, which is an invalid value. Invalid values will be ignored and $tail-spacer’s height will set the valid value of the previous state, instead of setting it to zero.

### Solution: ###

The solution is simple. Check `tailSpacerLenght`, if it is below zero then set it to zero:

[react-native-web/src/vendor/react-native/VirtualizedList/index.js](https://github.com/necolas/react-native-web/compare/master...Lucio-s-Forks:react-native-web:fix-tailSpacer-artificial-expanding#diff-a250e90659d45d2a93179da0bf6968e6f561a742ed0f74e097b84fb40a13ed50R1061-R1070)

[Screen Shot 2022-10-13 at 16 05 46](https://github.com/necolas/react-native-web/compare/master...Lucio-s-Forks:react-native-web:fix-tailSpacer-artificial-expanding#diff-a250e90659d45d2a93179da0bf6968e6f561a742ed0f74e097b84fb40a13ed50R1061-R1070)

### DIFF
--- a/packages/react-native-web/src/vendor/react-native/VirtualizedList/index.js
+++ b/packages/react-native-web/src/vendor/react-native/VirtualizedList/index.js
@@ -1058,10 +1058,16 @@ class VirtualizedList extends React.PureComponent<Props, State> {
           ? itemCount - 1
           : Math.min(itemCount - 1, this._highestMeasuredFrameIndex);
         const endFrame = this._getFrameMetricsApprox(end);
-        const tailSpacerLength =
+        // On some updates, tailSpacerLength is a negative value, which is an invalid value. 
+        // $tail-spacer’s height will ignore invalid values and  will 
+        // preserve the valid value of the previous state, instead of setting it to zero.
+        let tailSpacerLength =
           endFrame.offset +
           endFrame.length -
           (lastFrame.offset + lastFrame.length);
+        if(tailSpacerLength < 0) {
+            tailSpacerLength = 0
+        }
         cells.push(
           <View key="$tail_spacer" style={{[spacerKey]: tailSpacerLength}} />,
         );


roryabraham (changes requested):
I think this change belongs in the upstream React Native, not react-native-web.

The upstream implementation of VirtualizedList has [recently changed](https://github.com/facebook/react-native/commit/971599317b7bdf1152157206f9503a23ac8c4162#diff-7481ec8ed5532798b075df637e805a8e439807aa2ce671208c24068c286361e8), so unless I'm mistaken the new place to make this change would be [right here](https://github.com/facebook/react-native/blob/cf3747957ab210e31504109bb6b3e34e773a5b9a/Libraries/Lists/VirtualizedList.js#L910-L911), and the following change would do:

```diff
- const spacerSize =
-             lastMetrics.offset + lastMetrics.length - firstMetrics.offset;
+ const spacerSize = Math.max(
+     lastMetrics.offset + lastMetrics.length - firstMetrics.offset,
+     0,
+ );
```

necolas:
@LucioChavezFuentes thanks for working through this. And I forgot to mention this earlier but your bug reports with diagrams, details, and videos are really excellent. Thank you

LucioChavezFuentes:
> > I think this change belongs in the upstream React Native, not react-native-web.
> > The upstream implementation of VirtualizedList has [recently changed](https://github.com/facebook/react-native/commit/971599317b7bdf1152157206f9503a23ac8c4162#diff-7481ec8ed5532798b075df637e805a8e439807aa2ce671208c24068c286361e8), so unless I'm mistaken the new place to make this change would be [right here](https://github.com/facebook/react-native/blob/cf3747957ab210e31504109bb6b3e34e773a5b9a/Libraries/Lists/VirtualizedList.js#L910-L911), and the following change would do:
> > ```diff
> > - const spacerSize =
> > -             lastMetrics.offset + lastMetrics.length - firstMetrics.offset;
> > + const spacerSize = Math.max(
> > +     lastMetrics.offset + lastMetrics.length - firstMetrics.offset,
> > +     0,
> > + );
> > ```
> 
> @roryabraham I managed to run the VirtualizedList's latest changes on Web and this solution is no longer required but I don't know exactly why. I am still learning about the new VirtualizedList. The rest of the solutions still apply and fix the scroll issues in the new VirtualizedList on Web. I am on my way to propose these changes to react-native.



> > I think this change belongs in the upstream React Native, not react-native-web.
> > The upstream implementation of VirtualizedList has [recently changed](https://github.com/facebook/react-native/commit/971599317b7bdf1152157206f9503a23ac8c4162#diff-7481ec8ed5532798b075df637e805a8e439807aa2ce671208c24068c286361e8), so unless I'm mistaken the new place to make this change would be [right here](https://github.com/facebook/react-native/blob/cf3747957ab210e31504109bb6b3e34e773a5b9a/Libraries/Lists/VirtualizedList.js#L910-L911), and the following change would do:
> > ```diff
> > - const spacerSize =
> > -             lastMetrics.offset + lastMetrics.length - firstMetrics.offset;
> > + const spacerSize = Math.max(
> > +     lastMetrics.offset + lastMetrics.length - firstMetrics.offset,
> > +     0,
> > + );
> > ```
> 
> @roryabraham I managed to run the VirtualizedList's latest changes on Web and this solution is no longer required but I don't know exactly why. I am still learning about the new VirtualizedList. The rest of the solutions still apply and fix the scroll issues in the new VirtualizedList on Web. I am on my way to propose these changes to react-native.

@roryabraham I was wrong, this solution does provide more stability, especially on slow devices or when the user scrolls very fast. I will keep it for the react-native proposal.

PR #2415, opened 2022-10-24
Prevent VirtualizedList to save offsets equal to zero if item is not the first item of list (4th of 4 problems that cause #1254)

LucioChavezFuentes:
I found 4 problems that cause the [Inverted VirtualizedList to go wild (#1254)](https://github.com/necolas/react-native-web/issues/1254). This PR will explain and fix the **4th** of **4** problems found.

---
### **Update:** ### 

---

2. $lead_spacer expands scroll artificially when Inverted VirtualizedList is mounting new items —>  [Problem Explanation and Solution in this PR](https://github.com/necolas/react-native-web/pull/2413).

4. VirtualizedList gets offsets equal to zero for items that are not the list's first item (this happens also for normal lists) —> **Problem Explanation and Solution below**. 

### VirtualizedList gets offsets equal to zero for items that are not the list's first item (this happens also for normal lists, 4th of 4 problems that cause #1254) ###
As explained in the 3rd PR, a High Priority update cancels every Low Priority update. A canceled Low Priority update will make `onLayout` returns offset and height equal to zero for the items this update pretended to mount. These zero values will be saved in `_frames`, associated with the canceled item, even if that item was properly measured on previous updates.

`_frames` object with offsets equal to zero will cause a miss calculation to get items for the virtual area. Because it does not make sense that items in higher positions than the bottom of the list have an offset equal to zero.

### Solution: ###

Skip the action to save values on our `_frames` and trigger another update if two conditions happen: 1) `onLayout` returns zero for offset and 2) the item measured is not the first item of the FlatList.

[react-native-web/src/vendor/react-native/VirtualizedList/index.js](https://github.com/necolas/react-native-web/compare/master...Lucio-s-Forks:react-native-web:prevent-VirtualizedList-saves-offsets-equal-zero#diff-a250e90659d45d2a93179da0bf6968e6f561a742ed0f74e097b84fb40a13ed50R1326-R1333) → Inside: `_onCellLayout` function.
[image6](https://github.com/necolas/react-native-web/compare/master...Lucio-s-Forks:react-native-web:prevent-VirtualizedList-saves-offsets-equal-zero#diff-a250e90659d45d2a93179da0bf6968e6f561a742ed0f74e097b84fb40a13ed50R1326-R1333)

### DIFF
--- a/packages/react-native-web/src/vendor/react-native/VirtualizedList/index.js
+++ b/packages/react-native-web/src/vendor/react-native/VirtualizedList/index.js
@@ -1323,6 +1323,14 @@ class VirtualizedList extends React.PureComponent<Props, State> {
       inLayout: true,
     };
     const curr = this._frames[cellKey];
+
+    // Measures on discarded Low Priority updates will return zero for dimensions    
+    // no cell should have an offset of zero on frames, except for first cell
+    if (next.offset === 0 && index !== 0) {
+      this._scheduleCellsToRenderUpdate();
+      return;
+    }
+
     if (
       !curr ||
       next.offset !== curr.offset ||


packages/react-native-web/src/vendor/react-native/VirtualizedList/index.js
```diff
@@ -1323,6 +1323,14 @@ class VirtualizedList extends React.PureComponent<Props, State> {
       inLayout: true,
     };
     const curr = this._frames[cellKey];
+
+    // Measures on discarded Low Priority updates will return zero for dimensions    
+    // no cell should have an offset of zero on frames, except for first cell
```
roryabraham, line 1328:
```suggestion
    // no cell should have an offset of zero on frames, except for first cell,
    // so if that's the case, re-schedule another update without saving the incorrect layout
```

roryabraham (changes requested):
I believe this change also belongs in the upstream react-native, not react-native-web. You can apply the same change in the same place, [right here](https://github.com/facebook/react-native/blob/cf3747957ab210e31504109bb6b3e34e773a5b9a/Libraries/Lists/VirtualizedList.js#L1183-L1184) in `_onCellLayout`

PR #2425, opened 2022-11-12
Add task queue for InteractionManager

### LINKED ISSUE #2399: refactor: add the taskQueue in InteractionManager like react-native

In the react-native [documentation](https://reactnative.dev/docs/interactionmanager#methods) :
> InteractionManager.runAfterInteractions() is not working properly on web. It triggers immediately without waiting until the interaction is finished.

example:
In react-navigation,
we want a task to be deferred until the navigation animation ends. (after `clearInteractionHandle` is called)
Actually since the browser may be idle, it will be executed before `clearInteractionHandle` is called. (`requestIdleCallback`)

Can we submit a PR to migrate this feature? 🙂

ntdiary:
Migrate logic about task queues from `react-native`.

Please let me know if there are any irregularities or mistakes. 😄

Close #2399


ntdiary:
cc @necolas hello! Can you help with the review when you have free time? 🙂

package.json
```diff
@@ -58,7 +58,8 @@
     "lint-staged": "^13.0.3",
     "minimist": "^1.2.6",
     "npm-run-all": "^4.1.3",
-    "prettier": "^2.7.1"
+    "prettier": "^2.7.1",
+    "promise": "^7.3.1"
```
necolas, line 62: Why is native Promise not being used?

packages/react-native-web/src/exports/InteractionManager/__tests__/index-test.js
```diff
@@ -0,0 +1,323 @@
...
+}
+
+describe('InteractionManager', () => {
+  let InteractionManager;
+  let interactionStart;
+  let interactionComplete;
+
+  beforeEach(() => {
+    jest.resetModules();
+    InteractionManager = require('..');
```
necolas, line 19: Why is InteractionManager being required before each test?

packages/react-native-web/src/exports/InteractionManager/immediateShim.js
```diff
@@ -0,0 +1,75 @@
...
+          })
+    : setTimeout;
+
+/**
+ * Shim the setImmediate API on top of queueMicrotask.
+ * @param {function} callback Callback to be invoked before the end of the
+ * current JavaScript execution loop.
+ * @param args
+ */
+function setImmediate(callback: Function, ...args: any): number {
```
necolas, line 37: Do we need to use setImmediate? It's not a web standard and I'm hesitant to introduce a shim if you think there could be other ways to implement the InteractionManager behavior

package.json
ntdiary: This package is only used in `TaskQueue-test.js`, which has almost the same test cases as in [`react-native`](https://github.com/facebook/react-native/blob/main/Libraries/Interaction/__tests__/TaskQueue-test.js).
It ensures that [the code](https://github.com/ntdiary/react-native-web/blob/3db9b387a329ea755ebd069e54f028b20b237189/packages/react-native-web/src/exports/InteractionManager/TaskQueue.js#L103-L104) in `task.gen().then()` is called before the corresponding `expect`.
We can change the test cases and then remove this package, it will not affect normal use cases.
```diff
-import Promise from 'promise';

 function expectToBeCalledOnce(fn) {
   expect(fn.mock.calls.length).toBe(1);
@@ -68,8 +67,10 @@ describe('TaskQueue', () => {

     clearTaskQueue(taskQueue);

-    expectToBeCalledOnce(onMoreTasks);
-    expectToBeCalledOnce(task2);
+    return Promise.resolve(() => {
+      expectToBeCalledOnce(onMoreTasks);
+      expectToBeCalledOnce(task2);
+    })
   });

   it('should handle nested simple tasks', () => {
@@ -115,10 +116,12 @@ describe('TaskQueue', () => {

     clearTaskQueue(taskQueue);

-    expectToBeCalledOnce(task1);
-    expectToBeCalledOnce(task2);
-    expectToBeCalledOnce(task3);
-    expectToBeCalledOnce(task4);
+    return Promise.resolve(() => {
+      expectToBeCalledOnce(task1);
+      expectToBeCalledOnce(task2);
+      expectToBeCalledOnce(task3);
+      expectToBeCalledOnce(task4);
+    });
   });
```

packages/react-native-web/src/exports/InteractionManager/__tests__/index-test.js
ntdiary: `InteractionManager` has local variables and queue. This can easily avoid conflicts between these test cases. https://jestjs.io/docs/28.x/jest-object#jestresetmodules.

packages/react-native-web/src/exports/InteractionManager/immediateShim.js
ntdiary: Oh, sorry, the name here may be misleading. It is different from [`setImmediate` in MDN](https://developer.mozilla.org/en-US/docs/Web/API/Window/setImmediate#notes).
The former actually tries to use `queueMicrotask` to create a micro task ([polyfill](https://github.com/facebook/react/blob/2655c9354d8e1c54ba888444220f63e836925caa/packages/react-noop-renderer/src/createReactNoop.js#L421-L433) is `Promise.resolve`/`setTimeout`),  while the latter is a macro task(similar to `setTimeout`, except that it gets executed immediately, [polyfill](https://github.com/YuzuJS/setImmediate) is `postMessage`/`MessageChannel`/`setTimeout`).
I thought it would be better to execute these tasks that have been deferred by `InteractionManager` before the next macro task.
Here is my current understanding, please correct me if there is a mistake 😂:
image

* `queueMicrotask/Promise` compatibility is about 97.68%
* `requestAnimationFrame` compatibility is about 98.48%
* native `requestIdleCallback` compatibility is about 	77.96%

Which way do you prefer? 😄
### Some references:
* > As can be seen from the following examples, the best way of thinking about [queueMicrotask()](https://html.spec.whatwg.org/multipage/timers-and-user-prompts.html#dom-queuemicrotask) is as a mechanism for rearranging synchronous code, effectively placing the queued code immediately after the currently executing synchronous JavaScript has run to completion.
* https://stackoverflow.com/questions/43050448/when-will-requestanimationframe-be-executed

ntdiary: And actually, the shim file can be deleted to avoid misleading if needed.
```diff
diff --git a/packages/react-native-web/src/exports/InteractionManager/index.js b/packages/react-native-web/src/exports/InteractionManager/index.js
index e3b535ab..5ccc7caf 100644
--- a/packages/react-native-web/src/exports/InteractionManager/index.js
+++ b/packages/react-native-web/src/exports/InteractionManager/index.js
@@ -94,6 +94,20 @@ let _nextUpdateHandle: TimeoutID | number = 0;
 let _inc = 0;
 let _deadline = -1;

+const scheduleMicrotask =
+  typeof window !== 'undefined' && typeof window.queueMicrotask === 'function'
+    ? window.queueMicrotask
+    : typeof Promise !== 'undefined'
+      ? (callback) =>
+        Promise.resolve(null)
+          .then(callback)
+          .catch((error) => {
+            setTimeout(() => {
+              throw error;
+            });
+          })
+      : setTimeout;
+
 /**
  * Schedule an asynchronous update to the interaction state.
  */
@@ -102,7 +116,8 @@ function _scheduleUpdate() {
     if (_deadline > 0) {
       _nextUpdateHandle = setTimeout(_processUpdate);
     } else {
-      _nextUpdateHandle = setImmediate(_processUpdate);
+      scheduleMicrotask(_processUpdate);
+      _nextUpdateHandle = 1;
     }
   }
 }
```

necolas:
Thanks. Merged in 47d77ac256cf1324e684362465a2b0001d2db574

PR #2438, opened 2022-11-23
Fix: iOS not firing `resize` event

### LINKED ISSUE #2430: Dimensions change event is not fired on iOS (Safari)

### Describe the issue

Dimensions `change` event is not fired when keyboard is triggered on Safari.

It's similar to this issue #1369 which we have fixed in the past.

### Expected behavior

The event should be fired when keyboard state changes, like it does on Chrome.

CSB Preview: https://scdqh6.sse.codesandbox.io/dimensions

### Steps to reproduce

1. Go to above CSB Preview URL
2. Interact with the text input.
3. See the `window.height` does not change

### Test case

https://codesandbox.io/s/stupefied-brook-scdqh6?file=/pages/dimensions/index.js

### Additional comments

_No response_

rawalyogendra:
Fixes #2430 

This is related to iOS devices not firing `resize` event when keyboard opens/closes.

The workaround is listening to `window.visualViewport` `resize` event and updating the height and width by the values returned by `window.visualViewport`

### DIFF
--- a/packages/react-native-web/src/exports/Dimensions/__tests__/index-test.js
+++ b/packages/react-native-web/src/exports/Dimensions/__tests__/index-test.js
@@ -55,3 +55,60 @@ describe('apis/Dimensions', () => {
     expect(handler).toHaveBeenCalledTimes(0);
   });
 });
+
+describe('apis/Dimensions(VisualViewport)', () => {
+  window.visualViewport = {
+    width: window.innerWidth,
+    height: window.innerHeight,
+    scale: 1,
+    dispatchEvent: (event) => window.dispatchEvent(event)
+  };
+
+  test('get(VisualViewport)', () => {
+    const handler = jest.fn();
+    Dimensions.addEventListener('change', handler);
+    expect(Dimensions.get('screen')).toMatchInlineSnapshot(`
+      {
+        "fontScale": 1,
+        "height": 0,
+        "scale": 1,
+        "width": 0,
+      }
+    `);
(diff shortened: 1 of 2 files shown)


rawalyogendra:
@necolas I have added the previous implementation as fallback. All tests are passing.

flodnv:
@necolas could you please review this PR? 🙏

mananjadhav:
@necolas Anyway we can help to get this merge sooner?  🙇

flodnv:
@necolas any chance you can have one more look at this soon? 🙏 We're happy to sponsor you, either through this GH repo or through an upwork job along with other existing upwork jobs from https://github.com/Expensify/App/issues/11463

necolas:
Please update the unit tests to cover the new branching logic. Thanks

rawalyogendra:
@necolas wanted to clear something with you.

For the unit test:

- Should I update the existing tests to cover our new `VisualViewport` implementation? or
- Should I add new tests to cover the implementation on top of the old tests? or
- Are you suggesting something else?

necolas:
Yes both branches should be covered by tests. You've added a new branch but have not added tests

rawalyogendra:
@necolas I am not sure about the best way to test these new changes.

But I have added the tests nonetheless. Let me know if we could do this in a better way.

flodnv:
@necolas any further thoughts on this change? 🙏

necolas:
Thanks. I deleted the extra tests because they were just checking the fallback path

flodnv:
Thank you @necolas !

PR #2440, opened 2022-11-24
Only emit change events when Dimensions really change

sunnylqm:


necolas:
The current implementation is based on the browser's heuristic and I don't see a reason included to change that. Not to mention stringifying on every event is needless overhead. You can do this in your user code if you must

sunnylqm:
Stringifying on every event is surely some overhead, but still much less than dispatching a same event (especially on firefox) -> trigger `useWindowDimensions` -> trigger all components with that hook to rerender.

necolas:
Sorry but a PR with no described purpose isn't going to get reviewed let alone merged

sunnylqm:
i think the title is clear enough for the purpose. if you feel not enough just ask me to provide more instead of simply closing.

PR #2442, opened 2022-11-25
Image: support ImageSource with headers

kidroca:
## Details 

Extend `ImageLoader` functionality to be able to work with image sources containing headers

We preserve the existing strategy that works with `image.src` for cases where the `ImageSource` 
is just an `uri` with no headers

When headers are involved we use the `fetch` API to load the image

Why are there 2 ways to load images? 
Because `fetch` or `xhr` does not work for
- loading progressive JPEG images
- loading crossorigin blobs

Cross origin image requests can still be made with headers (`fetch`) if the server is configured correctly

## Fixed Issues

- #1019

## Test Strategy

1. Verify existing Image functionality has no regressions
    - build web and examples: ` npm run dev -w react-native-web` and `npm run dev -w react-native-web-examples`
    - open the examples page and go to `Image`: http://localhost:3000/image
    - see images are loading
    - take a screenshot and do the same from the `master` branch. You can switch back and forth and verify the image are loading the same way

2. Verify Images with headers can be loaded
    - build web and examples: ` npm run dev -w react-native-web` and `npm run dev -w react-native-web-examples`
    - open the examples page and go to `Image`: http://localhost:3000/image
    - modify `sourceWithHeaders` here `packages/react-native-web-examples/pages/image/index.js` and try to load images from a server that expects a GET request with a header
    - verify the image is loading on the examples page (near the bottom, labeled: "With Headers")


packages/react-native-web/src/modules/ImageLoader/index.js
```diff
@@ -111,37 +119,78 @@ const ImageLoader = {
...
       // avoid blocking the main thread
-      const onDecode = () => onLoad({ nativeEvent: e });
-      if (typeof image.decode === 'function') {
-        // Safari currently throws exceptions when decoding svgs.
-        // We want to catch that error and allow the load handler
-        // to be forwarded to the onLoad handler in this case
-        image.decode().then(onDecode, onDecode);
-      } else {
-        setTimeout(onDecode, 0);
-      }
```
kidroca, line 135: Removed the `if (typeof image.decode === 'function')` check and the `setTimeout` fallback, because the minimal supported browser versions are all capable of `image.decode`

packages/react-native-web/src/modules/ImageLoader/index.js
```diff
@@ -111,37 +119,78 @@ const ImageLoader = {
...
+        URL.revokeObjectURL(image.src);
+      };
+
+      fetch(request)
+        .then((response) => response.blob())
+        .then((blob) => {
+          image.src = URL.createObjectURL(blob);
+        })
+        .catch((error) => {
+          if (error.name !== 'AbortError') onError(error);
```
kidroca, line 180: Skip raising `onError` for aborted requests

packages/react-native-web/src/modules/ImageLoader/index.js
```diff
@@ -74,13 +78,17 @@ let id = 0;
...
+    const request = requests[requestId];
+    if (request) {
+      const { image, cleanup, source } = request;
+      if (cleanup) cleanup();
+
       image.onerror = null;
       image.onload = null;
-      image = null;
-      delete requests[`${requestId}`];
+      image.src = '';
```
kidroca, line 89: Setting image.src to empty string aborts any ongoing image loading

packages/react-native-web/src/exports/Image/__tests__/__snapshots__/index-test.js.snap
```diff
@@ -329,14 +329,14 @@ exports[`components/Image prop "style" removes other unsupported View styles 1`]
 >
   <div
     class="css-view-175oi2r r-backgroundColor-1niwhzg r-backgroundPosition-vvn4in r-backgroundRepeat-u6sd8q r-bottom-1p0dtai r-height-1pi2tsx r-left-1d2f490 r-position-u8s1d r-right-zchlnj r-top-ipm5af r-width-13qz1uu r-zIndex-1wyyakw r-backgroundSize-4gszlv"
-    style="filter: url(#tint-57);"
+    style="filter: url(#tint-86);"
```
kidroca, line 332: These tint id changes seems to be, because of the added/changed tests - each new mounted Image component increases an index used here

packages/react-native-web/src/exports/Image/__tests__/index-test.js
```diff
@@ -21,6 +21,11 @@ describe('components/Image', () => {
   beforeEach(() => {
     ImageUriCache._entries = {};
     window.Image = jest.fn(() => ({}));
+    ImageLoader.load = jest
+      .fn()
+      .mockImplementation((source, onLoad, onError) => {
+        onLoad({ source });
+      });
```
kidroca, line 28: Some tests were modifying the behavior of the `ImageLoader.load` method, since there's no cleanup 
the behavior would remain as the last `jest.fn().mockImplementation` assigned to the method

It's more consistent to have the behavior predefined uniformly before each test, while we still let some tests have a custom implementation

packages/react-native-web/src/exports/Image/index.js
```diff
@@ -94,61 +95,105 @@ function getFlatStyle(style, blurRadius, filterId) {
...
-    const { height, width } = getAssetByID(source);
-    return { height, width };
-  } else if (
-    source != null &&
-    !Array.isArray(source) &&
-    typeof source === 'object'
-  ) {
-    const { height, width } = source;
-    return { height, width };
-  }
```
kidroca, line 108: Updated code to always call this function with a source object so width/height can be destructed

packages/react-native-web/src/exports/Image/index.js
```diff
@@ -94,61 +95,105 @@ function getFlatStyle(style, blurRadius, filterId) {
...
+function resolveAssetDimensions(source: ImageSource) {
+  const { height, width } = source;
+  return { height, width };
 }
 
-function resolveAssetUri(source): ?string {
-  let uri = null;
```
kidroca, line 112: Updated this function to always return an object, the existing logic was just moved to separate helper functions
- `resolveNumericSoure`
- `resolveStringSource`
- `resolveObjectSource`
- `resolveSvgDataUriSource`

packages/react-native-web/src/exports/Image/index.js
```diff
@@ -94,61 +95,105 @@ function getFlatStyle(style, blurRadius, filterId) {
...
       );
     }
-    const scaleSuffix = scale !== 1 ? `@${scale}x` : '';
-    uri = asset
-      ? `${asset.httpServerLocation}/${asset.name}${scaleSuffix}.${asset.type}`
-      : '';
-  } else if (typeof source === 'string') {
-    uri = source;
+
+    return resolveSource(source[0]);
```
kidroca, line 118: Added this Array handling case
Since arrays aren't covered we try to recover by using the first object in the array

packages/react-native-web/src/exports/Image/index.js
```diff
@@ -191,33 +236,24 @@ const Image: React.AbstractComponent<
...
-      if (isLoaded) {
-        return LOADED;
-      }
-    }
-    return IDLE;
-  });
+  const imageLoadingProps = { onLoad, onLoadStart, onLoadEnd, onError };
+
+  const fallbackSource = useSource(imageLoadingProps, defaultSource);
+  const mainSource = useSource(imageLoadingProps, source);
```
kidroca, line 242: Extracted the `useEffect` hook here in order to load both `defaultSource` and `source` using the same logic (and supporting source headers)

Any dependencies were moved to the hook

packages/react-native-web/src/modules/ImageLoader/index.js
```diff
@@ -111,37 +119,78 @@ const ImageLoader = {
...
-      }
+      const onDecode = () => {
+        ImageUriCache.add(source.uri);
+        onLoad({
+          source: {
+            uri: image.src,
+            width: image.naturalWidth,
+            height: image.naturalHeight
+          }
+        });
```
kidroca, line 147: This is a BREAKING CHANGE 
The `nativeEvent` from the `onLoad` callback is now `{ nativeEvent: { source: { width, height, uri } } }`
I can preserve the old structure where `nativeEvent` is just the `Event` object from `image.onload`, but I'll have to add these width/height/uri properties under a different key and they won't match the same interface as used in react-native: https://reactnative.dev/docs/image#onload

Do you want to change the code to avoid a breaking change or do you want it to match `react-native` interface?

packages/react-native-web/src/modules/ImageLoader/index.js
```diff
@@ -111,37 +119,78 @@ const ImageLoader = {
...
+      };
+
+      // Safari currently throws exceptions when decoding svgs.
+      // We want to catch that error and allow the load handler
+      // to be forwarded to the onLoad handler in this case
+      image.decode().then(onDecode, onDecode);
```
kidroca, line 153: I think we're probably OK to also remove this comment and handling about Safari since it's from 2018 and iOS 11
Looks like the bug was addressed for newer versions of Safari: https://bugs.webkit.org/show_bug.cgi?id=201243

kidroca (commented):
Leaving some notes on the code changes

There's a breaking change that might be introduced or reverted - it's regarding the `onLoad` prop and the parameter structure used - I've opened a thread on the relevant source code chunk

necolas:
It would have been a good idea to ask about this before writing the PR, as there are numerous other plans around Image. 

There's this PR for crossOrigin support https://github.com/necolas/react-native-web/pull/2207

This milestone https://github.com/necolas/react-native-web/milestone/18

And a parallel effort to support W3C props on Image https://github.com/facebook/react-native/issues/34424#:~:text=native%2Did%22%0A%3E-,%3CImage%3E%20props,-These%20props%20should

kidroca:
Hi @necolas 
I thought this PR would be a nice place to get the conversation going

This PR is following the original idea to support headers via XHR, though I've opted to use `fetch` since the minimal browser requirements allow it 

Overall I think the changes here are compatible with #2207

A big chunk of the changes comes from added tests
And the change for handling `source` objects in hook dependencies, which is also a problem for #2207 - https://github.com/necolas/react-native-web/pull/2207/files#diff-7cb74a3a32d73857be80350ecd1ea131d256bd5af11d2000e4fc2d03c2230584R222-R223

The rest of the changes are small cleanups and moving existing logic - I can revert some of these namely `resolveAssetUri` to avoid conflicts with the other PR

If there are any problems, I'm happy to discuss and modify the PR

necolas:
Please resolve the conflicts with master

kidroca:
There are no conflicts with `master` right now 

There would be conflicts for the PR that happens to be merged last due to the changes to `resolveAssetUri` I made

Since the other PR just moves the function without changing the logic, it might be easier to do the same after this PR is merged
Alternatively I can apply the change after the other PR is merged

necolas:
Here's what GitHub shows me on this PR

image

necolas:
If you could rebase/squash your 20 commits into just a few, that would help too. Thanks

packages/react-native-web/src/exports/Image/__tests__/index-test.js
```diff
@@ -346,6 +425,67 @@ describe('components/Image', () => {
...
+    });
+
+    test('removes image if source or source.uri is removed and there is no default source', () => {
+      const source = { uri: 'https://google.com/favicon.ico' };
+
+      const { container, rerender } = render(<Image source={source} />);
+
+      rerender(<Image source={{ uri: '' }} />);
+      expect(container.querySelector('img')).toBe(null);
+    });
```
kidroca, line 488: Added 2 more tests

kidroca (commented):
Squashed existing commits to allow rebasing

NilsBaumgartner1994:
Push, we need Authorization headers

packages/react-native-web/src/modules/ImageLoader/index.js
```diff
@@ -122,9 +122,17 @@ const ImageLoader = {
...
-    image.onload = (e) => {
+    image.onload = (nativeEvent) => {
       // avoid blocking the main thread
-      const onDecode = () => onLoad({ nativeEvent: e });
+      const onDecode = () => {
+        nativeEvent.source = {
+          uri: image.src,
+          width: image.naturalWidth,
+          height: image.naturalHeight
+        };
```
kidroca, line 132: Attach `source` to `nativeEvent` so `onLoad` matches the RN Image interface https://reactnative.dev/docs/image#onload

packages/react-native-web/src/exports/Image/index.js
```diff
@@ -337,24 +345,94 @@ const Image: React.AbstractComponent<
...
+      .then((response) => response.blob())
+      .then((blob) => {
+        uri = URL.createObjectURL(blob);
+        setBlobUri(uri);
+      })
+      .catch((error) => {
+        if (error.name !== 'AbortError' && onError) {
+          onError({ nativeEvent: error.message });
+        }
+      });
```
kidroca, line 388: Perhaps it would be best to move this logic to `ImageLoader.loadUsingHeaders` as it would be easier to mock inside tests 

Otherwise we have to mock
- Request
- fetch
- URL.createObjectURL and revokeObjectURL

kidroca (commented):
PR Updated using an alternative strategy to load images with headers resulting in less changes overall

When a `source` with headers is used, we use a (kind of) a HOC component that fetches the `source` and then passes a local uri created with `URL.createObjectURL` to the original Image component

Beamanator:
@necolas any chance you or a fellow maintainer can check out these changes? We're fine to wait since, as you mentioned before, you have many updates planned - we're just hoping to get a possible timeline when this could get reviewed.

necolas:
It looks like the author is still working on the patch. It needs unit tests and the suggestion to make changes to make testing easier makes sense to me.

necolas:
What's next for this PR? The PR in the fork had no new unit tests and this one hasn't been updated for 2 weeks

kidroca:
I haven't synced my latest changes here, and I still need to write some unit tests 
I'll post an update when I'm done

packages/react-native-web/src/exports/Image/__tests__/index-test.js
```diff
@@ -106,10 +116,6 @@ describe('components/Image', () => {
 
   describe('prop "onLoad"', () => {
     test('is called after image is loaded from network', () => {
-      jest.useFakeTimers();
-      ImageLoader.load = jest.fn().mockImplementation((_, onLoad, onError) => {
-        onLoad();
-      });
```
kidroca, line 112: Fake timers no longer needed to pass this test

packages/react-native-web/src/exports/Image/index.js
```diff
@@ -146,6 +147,23 @@ function resolveAssetUri(source): ?string {
...
+function raiseOnErrorEvent(uri, { onError, onLoadEnd }) {
+  if (onError) {
+    onError({
+      nativeEvent: {
+        error: `Failed to load resource ${uri} (404)`
+      }
+    });
+  }
+  if (onLoadEnd) onLoadEnd();
+}
```
kidroca, line 164: This was extracted for reuse out of the original Image component hook
(We reuse this in case loading with headers fails)

packages/react-native-web/src/exports/Image/index.js
```diff
@@ -332,14 +343,76 @@ const Image: React.AbstractComponent<
...
-Image.displayName = 'Image';
+BaseImage.displayName = 'Image';
+
+/**
+ * This component handles specifically loading an image source with headers
+ * default source is never loaded using headers
+ */
+const ImageWithHeaders: ImageComponent = React.forwardRef((props, ref) => {
+  // $FlowIgnore: This component would only be rendered when `source` matches `ImageSource`
+  const nextSource: ImageSource = props.source;
```
kidroca, line 359: Perhaps there's a way to declare `ImageWithHeaders` as an `Image` component where the `source` prop is exactly `type ImageSource`, but I don't know exactly how 

I guess I need to declare `ImageWithHeaderProp` type that spreads the default Image prop type but changes the `source`. Should I do that?

packages/react-native-web/src/exports/Image/types.js
```diff
@@ -102,8 +102,8 @@ export type ImageStyle = {
   tintColor?: ColorValue
 };
 
-export type ImageProps = {
-  ...ViewProps,
+export type ImageProps = {|
+  ...$Exact<ViewProps>,
```
kidroca, line 106: I had `Flow` errors after my changes
Following the error messages and suggested fixes led to these changes

packages/react-native-web/src/exports/ImageBackground/index.js
```diff
@@ -16,12 +16,12 @@ import Image from '../Image';
...
 import View from '../View';
 
-type ImageBackgroundProps = {
+type ImageBackgroundProps = {|
   ...ImageProps,
   imageRef?: any,
   imageStyle?: $PropertyType<ImageProps, 'style'>,
   style?: $PropertyType<ViewProps, 'style'>
-};
+|};
```
kidroca, line 24: I had `Flow` errors after my changes
Following the error messages and suggested fixes led to these changes

kidroca (commented):
Ready for Review ✅

packages/react-native-web/src/exports/Image/__tests__/index-test.js
kidroca: Actually the fake timers are breaking the (new) tests that use `Promises` 

For some reason unless we remove all calls to `jest.useFakeTimers()` - `return waitFor(() => ...)` does not work

necolas:
Hi, please could you rebase this. Now that 0.19 is done, the 0.20 release will be focused on Image changes.

kidroca:
✅ Rebased

raqso (approved):

packages/react-native-web/src/exports/Image/index.js
```diff
@@ -353,14 +364,76 @@ const Image: React.AbstractComponent<
...
+    // Until the current component resolves the request (using headers)
+    // we skip forwarding the source so the base component doesn't attempt
+    // to load the original source
+    source: blobUri ? { ...nextSource, uri: blobUri } : undefined
+  };
+
+  return <BaseImage ref={ref} {...propsToPass} />;
+});
 
 // $FlowIgnore: This is the correct type, but casting makes it unhappy since the variables aren't defined yet
```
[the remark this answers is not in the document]
kidroca: I think I've seen the original error the comment references at some point 
Could it be the error was fixed by a flow version update, but now we have this other error?

PR #2461, opened 2022-12-30
[github] update actions used in workflows to fix warnings

Simek:
# Why

Fixes warnings from:
* https://github.com/necolas/react-native-web/actions/runs/3788572079
* https://github.com/necolas/react-native-web/actions/runs/3781121787

# How

Update `checkout` and `node-setup` action to the latest version, which makes those action run on newest version of Node.

Unfortunately, the same problem affects:
* forked `compressed-size-action` - fork sync should be enough, since base action already switched Node version:
  * https://github.com/preactjs/compressed-size-action/blob/master/action.yml#L43
* `hramos/label-actions` - this action now uses the different name, there is a `v2` but it still runs on Node 12, it might be worth to prepare PR and ping Hector, but alternatively you can switch to https://github.com/dessant/label-actions to address the issue

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature.yml
+++ b/.github/ISSUE_TEMPLATE/feature.yml
@@ -16,6 +16,6 @@ body:
   - type: textarea
     attributes:
       label: Describe the feature request
-      description: Please provide a concise description of the request, potential solutions, and addtional context.
+      description: Please provide a concise description of the request, potential solutions, and additional context.
     validations:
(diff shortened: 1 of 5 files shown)


necolas:
> alternatively you can switch to https://github.com/dessant/label-actions to address the issue

thanks, that sounds like a good idea to me

necolas:
Thanks for noticing this and providing a PR!

PR #2493, opened 2023-03-03
Fix Image Caching

### LINKED ISSUE #2492: Adding already rendered image has a small flicker

### Describe the issue

When we're (repeatedly) adding the same image on screen there's a small flicker happening the first time the component renders 

For example the avatars of a chat app 
- each time a message is added/received the image of the new message flickers 
- even though the avatar is the same as the one rendered on previous messages above

### Expected behavior

Adding the same image twice, should render instantly the 2nd time it's used

### Steps to reproduce

Browser: Chrome
ENV: codesandbox
react-native-web: 0.18.12 

Steps:
1. Go to the test case address
1. Press the "Add More" button at the top a few time
2. Notice each time the images flicker
3. (Optional) uncomment the `defaultSource` (L34) and repeat the steps - no flicker occurs

### Test case

https://codesandbox.io/s/rnw-image-flicker-bug-sample-obbhov

### Additional comments

There's no such problem if we use both `defaultSource` and `source` with the same source 

```jsx
<Image soruce={{ uri: '/my-image.jpg' }} defaultSource={{ uri: '/my-image.jpg' }} />
```

This is because of internal logic immediately rendering the `defaultSource`, while `source` is not rendered and always starts from IDLE state. 
I've traced this back to refactoring the `Image` from class based to functional component, where we no longer add images to `ImageUriCache` (probably it was overlooked in the refactor?)

---

The flicker can be easily observer if we put a breakpoint in Image render's code 
1. First render - no dom element added, loading state IDLE ❌ 
2. 2nd render - loading state IDLE, dom element added, but empty ❌ 
3. 3rd render - loading state LOADED, dom element updated with image content ✅ 

Since the image was already rendered and is available on screen it should be immediately rendered at step 1

kidroca:
## Description

This PR fixes some subtle inconsistencies around image caching and rendering the same image multiple times

Fixes #2492


kidroca:
BTW This PR also fixes `Image.queryCache` - I don't think we have an issue for this, but right now `Image.queryCache` does not work as intended (nothing already loaded is considered cached)
@necolas do you want me to open an issue for tracking purposes?

necolas:
Hi, please could you rebase this. Now that 0.19 is done, the 0.20 release will be focused on Image changes.

kidroca:
✅ rebased

packages/react-native-web/src/exports/Image/__tests__/index-test.js
```diff
@@ -248,8 +248,8 @@ describe('components/Image', () => {
       });
     });
 
-    test('is not set immediately if the image has not already been loaded', () => {
-      const uri = 'https://google.com/favicon.ico';
+    test('is set immediately while images is loading and there is no default source', () => {
+      const uri = 'https://google.com/not-yet-loaded-image.ico';
       const source = { uri };
       const { container } = render(<Image source={source} />);
```
kidroca, line 254: Perhaps this was originally true, and source was not rendered until loaded, but if you inspect the snapshot you can see that `` with the source was rendered, so I renamed the test

We do render the source while loading, unless there's a default source
https://github.com/necolas/react-native-web/blob/e8098fd029102d7801c32c1ede792bce01808c00/packages/react-native-web/src/exports/Image/index.js#L210-L211

necolas:
One unit test snapshot is failing

kidroca:
> One unit test snapshot is failing

Sorry, should be fine now

Issue #2432, opened 2022-11-20
Flatlist with expensive items breaks scroll


LucioChavezFuentes:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the issue

I noticed this issue in an App called Expensify. It uses Flatlist for a chat of reports that supports text, images, emojis, and reports as items (and perhaps others that I am not aware of). These items have complex code to support interactions and features on them. As you scroll in the report chat in Web, especially if you scroll fast, you will notice scroll issues like scroll jumps, items appearing and disappearing, or items not showing at all.

https://user-images.githubusercontent.com/48106652/202822381-3730f8ea-adfe-4e2e-a60c-cba09c3c29f4.mp4

### Expected behavior

Naturally, expensive items will take time to show, but you should find no issues on scroll fast.

### Steps to reproduce

Steps to Reproduce:

1. Have a Flatlist with expensive items and enough items to go beyond the virtual area.
2. Use the mouse and drag the scroll bar up and down.
3. Things should start looking weird: scroll jumps, items appearing and disappearing

### Test case

https://codesandbox.io/p/github/Lucio-s-Forks/react-native-web/only-longFlatlist-expensive-items-VirtualizedList-updated-no-proposal?file=%2Fpackages%2Freact-native-web-examples%2Fpages%2Findex.js&selection=%5B%7B%22endColumn%22%3A43%2C%22endLineNumber%22%3A90%2C%22startColumn%22%3A43%2C%22startLineNumber%22%3A90%7D%5D

### Additional comments

I found the solution to this issue. I discovered this issue is caused by three “sub-issues” and I will explain each of them along with their respective solution in react-native PRs:

1. $lead_spacer expands scroll artificially when VirtualizedList is mounting new items —> [Problem Explanation and Solution in this PR.](https://github.com/facebook/react-native/pull/35413)

2. VirtualizedList skip items for offset measuring when the user scrolls very fast while new items are mounted and measured —> [Problem Explanation and Solution in this PR.](https://github.com/facebook/react-native/pull/35414)

3. VirtualizedList gets offsets below or equal to zero for items that are not the list's first item —> [Problem Explanation and Solution in this PR.](https://github.com/facebook/react-native/pull/35415)

[...]

Issue #2424, opened 2022-11-09
onLayout width&height rounded numbers


burakgormek:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the issue

onLayout returns rounded numbers for width and height.

 **offsetWidth**, **offsetHeight** are the reason why it is returns rounded number.
`Note: This property will round the value to an integer.`
(https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/offsetHeight)

Might be one solution to fix this is using this approach to calculate width without include **transform**
https://stackoverflow.com/a/35543016

also worth to mention that **React Native** doesn't round the numbers for width&height.

### Expected behavior

value without rounded

### Steps to reproduce

Open codesandbox link:
1: Look console to see View's rounded width.

### Test case

https://codesandbox.io/s/vigorous-glitter-u2pddb

necolas:
Closing this as all the number will be rounded now, as part of fixing significant differences in values. There's no way to get sub pixel values with the web APIs necessary to fix other issues.

Jackman3005:
Hi @necolas we are having an issue where we measure a component and use the resulting `pageX` and `pageY` to position another absolute positioned element. On Web things are either spot on or we are getting slight mis-alignments between the components depending on the screen width of the browser. 

When I open the inspect element and try adjusting the `left: ...` value up or down by one, these do not align with the original component. Only if I adjust it by a decimal amount (e.g. `0.5`) can I get it to show perfectly aligned. This problem does not occur on native and I suspect it is because I see the values are coming through in decimal format.

It appears that web browsers to allow fractional pixel values as I was able to find a number of components with calculated dimensions that had fractional values.

I've looked around and I don't know what options I have left to get this to align correctly. I'm guessing that if I had access to the raw numbers I could get the result I'm looking for.

Any thoughts or suggestions?

Jackman3005:
For those that are running into this as well. I was able to develop a workaround by making use of `react-native-reanimated`'s [`measure`](https://docs.swmansion.com/react-native-reanimated/docs/advanced/measure) API. The values returned are not rounded which means I get perfect alignment now. 

In my situation, this is a step in the right direction anyways as reanimated allows me to make measurements on the UI thread, store the result in a shared value and then pass that down to the child component which can use `useAnimatedStyle(..)` using the shared value which keeps everything operating on the UI thread helping to eliminate stutter/jitter effects.

Issue #2321, opened 2022-06-29
Scrolling to index in lists uses wrong offset when content sizes has changed


AndreasHogstrom:
**The problem**

Scrolling to index positions in lists (FlatList, SectionList, VirtualizedList) uses the wrong offset if the sizes of items before the wanted index have changed size.

**How to reproduce**

Simplified test case: https://codesandbox.io/s/practical-khorana-gfcu1z?file=/src/App.js

Steps to reproduce:
Click the 3 buttons at the top of the example in order and observe the behaviour of the FlatList.

**Expected behavior**

ScrollToIndex (and related functions) should scroll to the correct offset even when content sizes dynamically change.

**Environment (include versions). Did this work in previous versions?**

* React Native for Web (version): 0.18.2
* React (version): 17.0.2
* Browser: Chrome, Safari (Doesn't seem browser related)

huuanhopp:
i have same issue

PR #2501, opened 2023-03-22
Ignore transformation and cancel callback measure for unmounted elements

LucioChavezFuentes:
Hello again @necolas, this the updated proposal to fix the VirtualizedList in web.

After investigation and feedback I have a new proposal of pull requests to fix the Inverted VirtualizedList in web. Some of the root issues are in react-native-web's end, another issue is fixed updating [VirtualizedList from react-native](https://github.com/facebook/react-native/tree/main/packages/virtualized-lists/Lists):

- **Problem 1**: VirtualizedList's code relies on that `onLayout` measures items without considering transformations **—>** **Problem Explanation and Solution below.**

- **Problem 2**: Exists inaccurate measures in onLayout when the setTimeout is executed and any of the targeted nodes (node or relativeNode) for measure are unmounted. **—>** **Problem Explanation and Solution below.**

- **Problem 3**: The scroll position can be forced to areas where items have not been measured and the list skips measuring items if the user scrolls too fast. **—>** [Problem Explanation and Solution in this PR](https://github.com/necolas/react-native-web/pull/2502).

---

### Problem 1 ###

VirtualizedList's code relies on that `onLayout` measures items without considering transformations 

React-native-web's onLayout, unlike its native counterpart, accounts for element transformations and its ancestors which leads to inaccurate measures for Inverted VirtualizedList with "scaleY / X: -1" transformation.

**Solution:**

Compose onLayout with only Web measurement API that ignores transformation following the W3C and WHATGH standards.

Screenshot 2023-03-22 at 13 52 37

#### I know the solution is similar to [@davepack's PR](https://github.com/necolas/react-native-web/pull/955/files) so Why should we merge it now? What is different? ####

The @davepack's solution is good, the problem back then was a bug in Firefox. Firefox [wasn't following the standard specification for offsetTop](https://w3c.github.io/csswg-drafts/cssom-view/#dom-htmlelement-offsettop). It wasn't fixed until Firefox >71 (the @davepack PR was tested in Firefox 61).

Now I can confirm the PR's changes works in Firefox >71 and in the rest of the browsers including the mobile versions. I [tested multiple scenarios here](https://test-new-measure-implementation-luciochavezfuentes.vercel.app/) and also tested in the @davepack sandbox.

The solution works the same as getBoundingclientrect but without considering transformations with a margin of error of +/- 1 pixel because the variables like offsetTop and clientTop round the values and return integers only.

#### How should we handle problems like this? ####

[WHATGH](https://html.spec.whatwg.org/multipage/) is a standard specification for browsers created by the browsers' creators (Mozilla, apple, Microsoft). It reached an [agreement with W3C ](https://www.w3.org/2019/04/WHATWG-W3C-MOU.html)and now both standards [reference each other](https://html.spec.whatwg.org/multipage/references.html) and recommend how the browser API should work.

Different results among browsers with the same code is likely that some them are not following the WHATGH and W3C specifications and can be reported as a bug in the browser's end.

### Problem 2 ###

Exists inaccurate measures in onLayout when the [setTimeout](https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/UIManager/index.js#L26) is executed and any of the targeted nodes (node or relativeNode) for measure are unmounted.

**Solution**

Check if the layout boxes of nodes are still present inside the setTimeout callback.
Screenshot 2023-03-27 at 11 59 35

### Problem 3 **—>** [Problem Explanation and Solution in this PR](https://github.com/necolas/react-native-web/pull/2502).

With all problems resolved and updating VirtualizedList from react-native we have a solid VirtualizedList in web, inverted or non-inverted.

Thank you for reading @necolas, if you have any questions or concerns please let me know how I can help.

### DIFF
--- a/packages/react-native-web/src/exports/UIManager/index.js
+++ b/packages/react-native-web/src/exports/UIManager/index.js
@@ -7,28 +7,38 @@
  * @noflow
  */
 
-import getBoundingClientRect from '../../modules/getBoundingClientRect';
 import setValueForStyles from '../../modules/setValueForStyles';
 
 const getRect = (node) => {
-  // Unlike the DOM's getBoundingClientRect, React Native layout measurements
-  // for "height" and "width" ignore scale transforms.
-  // https://developer.mozilla.org/en-US/docs/Web/API/CSS_Object_Model/Determining_the_dimensions_of_elements
-  const { x, y, top, left } = getBoundingClientRect(node);
-  const width = node.offsetWidth;
   const height = node.offsetHeight;
-  return { x, y, width, height, top, left };
+  const width = node.offsetWidth;
+  let left = node.offsetLeft;
+  let top = node.offsetTop;
+  node = node.offsetParent;
+
+  while (node && node.nodeType === 1 /* Node.ELEMENT_NODE */) {
+    left += node.offsetLeft + node.clientLeft - node.scrollLeft;
+    top += node.offsetTop + node.clientTop - node.scrollTop;
+    node = node.offsetParent;
+  }
+
+  top -= window.scrollY;
+  left -= window.scrollX;
+
+  return { width, height, top, left };
 };
 
 const measureLayout = (node, relativeToNativeNode, callback) => {
   const relativeNode = relativeToNativeNode || (node && node.parentNode);
   if (node && relativeNode) {
     setTimeout(() => {
-      const relativeRect = getBoundingClientRect(relativeNode);
-      const { height, left, top, width } = getRect(node);
-      const x = left - relativeRect.left;
-      const y = top - relativeRect.top;
-      callback(x, y, width, height, left, top);
+      if (node.isConnected && relativeNode.isConnected) {
+        const relativeRect = getRect(relativeNode);
+        const { height, left, top, width } = getRect(node);
+        const x = left - relativeRect.left;
+        const y = top - relativeRect.top;
+        callback(x, y, width, height, left, top);
+      }
     }, 0);
   }
 };


packages/react-native-web/src/exports/UIManager/index.js
```diff
@@ -7,28 +7,46 @@
...
+    node = node.offsetParent;
+  }
+
+  top -= window.scrollY;
+  left -= window.scrollX;
+
+  return { width, height, top, left };
 };
 
+function hasLayoutBox(node) {
```
janicduplessis, line 31: Can we use `node.isConnected` instead (https://developer.mozilla.org/en-US/docs/Web/API/Node/isConnected)?

LucioChavezFuentes: Didn't know about that API, I will give a try. Thank you for this insight.

necolas:
Thanks again for the clear and detail-oriented work. Great job with your test cases. Since this is technically a "fix" to match RN, do you think we can put this out in a patch release (0.19.1)?

LucioChavezFuentes:
> Thanks again for the clear and detail-oriented work. Great job with your test cases. Since this is technically a "fix" to match RN, do you think we can put this out in a patch release (0.19.1)?

Sure, so I can say this PR will eventually be merged?

necolas:
Yes, I can get it out in 0.19.1 today if you rebase your PR

necolas:
Also, FYI https://github.com/necolas/react-native-web/issues/2424

LucioChavezFuentes:
@necolas, the PR has been rebased!

LucioChavezFuentes:
The PR fixes these issues:
- https://github.com/necolas/react-native-web/issues/1254
- https://github.com/necolas/react-native-web/issues/2432 (partially, also need https://github.com/necolas/react-native-web/pull/2502 to be fully fixed)

This PR does **not** fix the following:

- Issue about rounded values https://github.com/necolas/react-native-web/issues/2424
- Issue about wrong offsets https://github.com/necolas/react-native-web/issues/2321

As for this issue: 

- https://github.com/necolas/react-native-web/issues/2109

The measures are correct with this PR's changes and with the current implementation (as long as there are no transformations) but the `scrollTo` of `ScrollView` does not work when the scrollView overflows only the window and not an element.

PR #2503, opened 2023-03-27
Implement TurboModuleRegistry

johnf:
This is a super naive implementation. I don't have any native experience.

I based this on the react-native implementation, which falls back to native based on the bridge type.
I remove all the checking code and always fall back to nativemodules.

From what I can tell, we could probably get by just having ```get``` always return null;

This was enough to get `react-native-vector-icons`, which recently got new arch support, to work in storybook.


johnf:
@necolas I figured it might be appropriate, given the equivalent exists for NativeModules. No problem, I'll create a PR for react-native-vector-icons

PR #2508, opened 2023-04-10
0.20 preview

necolas:
0.20 will be focused on the "Image" milestone: https://github.com/necolas/react-native-web/milestone/18

# Notes

### StyleSheet runtime for compilers

`babel-plugin-react-native-web` has a new option `legacy`, which is `true` by default. Set it to `false` and the plugin will no longer rewrite any `'react-native'` paths, with the exception of the `StyleSheet` import.

```
{
  "plugins": [
    [babelPluginReactNativeWeb, { "legacy": false }]
  ]
}
```

The `StyleSheet` import is rewritten to `[..]/StyleSheet/runtime`.

```js
import { StyleSheet } from 'react-native'
// => import StyleSheet from 'react-native-web/dist/exports/StyleSheet/runtime'
```

The runtime is only `StyleSheet()` without `create`, `compose`, `flatten`, etc., support.

```js
// errors at runtime
StyleSheet.create({})
```

The `legacy:false` option should be used only if a built-time compiler is being used to extract styles to CSS, and either inline results or error the other `StyleSheet` method calls React Native includes. The compiler output should follow the style object format described by the [styleq](https://github.com/necolas/styleq) runtime.

This is an experiment to add a mechanism to exclude the runtime style compiler from builds. It's open to feedback and iteration.

----

# Changelog

## New features

* Image: support for ImageSource with headers.
* Image: fix image caching.
* StyleSheet: option to remove runtime compiler.


programmrz:
I would like to use the preview in my application as I would love to take advantage of the Image http headers. I'll admit im not well versed in git... how would I reference this instead of 0.19? 0.20-dev doesnt appear to change anything

Issue #2523, opened 2023-05-22
[Animated] Regression runtime error if passing undefined to Animated.View style


Kudo:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the issue

If passing `undefined` to Animated.View style, there will be a runtime error. This issue does not happen on react-native-web 0.18. I think this should be a regression issue.

Example code:

```jsx
function App() {
  return (
    <View style={{ flex: 1 }}>
      <Animated.View style={{ height: 64, transform: undefined }}>
        <View style={{ width: 50, height: 50, backgroundColor: "green" }} />
      </Animated.View>
    </View>
  );
}
```

Error stacktrace:

```
AnimatedTransform.js:62 Uncaught TypeError: Cannot read properties of undefined (reading 'forEach')
    at AnimatedTransform.__attach (AnimatedTransform.js:62:1)
    at AnimatedTransform.__addChild (AnimatedWithChildren.js:34:1)
    at AnimatedStyle.__attach (AnimatedStyle.js:86:1)
    at AnimatedStyle.__addChild (AnimatedWithChildren.js:34:1)
    at AnimatedProps.__attach (AnimatedProps.js:63:1)
    at new AnimatedProps (AnimatedProps.js:29:1)
    at useAnimatedProps.js:29:1
    at mountMemo (react-dom.development.js:17225:1)
    at Object.useMemo (react-dom.development.js:17670:1)
    at useMemo (react.development.js:1650:1)
```

### Expected behavior

the undefined style should not raise runtime error.

### Steps to reproduce

I have a minimal reproducible example created from create-react-app

```sh
$ git clone https://github.com/Kudo/repro-rn-web-animated-transform.git
$ cd repro-rn-web-animated-transform
$ npm i
$ npm start
```

### Test case

https://github.com/Kudo/repro-rn-web-animated-transform/

necolas:
In the future, please create test cases using the link provided in the issue template. It saves me a lot of time and inconvenience. Creating a separate repo isn't helpful

Kudo:
sorry for the inconvenience. i did follow the issue template. in the "Test case" field, does it require a codesandbox rather than a separate repo?

```
  - type: input
    attributes:
      label: Test case
      description: Please provide a link to a reduced test case that reproduces the issue.
      placeholder: "https://codesandbox.io/s/6lx6ql1w5r"
    validations:
      required: true
```

necolas:
[...]

PR #2530, opened 2023-05-31
fix(AnimatedStyle): runtime error when passing undefined transform style

tarunrajput:
Fixes: #2523 
If passing undefined to Animated.View transform style, there will be a runtime error. This issue does not happen on react-native-web 0.18.


tarunrajput:
@necolas, as per RN [docs](https://reactnative.dev/docs/next/transforms#transform), transform accepts an array of transformation objects or space-separated string values. But I am getting same runtime error with string values.

image

Issue #2391, opened 2022-09-02
Unable to interact with nested Pressables if parent is disabled


SkySails:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the issue

Hovering Pressables inside of another, _disabled_ pressable no longer works.

### Expected behavior

I would expect this behavior to be the same as in `0.17.x`, where hovering of Pressables nested inside a disabled Pressable was possible.

### Steps to reproduce

1. Install `react-native-web@0.17.7` 
2. Wrap a pressable within another pressable, and add a visual hover state to it through a nested view or similar.
3. Disable the parent pressable
4. Attempt to trigger the hover-state for the nested pressable. You should be successful!

then...

1. Install `react-native-web@0.18.0`
2. Repeat steps 2 & 3 from the previous steps
3. Attempt to trigger the hover state for the nested pressable. Nothing should happen.

This scenario has already been set up in the provided CodeSandbox. Just change the version of the `react-native-web` package!

### Test case

https://codesandbox.io/s/react-native-web-pressable-nesting-zjo2sz

### Additional comments

I can't seem to find any references to this particular change, so I'm not sure whether this is a regression or an intended behavior change.

I'm very thankful for this project and all the work that goes into improving it! 🙌

necolas:
It will have been this change https://github.com/necolas/react-native-web/commit/094bd0efc5ac1c566a81dec7ee825deae2c5a9a7

I'm not sure why the contents of a disabled element would be expected to be interactive. If anything, it was a bug in 0.17 that they were

SkySails:
I understand and I kind of agree! My use case is something like the following:

![pressables](https://user-images.githubusercontent.com/46646495/188213097-2d06f806-0c55-4d73-9198-55938415ccb0.png)

I want the area encircled with red color to be clickable to toggle the collapsible. I also want the buttons encircled in blue to be visible and interactive at all times. As mentioned, this can be worked around - but not as convenient as before. 

Here are two possible alternatives:
<p float="left">
  <img src="https://user-images.githubusercontent.com/46646495/188214160-ec904f1a-53bd-41f5-b003-95aa5e059bfb.png" width="370" />
  <img src="https://user-images.githubusercontent.com/46646495/188214226-6f60d5ef-d17b-41ea-8728-4a2f2fddff2c.png" width="370" />
</p>

[...]

PR #2536, opened 2023-06-13
(fix) Make children interactive when parent pressable is disabled

jennifersalas:
This PR addresses the following issue: https://github.com/necolas/react-native-web/issues/2391.

It fixes the inconsistency between native builds and react-native-web, where when disabling Pressables and other tappable elements, the children on native builds are still interactable (swiping, tapping, etc) but un-interactable in react-native-web. On react-native-web, the pointerEvents were set to 'none'  to address https://github.com/necolas/react-native-web/commit/094bd0efc5ac1c566a81dec7ee825deae2c5a9a7, which causes this inconsistency.

The solution was to change from 'none' to 'box-none', which maintains the desired behavior of the original issue and addresses the current.


necolas:
Thanks for looking into this and putting up a PR

necolas:
This fix is included in 0.19.5

PR #2542, opened 2023-06-22
fix: add support for caretHidden prop in TextInput

### LINKED ISSUE #2541: caretHidden prop is not supported for TextInput

### Describe the issue

The [caretHidden](https://reactnative.dev/docs/0.67/textinput#carethidden) prop for TextInput component is not supported in `react-native-web`. 

### Expected behaviour

Adding caretHidden prop to a TextInput component should hide the text insertion caret on browser 

### Steps to reproduce

`<TextInput caretHidden />` should render an input with no visible caret.

### Test case

https://codesandbox.io/s/goofy-https-tjkghf?file=/src/App.js

### Additional comments

_No response_

aswin-s:
## Details

This PR adds support for `caretHidden` prop in TextInput component. 

Adding `caretHidden` prop should ideally hide the text insertion caret on browsers. This is not working right now. This PR intends to fix the issue by setting `caretColor` as `transparent` while this prop is used.

fixes #2541

### DIFF
--- a/packages/react-native-web/src/exports/TextInput/index.js
+++ b/packages/react-native-web/src/exports/TextInput/index.js
@@ -99,6 +99,7 @@ const TextInput: React.AbstractComponent<
     autoCompleteType,
     autoCorrect = true,
     blurOnSubmit,
+    caretHidden,
     clearTextOnFocus,
     dir,
     editable,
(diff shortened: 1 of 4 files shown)


packages/react-native-web/src/exports/TextInput/index.js
```diff
@@ -402,6 +403,7 @@ const TextInput: React.AbstractComponent<
     { '--placeholderTextColor': placeholderTextColor },
     styles.textinput$raw,
     styles.placeholder,
+    caretHidden && styles.caretHidden,
```
s77rt, line 406: Shouldn't this style be the last not to be overwritten by props.style?

PR #2555, opened 2023-07-16
Remove StyleSheet.compose(a, b) instances in favor of lists

berkcoker:
`StyleSheet.compose(a, b) is deprecated; use array syntax, i.e., [a,b]. `


necolas:
Thanks!

PR #2558, opened 2023-07-17
Fix Modal onDismiss callback is called too early

bernhardoj:
### Description
Currently, `onDismiss` is called right after setting the `isRendered` state to `false` which leads to `onDismiss` being called before the modal is completely dismissed. This makes chaining action after the modal is dismissed fails, for example, focusing an input.

### Test plan
1. Add a text input to the modal example
2. Add `onDismiss` props to the Modal which will focus the text input
3. Open the modal. Verify the modal shows
4. Close the modal
5. Verify the modal closes and the text input is focused

Currently, the 5th step fails and this PR is trying to fix it.

**Before:**

https://github.com/necolas/react-native-web/assets/50919443/40c72c36-2354-48f6-a5a6-997a6dfdba9f

**After:**

https://github.com/necolas/react-native-web/assets/50919443/49452b30-e9d0-4176-9d9f-5462da4cb9c7

### DIFF
--- a/packages/react-native-web/src/exports/Modal/ModalAnimation.js
+++ b/packages/react-native-web/src/exports/Modal/ModalAnimation.js
@@ -37,6 +37,7 @@ function ModalAnimation(props: ModalAnimationProps): React.Node {
 
   const [isRendering, setIsRendering] = React.useState(false);
   const wasVisible = React.useRef(false);
+  const wasRendering = React.useRef(false);
 
   const isAnimated = animationType && animationType !== 'none';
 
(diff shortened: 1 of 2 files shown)


necolas:
Is there any way we can add unit test coverage for this?

necolas:
Try to create a unit test that reproduces the test case and fails, then show that it passes with your patch. You didn't share a test case for the issue, so I don't know what it takes to reproduce, but it seems like it's related to animations. A unit test combining animation props with onDismiss might surface the issue

https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/Modal/__tests__/index-test.js

necolas:
Looks good, thanks

Issue #2549, opened 2023-07-07
Switch component doesn't respect aria-label property: must use deprecated accessibilityLabel


kpcurai:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the issue

The `aria-label` prop is applied to the enclosing `View` component of a `Switch`, which is not an accessible pattern.

When the deprecated `accessibilityLabel` prop is used, it correctly is applied to the native `input` component, but makes loud deprecation warnings.

### Expected behavior

I would expect an aria-label prop to pass through to the input element, not be placed on the `View` component.

### Steps to reproduce

* version 0.19.4
* use the Switch component with aria-label prop and observe it does not get applied to the input element
* use the Switch component with accessibilityLabel and observe it does get applied to the input element

### Test case

https://codesandbox.io/s/switch-deprecated-accessibilitylabel-9v6md8?file=/src/App.js

PR #2559, opened 2023-07-17
fix(Switch): aria-label prop not applying to native control

tarunrajput:
Fixes: https://github.com/necolas/react-native-web/issues/2549

Fix accessibility issue in Switch component by applying the `aria-label` prop to the native control component

### DIFF
--- a/packages/react-native-web/src/exports/Switch/index.js
+++ b/packages/react-native-web/src/exports/Switch/index.js
@@ -44,6 +44,7 @@ const Switch: React.AbstractComponent<
   React.ElementRef<typeof View>
 > = React.forwardRef((props, forwardedRef) => {
   const {
+    'aria-label': ariaLabel,
     accessibilityLabel,
     activeThumbColor,
     activeTrackColor,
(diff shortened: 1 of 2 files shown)


packages/react-native-web/src/exports/Switch/index.js
```diff
@@ -166,6 +167,7 @@ const Switch: React.AbstractComponent<
   ];
 
   const nativeControl = createElement('input', {
+    'aria-label': ariaLabel,
     accessibilityLabel,
```
necolas, line 171: We should remove this prop

packages/react-native-web/src/exports/Switch/index.js
```diff
@@ -166,6 +167,7 @@ const Switch: React.AbstractComponent<
   ];
 
   const nativeControl = createElement('input', {
+    'aria-label': ariaLabel,
```
necolas, line 170: This should probably be `ariaLabel || accessibilityLabel` for now

PR #2561, opened 2023-07-18
fix: handle undefined for aspect ratio

itsramiel:
## Summary
[AspectRatio can technically be passed as `undefined` in the stylesheet](https://github.com/facebook/react-native/blob/080a3d300dc357fec62d767d140f964f9db66b2e/packages/react-native/Libraries/StyleSheet/StyleSheetTypes.d.ts#L46) and that is handled on ios/android.

Passing `undefined` to `aspectRatio` on web leads to the following error:
`TypeError: Cannot read properties of undefined (reading 'toString')`

While it does not make sense as to why you might explicitly want to pass undefined to aspectRatio, my use case is that I had a custom component that was setting the aspectRatio as well as accepting styles prop that overrides the defined styles so I had to pass undefined for aspectRatio. This works fine or ios/android but crashed on web.

I also added a test case

### DIFF
--- a/packages/react-native-web/src/exports/StyleSheet/preprocess.js
+++ b/packages/react-native-web/src/exports/StyleSheet/preprocess.js
@@ -182,7 +182,7 @@ export const preprocess = <T: {| [key: string]: any |}>(
       continue;
     }
 
-    if (prop === 'aspectRatio') {
+    if (prop === 'aspectRatio' && value !== undefined) {
       nextStyle[prop] = value.toString();
     } else if (prop === 'fontVariant') {
       if (Array.isArray(value) && value.length > 0) {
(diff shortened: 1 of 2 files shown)


packages/react-native-web/src/exports/StyleSheet/preprocess.js
```diff
@@ -182,7 +182,7 @@ export const preprocess = <T: {| [key: string]: any |}>(
       continue;
     }
 
-    if (prop === 'aspectRatio') {
+    if (prop === 'aspectRatio' && value !== undefined) {
```
necolas, line 185: This check should probably be `!= null`

itsramiel: The thing is aspectRatio's type is either string, number, or undefined. If null was used, then a type error should be shown I guess, but undefined is valid. That's why I put a check for undefined. If we dont check for undefined, then the app crashes.

itsramiel: We can do (typeof aspectRatio === 'number' || typeof aspectRatio === 'string') if you want. This way we check it is what we want rather than checking the opposite

Issue #1703, opened 2020-08-14
View collapsable prop seems not work


TQCasey:
View collapsable prop seems not work

necolas:
It doesn't do anything on web

PR #2569, opened 2023-08-09
Remove collapsable props from Animated componnet

shubham1206agra:
As per https://github.com/necolas/react-native-web/issues/1703.

Since `collapsable` does nothing anywhere. This PR will remove the prop from `reduceAnimatedProps` function, which will help us to avoid warnings like these.

https://github.com/software-mansion/react-native-svg/issues/1484


shubham1206agra:
@necolas Bump on this

PR #2570, opened 2023-08-10
fix: measureLayout relative to ScrollView returns incorrect y value

### LINKED ISSUE #2109: `measureLayout` relative to `ScrollView` returns incorrect `y` value
**The problem**

Calling `measureLayout` relative to a `ScrollView` returns the incorrect `y` value when the scroll position is greater than 0. 

**How to reproduce**

Simplified test case: https://codesandbox.io/s/lucid-sea-rq0ny?file=/src/App.js:0-1339

Steps to reproduce:
1. Click on a number in the reproduction. This should scroll to it.
2. Look in the console. It outputs the location it should scroll to vs the result of `measureLayout`.
3. Notice that the `y` value returned from `measureLayout` is incorrectly subtracting the current scroll position. 

**Expected behavior**

The behavior should match that of native. Try the **iOS** here, and notice that it always scrolls to the correct spot: https://snack.expo.dev/@beatgig/dedc14

On native, calling `measureLayout` relative to a `ScrollView` gives the `y` coordinate relative to the top of the the scrollable content, regardless of scroll position.

* React Native for Web (version): `0.17.0`
* React (version): `17.0.2`
* Browser: Chrome

shubham1206agra:
Trying to fix #2109 

I will put details once I confirm this works perfectly.

### DIFF
--- a/packages/react-native-web/src/exports/ScrollView/index.js
+++ b/packages/react-native-web/src/exports/ScrollView/index.js
@@ -698,10 +698,7 @@ class ScrollView extends React.Component<ScrollViewProps> {
 
   _setInnerViewRef = (node) => {
     this._innerViewRef = node;
-  };
 
-  _setScrollNodeRef = (node) => {
-    this._scrollNodeRef = node;
     // ScrollView needs to add more methods to the hostNode in addition to those
     // added by `usePlatformMethods`. This is temporarily until an API like
     // `ScrollView.scrollTo(hostNode, { x, y })` is added to React Native.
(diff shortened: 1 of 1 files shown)


shubham1206agra:
@necolas I think it's working fine.
https://codesandbox.io/s/issue-react-native-for-web-forked-9zytqk?file=/src/App.js
Can you confirm the same?

necolas:
> I think it's working fine

The existing unit tests fail. PR needs to pass unit tests and add new tests for the issue that 1) fail on master, 2) pass with this patch.

PR #2576, opened 2023-08-28
fix(web)[TextInput]: cursor jumps to the start when secureTextEntry is toggled

pac-guerreiro:
This PR fixes https://github.com/Expensify/App/issues/21727

# Problem

On web, the cursor jumps to the start of the input when `secureTextEntry` is toggled.

# Solution

Preserve and restore previous selection after `secureTextEntry` is toggled.

The proposal and demo are available in this comment: https://github.com/Expensify/App/issues/21727#issuecomment-1642507490

### DIFF
--- a/packages/react-native-web/src/exports/TextInput/index.js
+++ b/packages/react-native-web/src/exports/TextInput/index.js
@@ -332,8 +332,7 @@ const TextInput: React.AbstractComponent<
 
   function handleSelectionChange(e) {
     try {
-      const node = e.target;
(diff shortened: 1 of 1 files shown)


necolas:
Please include unit tests

pac-guerreiro:
@necolas

I'm unable to reproduce this bug using testing-library, so I cant write unit tests that ensure that this fix works.

My suspicion is that the test environment doesn't replicate the browser behaviour which leads to this bug.

necolas:
To report an issue open a new issue

Issue #1651, opened 2020-06-26
Label tag should allow for htmlFor


walidvb:
**Is your feature request related to a problem? Please describe.**
`accessibilityRole` allows to have better tags, such as `<label>`, along with aria labels. One thing that is missing is the `for` attribute, which allows for the user to click on the label to focus the field.

**Describe a solution you'd like**
Passing htmlFor to a `<Text accessibilityRole="label" htmlFor="email">Email</Text>` would yield the expected html

**Additional context**
While it doesn't concern UX directly, I use of capybara for my automated e2e testing, which heavily relies on this feature to interact with forms

I'm currently patching a follows:
```diff
diff --git a/node_modules/react-native-web/dist/exports/View/filterSupportedProps.js b/node_modules/react-native-web/dist/exports/View/filterSupportedProps.js
index a7a6b96..5a087a7 100644
--- a/node_modules/react-native-web/dist/exports/View/filterSupportedProps.js
+++ b/node_modules/react-native-web/dist/exports/View/filterSupportedProps.js
@@ -66,6 +66,7 @@ var supportedProps = {
   onMouseUp: true,
   // unstable escape-hatches for web
   href: true,
+  htmlFor: true,
   itemID: true,
   itemRef: true,
   itemProp: true,
diff --git a/node_modules/react-native-web/dist/modules/createDOMProps/index.js b/node_modules/react-native-web/dist/modules/createDOMProps/index.js
index 57d9054..5ce1440 100644
--- a/node_modules/react-native-web/dist/modules/createDOMProps/index.js
+++ b/node_modules/react-native-web/dist/modules/createDOMProps/index.js
@@ -72,6 +72,7 @@ var createDOMProps = function createDOMProps(component, props, styleResolver) {
       testID = _props.testID,
       accessible = _props.accessible,
       accessibilityRole = _props.accessibilityRole,
+      htmlFor = _props.htmlFor,
       domProps = _objectWithoutPropertiesLoose(_props, ["accessibilityLabel", "accessibilityLiveRegion", "accessibilityRelationship", "accessibilityState", "classList", "className", "disabled", "importantForAccessibility", "nativeID", "pointerEvents", "style", "testID", "accessible", "accessibilityRole"]);
 
   var disabled = accessibilityState != null && accessibilityState.disabled === true || providedDisabled;
@@ -142,7 +143,9 @@ var createDOMProps = function createDOMProps(component, props, styleResolver) {
     } else {
[...]

PR #2595, opened 2023-10-18
fix: role='label' not changing element to label.

Viraj-10:
### Problem Statement
with `accessibilityRole="label`
Code
```
<View accessibilityRole="label">
```
OutPut
```
<label />
```

with `role="label`

Code
```
<View role="label">
```
OutPut
```
<div />
```
### Solution

- Added condition for `role="label"`.

### DIFF
--- a/packages/react-native-web/src/modules/AccessibilityUtil/propsToAccessibilityComponent.js
+++ b/packages/react-native-web/src/modules/AccessibilityUtil/propsToAccessibilityComponent.js
@@ -37,7 +37,7 @@ const propsToAccessibilityComponent = (
   props: Object = emptyObject
 ): void | string => {
   // special-case for "label" role which doesn't map to an ARIA role
-  if (props.accessibilityRole === 'label') {
+  if (props.accessibilityRole === 'label' || props.role === 'label') {
     return 'label';
   }
 


albertcito:
Looks like it is already implemented. I read this [comment](https://github.com/necolas/react-native-web/issues/1651#issuecomment-856207995)


>You can still output a <label> element by setting the accessibilityRole to "label". That could be used to wrap fields in a <label>. IIRC not every screen reader likes that pattern, which is why using accessibilityLabelledBy (web-only) along with JS to move the focus has benefits.

I added it to my code and it worked:

`accessibilityRole="label"`

Viraj-10:
> Looks like it is already implemented. I read this [comment](https://github.com/necolas/react-native-web/issues/1651#issuecomment-856207995)
> 
> > You can still output a  element by setting the accessibilityRole to "label". That could be used to wrap fields in a . IIRC not every screen reader likes that pattern, which is why using accessibilityLabelledBy (web-only) along with JS to move the focus has benefits.
> 
> I added it to my code and it worked:
> 
> `accessibilityRole="label"`

`accessibilityRole` prop was mapped to `role` in this [PR](https://github.com/facebook/react-native/pull/34538) with react-native. when you pass `role="label"` renders a <div> not a <label>.

necolas:
Duplicated by #2666

Issue #2596, opened 2023-10-18
TextInput incompatible with FormData due to missing name forwarding


TimPetricola:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the issue

I'm using `react-native-web` in a project targeting a web environment only.

To manage forms, in simple case I need to rely uncontrolled inputs and the [standard `FormData` API](https://developer.mozilla.org/en-US/docs/Web/API/FormData):

```jsx
const Form = () => {
  const handleSubmit = (e) => {
    e.preventDefault();
    const formData = new FormData(e.target);
    // do stuff to process formData, e.g. send it to API, ...
  }

  return (
    <form onSubmit={handleSubmit}>
      <TextInput name="firstName" defaultValue="Foo" />
      <Button>Submit</Button>
    </form>
}
```

In the above, I'm expecting `formData` to get a `firstName` entry with `"Foo"` value. But instead it's empty.

This is due to `<TextInput />` not forwarding the `name` prop to the underlying `<input />`

### Expected behavior

I'm expecting `TextInput` (and potentially other components) to forward `name` prop to the resulting DOM node.

### Steps to reproduce

See codesandbox

### Test case

https://codesandbox.io/s/react-native-web-input-name-2g7d2l

### Additional comments

Once it's validated that `name` should be forwarded. I can take care of opening a PR

TimPetricola:
React is also introducing 2 new hooks in an upcoming version (see documentation for [useFormState](https://react.dev/reference/react-dom/hooks/useFormState) and [useFormStatus](https://react.dev/reference/react-dom/hooks/useFormStatus)) which requires the `name` attribute.

I took the liberty of opening a PR to bring support to react-native-web: https://github.com/necolas/react-native-web/pull/2603

efoken:
FYI, we solved this issue using a custom TextInput component using a ref and then `ref.current.setAttribute('name', name)`. I also did this for setting the `type` attribute, also for `<button>`-elements

TimPetricola:
@efoken thanks for the tip! I didn't think about it and it's a great alternative! (I'm currently patching `react-native-web` for this, and I prefer your approach very much)

PR #2603, opened 2023-10-27
[fix] forward name prop

TimPetricola:
Fix  https://github.com/necolas/react-native-web/issues/2596

Add `name` to the list of forwarded props. This is important to have `input` element be compatible with standards:
- `FormData` web standard ([doc](https://developer.mozilla.org/en-US/docs/Web/API/FormData))
- new React `useFormStatus` hook ([doc](https://react.dev/reference/react-dom/hooks/useFormStatus))
- new React `useFormState` hook ([doc](https://react.dev/reference/react-dom/hooks/useFormState))

### DIFF
--- a/packages/react-native-web/src/modules/forwardedProps/index.js
+++ b/packages/react-native-web/src/modules/forwardedProps/index.js
@@ -12,6 +12,7 @@ export const defaultProps = {
   dataSet: true,
   dir: true,
   id: true,
+  name: true,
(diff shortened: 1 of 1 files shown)


mathieutu:
I'm working on a plain form with an old basic action URL (so redirected with a POST), unsuccessfully, and this was specifically what was missing. Thanks! 🙏

Issue #2528, opened 2023-05-29
[Animated] Cannot animate string transform properties (v0.19)


Aashu-Dubey:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the issue

It seems that after rn-web v0.19, it is recommended to use space-separated string functions on web instead of array, as we get these deprecation warnings

<img width="370" alt="image" src="https://github.com/necolas/react-native-web/assets/46301285/ccf3116d-c7db-48de-8c98-58aa6e974ff0">

though they work fine on web, but we can't apply animations using them as of now, as I'm getting this error even when using the styles without animation on any `Animated` component
<img width="909" alt="image" src="https://github.com/necolas/react-native-web/assets/46301285/e006a7bf-85fc-4184-a450-2c4e275feaac">

### Expected behavior

animation on these new string functions should work fine, as it is working in case of array transform right now.

### Steps to reproduce

You can clearly see the issue in the attached sandbox.

For a bigger example, I had the issue when I was recently trying to fix these warning on [this sample app](https://gitlab.com/ashud/rn_travel_app) ([Web Demo](https://ngefly.vercel.app/)), there are few places where I'm using these styles and they are working fine, but can't use them where animations are needed.

### Test case

https://codesandbox.io/s/rn-web-string-transform-bug-g1eqvy

### Additional comments

Currently I'm using `...Platform.select({ web : { transform: 'rotate(45deg)' } })` approach to apply these styles, as they don't yet seems to be cross compatible with native platforms.

Do let me know if there's any particular approach to apply animations here or if this is something to be fix in future version of the package.

Thanks!

chrispader:
Having the same issue in `>=0.19`

necolas:
This is a bug in React Native too [34425](https://github.com/facebook/react-native/issues/34425#issuecomment-1608512159)

dan-doyon-endear:
Here is patch that works for us in interim

```
diff --git a/node_modules/react-native-web/dist/vendor/react-native/Animated/nodes/AnimatedTransform.js b/node_modules/react-native-web/dist/vendor/react-native/Animated/nodes/AnimatedTransform.js
index 66072d9..b92ada1 100644
--- a/node_modules/react-native-web/dist/vendor/react-native/Animated/nodes/AnimatedTransform.js
+++ b/node_modules/react-native-web/dist/vendor/react-native/Animated/nodes/AnimatedTransform.js
[...]

PR #2622, opened 2023-12-19
remove transform style warning since the native alternative doesn't work for most cases.

EvanBacon:
The string interpolation system doesn't appear to work with React Animated as the string would convert the animated object to `[Object object]`.


necolas:
Re: interpolation https://github.com/necolas/react-native-web/issues/2528#issuecomment-1611964725

PR #2625, opened 2024-01-05
Fix potential infinite render

jwaldrip:


necolas:
This PR causes 20 unit tests for `Modal` to fail

necolas (changes requested):
This PR causes 20 unit tests for `Modal` to fail

Issue #2605, opened 2023-11-09
Pressable triggers onPress twice, when activated via Enter key and set to disabled


efoken:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the issue

When using the `onPress` prop on the `Pressable` component and setting it to `disabled` inside the `onPress`-function, the function will be called a second time. But only when activating the button via the <kbd>Enter</kbd> key.

First time `onPress` is called when pressing the <kbd>Enter</kbd> key down and the second time it is called when releasing the <kbd>Enter</kbd> key. I created a test case which demonstrates this behaviour.

### Expected behavior

`onPress` is only called once, like it is when using <kbd>Space</kbd> key oder pressing the button with the mouse.

### Steps to reproduce

1. Use `<Pressable role="button">` and apply an `onPress` handler to it
2. Set a `disabled` state inside the `onPress` handler
3. Set the `disabled` prop using the state on your `Pressable`
4. Put some `console.log` inside the `onPress` handler
5. Activate your button using the keyboard, pressing the <kbd>Enter</kbd> key
6. See your `console.log` twice in the console

### Test case

https://codesandbox.io/s/react-native-web-on-press-disabled-77crwk

seemX17:
Is there a workaround for this till the fix is provided?

jkjustjoshing:
@necolas any idea when a new release will go out that includes this fix?

necolas:
It's been out for a week already 54c14d64dabd175e8055e1dc92e9196c821f9b7d

PR #2632, opened 2024-01-24
Fix onPress is called twice when activated with keyboard and set to disabled (blurred)

bernhardoj:
### Description
Fixes https://github.com/necolas/react-native-web/issues/2605

We are attaching the `keyup` listener to the document.
https://github.com/necolas/react-native-web/blob/5f9c32eba0f840bde7b462a57d0748358ff866f6/packages/react-native-web/src/modules/usePressEvents/PressResponder.js#L348-L350

If the Pressable (with a button role) is disabled after the first `onPress` is triggered, the focus is transferred to `body` and `isNativeInteractiveElement` will be false when releasing the press, thus the `onPress` is called for the second time.
https://github.com/necolas/react-native-web/blob/5f9c32eba0f840bde7b462a57d0748358ff866f6/packages/react-native-web/src/modules/usePressEvents/PressResponder.js#L316-L326

### Test plan
1. On the pressable example code, disable the pressable in`onPress`
2. Run the pressable example
3. Hold down the pressable using a keyboard. Verify the `onPress` is called
4. Release the press. Verify the `onPress` isn't called.

Currently, the 4th step fails and this PR is trying to fix it.

**Before:**

https://github.com/necolas/react-native-web/assets/50919443/26a9a3f2-d220-4b0f-8642-8cfdba2370d0

**After:**

https://github.com/necolas/react-native-web/assets/50919443/44ca0796-9a2e-4efa-98b8-485b2dd6a285

### DIFF
--- a/packages/react-native-web/src/modules/usePressEvents/PressResponder.js
+++ b/packages/react-native-web/src/modules/usePressEvents/PressResponder.js
@@ -233,7 +233,7 @@ export default class PressResponder {
     pageY: number
   |}>;
   _touchState: TouchState = NOT_RESPONDER;
-  _responderElement: HTMLElement;
+  _responderElement: ?HTMLElement = null;
 
   constructor(config: PressResponderConfig) {
     this.configure(config);
--- a/packages/react-native-web/src/modules/usePressEvents/PressResponder.js
+++ b/packages/react-native-web/src/modules/usePressEvents/PressResponder.js
@@ -326,6 +326,8 @@ export default class PressResponder {
         if (onPress != null && !isNativeInteractiveElement && isActiveElement) {
           onPress(event);
         }
+
+        this._responderElement = null;
       }
     };
 
--- a/packages/react-native-web/src/modules/usePressEvents/PressResponder.js
+++ b/packages/react-native-web/src/modules/usePressEvents/PressResponder.js
@@ -233,6 +233,7 @@ export default class PressResponder {
     pageY: number
   |}>;
   _touchState: TouchState = NOT_RESPONDER;
+  _responderElement: HTMLElement;
 
   constructor(config: PressResponderConfig) {
     this.configure(config);
@@ -320,8 +321,9 @@ export default class PressResponder {
           elementType === 'input' ||
           elementType === 'select' ||
           elementType === 'textarea';
+        const isActiveElement = this._responderElement === target;
 
-        if (onPress != null && !isNativeInteractiveElement) {
+        if (onPress != null && !isNativeInteractiveElement && isActiveElement) {
           onPress(event);
         }
       }
@@ -345,6 +347,7 @@ export default class PressResponder {
         if (!disabled && isValidKeyPress(event)) {
           if (this._touchState === NOT_RESPONDER) {
             start(event, false);
+            this._responderElement = target;
             // Listen to 'keyup' on document to account for situations where
             // focus is moved to another element during 'keydown'.
             document.addEventListener('keyup', keyupHandler);
--- a/packages/react-native-web/src/exports/Pressable/__tests__/__snapshots__/index-test.js.snap
+++ b/packages/react-native-web/src/exports/Pressable/__tests__/__snapshots__/index-test.js.snap
@@ -61,14 +61,14 @@ exports[`components/Pressable hover interaction 3`] = `
 />
 `;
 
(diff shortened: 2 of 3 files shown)


packages/react-native-web/src/modules/usePressEvents/PressResponder.js
```diff
@@ -345,9 +321,6 @@ export default class PressResponder {
         if (!disabled && isValidKeyPress(event)) {
           if (this._touchState === NOT_RESPONDER) {
             start(event, false);
-            // Listen to 'keyup' on document to account for situations where
-            // focus is moved to another element during 'keydown'.
-            document.addEventListener('keyup', keyupHandler);
```
bernhardoj, line 350: I think if the focus is moved to another element, then the `onPress` shouldn't be triggered anymore on the blurred element. So, I attach back the key-up listener to the element.

bernhardoj: @necolas hi, sorry if it's disturbing you, curious to know your thoughts about the solution here to fix the double `onPress` event.

necolas: This change will regress the issue that is mentioned in the comment. The responder system has to "end" the current responder whenever keyup happens, so using `onKeyUp` will not work.

bernhardoj: Ah, I see. Thanks for the info!

> Maybe you just want to check again for disabled === true in the handler logic

Yes, this will fix for the disabled case, but the issue happens when the button lost focus too.

I'm thinking of 2 other solutions:
1. Instead of checking whether it's disabled or not, we can check whether the element is the active element or not. We will store the element when `onKeyDown` is triggered.
```diff
onKeyDown: (event) => {
  const { key, target } = event;
  if (!disabled && isValidKeyPress(event)) {
    if (this._touchState === NOT_RESPONDER) {
+     this.responderElement = target;
...

const keyupHandler = (event: KeyboardEvent) => {
  if (this._touchState !== NOT_RESPONDER && isValidKeyPress(event)) {
    end(event);
    document.removeEventListener('keyup', keyupHandler);
    ...
    const isNativeInteractiveElement = ...
+   const isActiveElement = this.responderElement === target;

+   if (onPress != null && !isNativeInteractiveElement && isActiveElement) {
      onPress(event);
    }
  }
```

or

2. Check `isNativeInteractiveElement` value in `onKeyDown` instead of `keyUpHandler`.
```diff
onKeyDown: (event) => {
  const { key, target } = event;
  if (!disabled && isValidKeyPress(event)) {
    if (this._touchState === NOT_RESPONDER) {
+     this.isNativeInteractiveElement = role === 'link' || ...
...
const keyupHandler = (event: KeyboardEvent) => {
  if (this._touchState !== NOT_RESPONDER && isValidKeyPress(event)) {
    end(event);
    document.removeEventListener('keyup', keyupHandler);
    ...
-   const isNativeInteractiveElement = ...

+   if (onPress != null && !this.isNativeInteractiveElement) {
      onPress(event);
    }
  }
```

Let me know what you think, thanks!

necolas: I don't know, maybe the first one. Up to you to do the testing and make sure your PR doesn't break existing expectations.

necolas (changes requested):
.

packages/react-native-web/src/modules/usePressEvents/PressResponder.js
```diff
@@ -345,6 +347,7 @@ export default class PressResponder {
         if (!disabled && isValidKeyPress(event)) {
           if (this._touchState === NOT_RESPONDER) {
             start(event, false);
+            this._responderElement = target;
```
necolas, line 350: Can we set this back to `null` when the interaction is over

necolas (changes requested):

packages/react-native-web/src/modules/usePressEvents/PressResponder.js
bernhardoj: Done 👍

bernhardoj:
@necolas hi, is there anything left to update?

necolas:
Sorry for the delay. Merged

bernhardoj:
Thanks!

PR #2638, opened 2024-02-02
fix(web): keep function identiy accross renders

Gregoirevda:
## Description
Currently when an inline function is passed to the `Image` or `Animated.Image` component, the `useEffect` will be triggered on **every** render, causing the image the be fetched on every render.

props accepting a function and declared in the dependency array are: `onLoad` `onLoadEnd` `onLoadStart` `onLoadError`.
It shouldn't be up to the developer to pass a memoized function.

The browser will cache subsequent image fetches (not sure about all platforms), so as it may seem acceptable setting the cache key to `reload` will fetch the image on every render.

## Example
```jsx
function App() {
  const Component = Image || Animated.Image;
 
  // Not memoized, causing the fetch on every render
  function bad() {}
  
  const good = useCallback(() => {}, [])

  return <Component source={{uri}} onLoadEnd={bad} onLoad={() => { // equally bad }} onLoadStart={good)  />
}
```

### DIFF
--- a/packages/react-native-web/src/vendor/react-native/Utilities/useCallbackRef.js
+++ b/packages/react-native-web/src/vendor/react-native/Utilities/useCallbackRef.js
@@ -1,0 +1,16 @@
+/*
+ *  Helper to keep the same ref of a function accross renders 
+ * 
+ */
+import { useLayoutEffect, useEffect, useRef } from 'react';
+
+export default function useCallbackRef(callback) {
+  const callbackRef = useRef(callback);
+  const useIsomorphicLayoutEffect = typeof window !== 'undefined' ? useLayoutEffect : useEffect;
+
+  useIsomorphicLayoutEffect(() => {
+    callbackRef.current = callback;
+  }, [callback]);
+
+  return callbackRef;
+}
--- a/packages/react-native-web/src/exports/Image/index.js
+++ b/packages/react-native-web/src/exports/Image/index.js
@@ -20,6 +20,7 @@ import StyleSheet from '../StyleSheet';
 import TextAncestorContext from '../Text/TextAncestorContext';
 import View from '../View';
 import { warnOnce } from '../../modules/warnOnce';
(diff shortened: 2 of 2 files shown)


Gregoirevda:
@necolas Does the fix make sense? I didn't know where to put the utility method so added in vendor to get feedback, but tell me where the right place is please

packages/react-native-web/src/vendor/react-native/Utilities/useCallbackRef.js
```diff
@@ -0,0 +1,16 @@
...
+ *  Helper to keep the same ref of a function accross renders 
+ * 
+ */
+import { useLayoutEffect, useEffect, useRef } from 'react';
+
+export default function useCallbackRef(callback) {
+  const callbackRef = useRef(callback);
+  const useIsomorphicLayoutEffect = typeof window !== 'undefined' ? useLayoutEffect : useEffect;
+
+  useIsomorphicLayoutEffect(() => {
```
lkinasiewicz, line 11: Is there any point delaying the ref update through an effect?

Gregoirevda: initial ref is set with `const callbackRef = useRef(callback);`

it then updates in the effect

lkinasiewicz: I understand that. I mean, you can simply update the ref without any effects. Simpler code and it guarantees that the most recent value of `callback` is assigned to the ref and will be called by Image - you don't know when exactly the effect will be executed and Image may stop loading between the render and the effect.

Gregoirevda: That's true. I'm wondering why `swr` does it [this](https://github.com/vercel/swr/blob/main/src/index/use-swr.ts#L575) way though? (They also don't have the config in the dependency array). There's also the `experimental_useEffectEvent` from react, so I guess there are edge cases with both solutions

Issue #2629, opened 2024-01-17
Confusion with `@react-native/normalize-colors` vs. `@react-native/normalize-color`


efoken:
### Is there an existing request?

- [X] I have searched for this request

### Describe the feature request

I think `@react-native/normalize-colors` (with S) should be used as a dependency, because `@react-native/normalize-color` seems to be the old package. Was probably renamed?!

necolas:
Oh, yes probably. If anyone wants to fix that in a PR, would be great

PR #2652, opened 2024-03-24
[fix] Correct @react-native/normalize-colors dependency

efoken:
Addresses https://github.com/necolas/react-native-web/issues/2629


Rexogamer:
duplicate of #2639

PR #2678, opened 2024-06-03
feat: add basic support for React Server Components

EvanBacon:
I'm adding React Server Components support to Expo Router, and need basic support added to React Native Web. This PR contains a cleaned up version of the patches I've been developing against for the last couple months. I just open sourced this branch which can be used to verify support: https://github.com/expo/expo/pull/29404/

Here are a few of the migration principles:

- `useState` and other dynamic hooks cannot be used in server components.
- `createContext` is not available in a `react-server` environment.
- `Component` and `PureComponent` are not allowed in `react-server`.
- `React.Children.toArray` appears to clobber transitional react elements. This caused a lot of headache since we perform the check in `<View />`
- You cannot dot-access into client modules. This wasn't a huge issue with rnw since most of the modules are already optimized for the server, but in `react-native` this means any use of `StyleSheet.create` would be invalid, since `StyleSheet` is naively (by me) marked as a client module. 

We can (and probably should) circle back and optimize the modules even more for `react-server` environments, but basic support is substantially nicer than no support.

### DIFF
--- a/packages/react-native-web/src/exports/View/index.js
+++ b/packages/react-native-web/src/exports/View/index.js
@@ -8,6 +8,8 @@
  * @flow
  */
 
+'use client';
+
 import type { PlatformMethods } from '../../types';
 import type { ViewProps } from './types';
 
@@ -69,13 +71,16 @@ const View: React.AbstractComponent<ViewProps, HTMLElement & PlatformMethods> =
     } = props;
 
     if (process.env.NODE_ENV !== 'production') {
-      React.Children.toArray(props.children).forEach((item) => {
-        if (typeof item === 'string') {
-          console.error(
-            `Unexpected text node: ${item}. A text node cannot be a child of a <View>.`
-          );
-        }
-      });
+      // Skip parsing RSC objects as this will prevent errors from surfacing in the renderer.
+      if (!props.children || !('$$typeof' in props.children)) {
+        React.Children.toArray(props.children).forEach((item) => {
+          if (typeof item === 'string') {
+            console.error(
+              `Unexpected text node: ${item}. A text node cannot be a child of a <View>.`
+            );
+          }
+        });
+      }
     }
 
     const hasTextAncestor = React.useContext(TextAncestorContext);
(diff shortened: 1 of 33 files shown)


packages/react-native-web/src/exports/View/index.js
```diff
@@ -69,13 +71,16 @@ const View: React.AbstractComponent<ViewProps, HTMLElement & PlatformMethods> =
...
     if (process.env.NODE_ENV !== 'production') {
-      React.Children.toArray(props.children).forEach((item) => {
-        if (typeof item === 'string') {
-          console.error(
-            `Unexpected text node: ${item}. A text node cannot be a child of a <View>.`
-          );
-        }
-      });
+      // Skip parsing RSC objects as this will prevent errors from surfacing in the renderer.
+      if (!props.children || !('$$typeof' in props.children)) {
```
necolas, line 75: This fails when the child is a string

EvanBacon: I think the issue may have been coming from a mismatched version of `react-is` that I had. Updating everything seems to fix the issue, I'll remove the statement.

packages/react-native-web/src/modules/useLocale/index.js
```diff
@@ -6,6 +6,7 @@
  *
  * @flow strict
  */
+'use client';
```
necolas, line 9: Can this move to [`exports/useLocaleContext`](https://github.com/necolas/react-native-web/tree/master/packages/react-native-web/src/exports/useLocaleContext)?

packages/react-native-web/src/modules/useResponderEvents/index.js
```diff
@@ -7,6 +7,8 @@
  * @flow
  */
 
+'use client';
```
necolas, line 10: Why does this module need `use client`?

packages/react-native-web/src/vendor/react-native/FlatList/index.js
```diff
@@ -8,6 +8,8 @@
  * @format
  */
 
+'use client';
```
necolas, line 11: Can this move to [`exports/FlatList`](https://github.com/necolas/react-native-web/tree/master/packages/react-native-web/src/exports/FlatList)?

PR #2680, opened 2024-06-06
Add basic support for React Server Components

### LINKED ISSUE #2672: Add "use client" directive for a proper usage with Next App Router

Would you be open to add [React `"use client"`](https://react.dev/reference/rsc/use-client) directive to all files that requires it so react-native-web can be easily used with Next.js without having to add this directive too early on pages ?

Problem when you add it yourself in your pages or in some component that use RNW primitives: this can conflict with some Server Components or server code that will try to be bundled in frontend bundle.

I can do a PR for this, if you are open to.

necolas:
Close #2678
Fix #2672

### DIFF
--- a/packages/react-native-web/src/exports/AccessibilityInfo/index.js
+++ b/packages/react-native-web/src/exports/AccessibilityInfo/index.js
@@ -6,6 +6,9 @@
  *
  * @flow
  */
+
+'use client';
+
 import canUseDOM from '../../modules/canUseDom';
 
 function isScreenReaderEnabled(): Promise<*> {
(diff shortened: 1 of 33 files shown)


johnf:
@necolas FYI I'm using a local patch very similar to this and it works quite well for me.
Happy to help out with any testing you need

necolas:
> I'm using a local patch very similar to this and it works quite well for me.

Did you have to make any changes? I didn't hear back from @EvanBacon as to whether this was sufficient

johnf:
> Did you have to make any changes? I didn't hear back from @EvanBacon as to whether this was sufficient

I also added 'use client' to
* AppRegistry/AppContainer
* useLocale
* useResponderEvents

But I don't remember if they were "needed" or if I was just adding to everything in sight.

I can do some more testing if you think those aren't needed.

EvanBacon:
@necolas this seems good enough for general usage. Will let you know if we find anything broken.

PR #2682, opened 2024-06-10
chore: use @react-native/assets-registry

EvanBacon:
- Metro has a [single string value](https://github.com/facebook/metro/blob/6e716997fa22b65b0c78931fc6eb648fb9b2f5a1/packages/metro-config/src/defaults/index.js#L94) that it uses for registering assets, and this value isn't multi-platform. 
- When an asset is found by Metro, this value is used to register the module ID for runtime lookup in modules such as `Image`.
- To support both native and web platforms in Expo CLI, we alias the resolution for `react-native-web`s asset registry to a shared module ID to ensure `getAssetByID` can find the same module that Metro uses for `registerAsset` [here](https://github.com/expo/expo/blob/7a4a479b2f311781c3444590ea8c15133bbf1fa3/packages/%40expo/cli/src/start/server/metro/withMetroMultiPlatform.ts#L490-L502). 
- This PR simplifies the logic by using the same module as React Native.


necolas:
Looks like the unit tests and builds fail because the new dependency contains uncompiled JS code.

```
FAIL packages/react-native-web/src/exports/Image/__tests__/index-test.js
  ● Test suite failed to run

    Jest encountered an unexpected token

    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.
```

necolas (changes requested):
Upstream package needs fixes

headfire94:
flow removal PR: https://github.com/facebook/react-native/pull/44002

headfire94:
For those who don't want to wait i solve it using:
1. babel alias `'@react-native/assets/registry': '@exodus/react-native-web/dist/cjs/modules/AssetRegistry',` so that mobile request AssetRegitry from web (code is the same)
2. `assetRegistryPath: @exodus/react-native-web/dist/cjs/modules/AssetRegistry/index.js` in metro config` so that metro use same instance

PR #2709, opened 2024-08-16
feat: check img uri is in cache reliably

### LINKED ISSUE #2708: Suggestion: A more reliable way to check if image is in the cache

### Problem

Original issue: https://github.com/Expensify/App/issues/45853

Most of the time if the image is loaded once, it will be in the cache, and subsequent fetch for the same image URI will hit the cache. However, this implementation

https://github.com/necolas/react-native-web/blob/54c14d64dabd175e8055e1dc92e9196c821f9b7d/packages/react-native-web/src/exports/Image/index.js#L203-L212

indicate that if the initial state of the image will always be `IDLE` even if the image is loaded before.

### Solution

We will modify the `ImageLoader.has` logic to be able to check reliably if the image is in the cache. It works both when the image URI is previously loaded by `<img` tag or by the `Image` component in `react-native-web`

```js
let globalImageToCheckCache = new window.Image()
ImageLoader.has = (uri) => {
   globalImageToCheckCache.src = uri;
   let isInCache = globalImageToCheckCache.complete
    globalImageToCheckCache.src  = '' // unassign to avoid firing extra request
    return isInCache
}
```

dominictb:
Fix #2708

### DIFF
--- a/packages/react-native-web/src/modules/ImageLoader/index.js
+++ b/packages/react-native-web/src/modules/ImageLoader/index.js
@@ -8,6 +8,7 @@
  */
 
 const dataUriPattern = /^data:/;
+const imageComponentToCheckCache = new window.Image();
 
 export class ImageUriCache {
   static _maximumEntries: number = 256;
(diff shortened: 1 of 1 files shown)


necolas:
How is this issue supposed to be reproduced within the RNfWeb codebase, so the fix can be tested?

PR #2712, opened 2024-09-11
Update installation.md

gabriel-logan:
Basic starter template, showing how to correctly configure React Native Web using React Native Cli, using webpack and typescript.


gabriel-logan:
Feel free to modify any information in my pr, I just think it would be good to have a doc on how to make react native web work with webpack since react scripts is deprecated and in my opinion no one should use expo to do a serious and real project. 

Furthermore, when I browsed the project doc I felt immense difficulty in knowing how to start my project with react native cli and react native web.

For this reason I created this template and this readme to help other people too, documentation is important guys.

PR #2722, opened 2024-10-11
Remove misleading fake http status code.

Romick2005:
Image load can fail with any http error, not just 404 (Not found), but as in my case it was 401 (Unauthorized).

I guess if we do not know http status we should not make it. But I agree that in 95% it will 404, but not always.

If somebody really need http image status then might use XMLHttpRequest in a separate image load request call.


necolas:
Because it was merged

PR #2725, opened 2024-10-16
[fix] dont throw if invalid locale

ayroblu:
React native web throws a hard error if the locale is invalid.
This is one of the most common errors according to our error logging.
This hard error seems unnecessary, so try catch and just assume ltr if not supported locale


ayroblu:
@necolas let me know if this is reasonable

PR #2726, opened 2024-10-16
Remove API deprecation warnings

necolas:
Removes warnings for non-standard APIs in React Native, given there has been no upstream interest in deprecating them. And the standards-based approach to cross-platform React is instead being built into React Strict DOM.

cc @EvanBacon for Expo approval


EvanBacon (approved):
looks good, agree with the reasoning—let’s get users on to strict dom.

PR #2727, opened 2024-10-19
refactor(deps): removing fbjs dependency by creating a local module for invariant and warning

mateoguzmana:
As part of https://github.com/necolas/react-native-web/issues/2333, `invariant` and `warning` are the last two dependencies in the project used from `fbjs`.

In this PR:
- Local module for `invariant` and replacing the imports
- Local module for `warning` which had another dependency of `emptyFunction` hence it has been added as well.
- `fbjs` dependency has been removed from `packages/react-native-web/package.json`.

I can still see `fbjs` in the main `package-lock.json` as there are other packages within the repo that have `react-native-web` installed as a dependency, hence it might potentially be removed once that dependency is upgraded as well.

Ps: I did the removal of warning in the same PR as it was fairly small, but happy to split this up in another PR if needed.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -16569,7 +16569,6 @@
       "dependencies": {
         "@babel/runtime": "^7.18.6",
         "@react-native/normalize-colors": "^0.74.1",
-        "fbjs": "^3.0.4",
         "inline-style-prefixer": "^6.0.1",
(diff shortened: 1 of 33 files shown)


quantizor:

```
{
  "overrides": {
  }
}
```

Issue #2732, opened 2024-10-31
Add missing fontVariants


benharri:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the issue

additional font variants are available in react native
https://github.com/facebook/react-native/blob/e5dd7d68bf264669fc5c4ce5e69b24249d28558b/packages/react-native/Libraries/StyleSheet/StyleSheetTypes.d.ts#L440C17-L472C24

applying any of the missing ones in react-native-web causes them to be silently ignored

### Expected behavior

fontVariants are applied correctly

### Steps to reproduce

add a fontVariant that isn't available https://github.com/necolas/react-native-web/blob/a5ba27c6226aa182979a9cff8cc23c0f5caa4d88/packages/react-native-web/src/exports/Text/types.js#L38-L42

### Test case

https://codesandbox.io/p/sandbox/cool-resonance-3kqps8?workspaceId=af88e762-6261-40a3-a475-c631585426e7

PR #2733, opened 2024-10-31
Add missing Font Variants

benharri:
tentative fix for #2732


benharri:
this still doesn't seem to be working. where else are the rendered variants filtered?

Issue #2746, opened 2024-12-24
alignContent default value


rosko:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the issue

In `<View>` component's `alignContent` style property, `react-native-web` uses the CSS default value of `stretch` instead of `flex-start` (according to [Yoga](https://www.yogalayout.dev/docs/styling/align-content) layout engine that RN uses)

### Expected behavior

I guess it should be either fixed or at least documented.

### Steps to reproduce

```jsx
import { Text, SafeAreaView, StyleSheet, View } from 'react-native';

export default function App() {
  return (
    <SafeAreaView style={styles.container}>
      <View style={styles.view}>
        <Text style={styles.block}>Text 1</Text>
        <Text style={styles.block}>Text 2</Text>
      </View>
    </SafeAreaView>
  );
}

const styles = StyleSheet.create({
  container: {
    flex: 1,
    justifyContent: 'center',
    backgroundColor: '#ecf0f1',
    padding: 8,
  },
  view: {
    flexWrap: 'wrap',
    // alignContent: 'flex-start',
    width: '100%',
    height: 100,
    gap: 20,
    backgroundColor: 'lightcyan',
  },
  block: {
    width: 100,
    height: 100,
    backgroundColor: 'darkseagreen',
  },
});
```

### Test case

https://snack.expo.dev/@rosko/view-aligncontent-default-value

necolas:
Yoga and RN have different defaults

rosko:
But for `alignContent`, RN has the default value of `flex-start`. Just check the example and link I provided.
It's obvious inconsistency:
Web:
<img width="337" alt="image" src="https://github.com/user-attachments/assets/05b50e86-b11b-4928-9f23-ff0940c3df2c" />
iOS:
<img width="311" alt="image" src="https://github.com/user-attachments/assets/59aa157a-4249-4c8c-b80f-fd08ae85626a" />
Android:
<img width="319" alt="image" src="https://github.com/user-attachments/assets/cb6f3cd9-9db0-4995-8953-33e341063d05" />

rosko:
Also, I copied the example to the code sandbox: https://codesandbox.io/p/devbox/56ngyw

rosko:
Also, I dug the RN code a little bit and here are some proofs as well:
[...]

PR #2747, opened 2024-12-26
[fix] Default value for alignContent style property

rosko:
Sets the default value for `alignContent` style property according to the `react-native` default value.

Related to #2746


rosko (commented):
This PR adds a default value for `alignContent` style property. Also, it changes the hashed CSS class name. Therefore many snapshots were recreated.

PR #2748, opened 2024-12-27
[fix] Support array syntax on transformOrigin

Freddy03h:
> **Array syntax**
> `transformOrigin` also supports an array syntax. It makes it convenient to use it with Animated APIs. It also avoids string parsing, so should be more efficient.

[React Native documentation](https://reactnative.dev/docs/transforms#array-syntax)


Freddy03h:
Is it possible to release a new version? Thank you.

PR #2750, opened 2024-12-30
[ImageLoader] Simplify getSize implementation, call failure callback when decoding fails

### LINKED ISSUE #2749: ImageLoader.getSize can lead to an infinite interval loop

### Describe the issue

`getSize` implementation is more complicated than should be. There are intervals involved which, in some conditions, can keep running "forever". I have stumbled upon this issue, when I used `getSize` in some jest tests. The problem occurs, when an image fails to decode and `naturalWidth`, `naturalHeight` are `0`.

### Expected behavior

No potential for infinite timers. `onerror` callback is called when an image fails to decode.

### Steps to reproduce

Unit tests will be provided in a merge request

### Test case

Unit tests will be provided in a merge request

### Additional comments

_No response_

lkinasiewicz:
Fixes #2749. 

1. Simplified `getSize` implementation, so that there is no possibility for infinite timers.
2. Added handling of decoding failure.
3. Added unit tests for `getSize`.


TeaFather (approved):
looks like it was over complicated before

quantizor:

```
{
  "overrides": {
  }
}
```

Issue #2752, opened 2025-01-04
Style property "boxShadow" doesn't support array of objects


rosko:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Describe the issue

`react-native` allows the setting of `boxShadow` using an array of objects: https://reactnative.dev/blog/2024/10/23/release-0.76-new-architecture#boxshadow.

`react-native-web` supports only strings.

### Expected behavior

`react-native-web` should support setting `boxShadow` as an array of objects. Like it already does for `transform`
(https://github.com/necolas/react-native-web/blob/0.19.13/packages/react-native-web/src/exports/StyleSheet/preprocess.js#L208-L212)

### Steps to reproduce

Compare these two style properties:
```js
    // boxShadow: '#ff0000 5px 5px 5px 10px inset',
    boxShadow: [{ offsetX: 5, offsetY: 5, color: '#ff0000', blurRadius:5, spreadDistance: 10, inset: true }]
```

### Test case

https://snack.expo.dev/@rosko/boxshadow-as-array-vs-string

PR #2753, opened 2025-01-04
[feat] Support style boxShadow as array

rosko:
Related to https://github.com/necolas/react-native-web/issues/2752


rosko (commented):
PR allows to use array syntax/notation for `boxShadow` style property. It also works well with deprecated `shadow*` properties. Added plenty of tests.

PR #2754, opened 2025-01-05
fix: fix weakmap error when create stylesheet

llucn:
`react-native-web` use `WeakMap` in creating `StyleSheet`. Since keys of `WeakMap` must be objects, it raise an error when input style is a one level object. Fix this error by skip the non-object value.

### DIFF
--- a/packages/react-native-web/src/exports/StyleSheet/__tests__/index-test.js
+++ b/packages/react-native-web/src/exports/StyleSheet/__tests__/index-test.js
@@ -133,6 +133,19 @@ describe('StyleSheet', () => {
       expect(() => StyleSheet.flatten([null, false, undefined])).not.toThrow();
     });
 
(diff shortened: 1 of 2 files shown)


necolas:
What is a "one level object"? Please describe how to reproduce the issue.

rm-vrf:
@necolas 
Add unit test for this PR. 
When input single flatten style object, it runs into error because the key of `staticStyleMap` is not `object` type. For example:

```
{
  opacity: 1,
  position: 'absolute'
}
```

llucn:
@necolas 
Add unit test for this PR. 
When input single flatten style object, it runs into error because the key of `staticStyleMap` is not `object` type. For example:

```
{
  opacity: 1,
  position: 'absolute'
}
```

necolas:
That's not a valid use of StyleSheet.create

PR #2758, opened 2025-01-29
[fix] z-index causes images to not show in iOS 18.3 sometimes

ayroblu:
Hey @necolas,

I noticed after I updated my phone to iOS 18.3 that some avatar images stopped loading for me in the account switcher in x.com. A fix appears to be removing this z-index, though I'm not sure about the exact issue. I'm also not sure why a negative z-index was added in the first place here.

Let me know if you need anything


ayroblu:
@necolas ?

ayroblu:
Thanks! @danicunhac Wondering if you have any context on what the issue is?

PR #2764, opened 2025-03-03
inline-style-prefixer version bump

karmatys8:
react-native-web is polluted by https://github.com/robinweser/fast-loops/issues/18

It can be easily fixed by upgrading the inline-style-prefixer to the 7.0.1 version, which deletes redundant fast-loops dependency, that is causing the pollution. Relevant [issue](https://github.com/robinweser/inline-style-prefixer/issues/233)

I deleted backgroundClip import as it has been deleted from the inline-style-prefixer package. Relevant [PR](https://github.com/robinweser/inline-style-prefixer/pull/221)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -41,7 +41,7 @@
         "gen-flow-files": "^0.4.11",
         "glob": "^8.0.3",
         "husky": "^8.0.0",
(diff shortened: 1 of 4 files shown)


quantizor:

{
  "overrides": {
  }
}

Issue #2801, opened 2025-08-22
[Bug] Wrong `pointerEvents` behaviour on 0.21.1


staszekscp:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the issue

While working on the bump we've noticed that `pointerEvents` behaviour has changed causing multiple strange bugs in the app. After an investigation I noticed that in many places the `pointerEvents` got overridden by other inherited values. 

After creating a minimal reproduction I discovered that the `pointerEvents` _worked correctly in the previous version_. However they must have been passed to components via `StyleSheet.create()`. When passed by inline styling or a JavaScript object, `box-none` nor `box-only` properties were **not** applied. 

The previous version worked correctly, because even when passing the `pointer-event` only to the closest children via `> *` CSS was taking care of everything via inheritance. 

https://github.com/user-attachments/assets/e75a15fa-967a-4d34-8edf-daa13d22e1e8

### Expected behavior

Passing the `pointerEvents` property should behave the same no matter if passed by `StyleSheet.create()` or inline styling. 

What's more I believe that the previous StyleSheet.create() implementation should be brought back, since it worked as I would expect:

```
<View style={styles.pointerEventsNone}> // none
  <View style={styles.pointerEventsBoxNone}> // none
  // From here we should have 'auto'
    <View> 
      <View 
        onClick={() => {
          console.log("Should work");
        }}
      />
    </View>
  </View>
</View>
```

The only thing would be to make sure the same logic as in [StyleSheet/compiler/index](https://github.com/necolas/react-native-web/blob/0.20.0/packages/react-native-web/src/exports/StyleSheet/compiler/index.js#L382-L400) was applied for any other way of styling.

### Steps to reproduce

Inline styling bug
1. Add a `View` with inline `style={{ pointerEvents: 'box-none'}}`
2. Open `Elements` in the Inspector and notice that the style was not applied

Wrong propagation
1. Add a `View` with `style={{ pointerEvents: 'none'}}` (created via `StyleSheet.create()`)
2. Nest another `View` with `style={{ pointerEvents: 'box-none'}}` (created via `StyleSheet.create()`)
3. Nest one more `View` with an `onPress` 
4. The `onPress` should be triggered, since its parent has `box-none` value, therefore its children should get `pointerEvents: 'auto'` following heuristic that descendants should inherit from the closest ancestor.

### Test case

https://codesandbox.io/p/sandbox/condescending-wildflower-z6vzpv

### Additional comments

[...]

PR #2802, opened 2025-08-26
Fix: `pointerEvents` propagation

staszekscp:
### Overview
The PR fixes two issues reported [here](https://github.com/necolas/react-native-web/issues/2801):
1. Wrong behaviour of `pointerEvents` being set explicitly for all the descendants instead of leaving the CSS inheritance to take care of the style propagation. 
2. Incorrect `pointerEvents` for values passed via inline styles instead of `StyleSheet.create()`

### Testing
I've tested the change on the Expensify app that uses `react-native-web`. The [bug that was reported by a reviewer](https://github.com/Expensify/App/pull/68902#issuecomment-3211046596) on my PR doesn't appear with my fix included.

I also did a [small reproduction](https://github.com/staszekscp/memory-leak-repro/tree/pointer-events-bug-reproduction) where the changes can be tested. The fix is applied via `patch-package` there.


packages/react-native-web/src/modules/createDOMProps/index.js
```diff
@@ -879,8 +879,11 @@ const createDOMProps = (elementType, props, options) => {
       `props.pointerEvents is deprecated. Use style.pointerEvents`
     );
   }
+  const pointerEventsValue =
+    StyleSheet.flatten(style).pointerEvents || pointerEvents;
```
necolas, line 883: My concern here is that this has a good chance of regressing performance because we're flattening the styles before resolving them. That's probably why the special handling of the pointerEvent prop wasn't ported.

It's been a while since I've looked at the e2e benchmarks but that would be the way to confirm this. Might be a good idea to separate the fix for the pointerEvents behavior from the attempt to support inline styles pointerEvent. The latter is an edge case that can easily be avoided by changes to product code anyway

staszekscp: What I meant is that if the docs say that it's fine to use plain JS objects people will follow this instruction anyways. That's why I think it is should be covered. Maybe there should be at least a warning in the console to provide some feedback to the library users that something might not work as expected.

I've removed the code from the PR so it includes only the fix for the incorrect `pointerEvents` propagation

necolas:
Sorry forgot to do that. I'll merge a couple of other fixes and then get that out today

staszekscp:
Thanks!

PR #2805, opened 2025-09-16
Modal: Significantly improve animation

qwertychouskie:
`ease-in` and `ease-out` were mixed up, making the animation jarring, as it would accelerate faster towards the destination.

The defined `cubic-bezier` animations now used are approximately equivalent to using `Easing.out(Easing.cubic)` for the animation in and `Easing.in(Easing.sin)` for the animation out.

Before:

https://github.com/user-attachments/assets/ab5c6bee-c1f6-4849-988e-8dbe78666f85

After:

https://github.com/user-attachments/assets/181f1a39-b745-4713-ba92-f23287d20165


necolas:
Thanks!

PR #2809, opened 2025-10-11
Fix memory leak in AnimatedProps constructor

### LINKED ISSUE #2756: Memory Leak in Animated API with AnimatedWithChildren

- [x] I have searched the existing issues

### Describe the issue

A memory leak has been identified in the Animated API for react-native-web, specifically in the AnimatedWithChildren class and related animations. The issue arises when rendering multiple animated components using transforms and interpolation. Over time, internal arrays (like children in AnimatedWithChildren) grow indefinitely, leading to increased memory usage and performance degradation.

### Expected behavior

- Memory usage should remain stable over time as animations run and re-renders occur, regardless of the number of Animated components rendered.

- Animated values (Animated.Value, Animated.ValueXY, etc.) should not accumulate indefinitely in the internal array of AnimatedWithChildren.

- The AnimatedWithChildren class or related classes should ensure that duplicate or unused children are not repeatedly added to its array.

- Animations should not degrade in performance as time progresses, even with multiple animated components and transformations (e.g., transform styles with interpolated values).

### Steps to reproduce

1.	Implement a component or file that periodically renders multiple Animated elements using transform and interpolate, as these trigger the issue more frequently.
1.	Monitor performance degradation visually as animations run.
1.	Use tools like Chrome DevTools to confirm memory growth over time, by taking heap snapshots.

<br>

Package Versions
- 	Expo: ~52.0.27
- 	React Native: 0.76.6
- 	React Native Web: ^0.19.13
- 	React: 18.3.1
- 	React DOM: 18.3.1

### Test case

https://codesandbox.io/p/sandbox/animated-memory-leak-7j5d2n

### Additional comments

CodeSandbox isn’t the most reliable environment for demonstrating memory growth due to other objects accumulating during runtime. Instead, I’ve hosted a reproduction of the issue on Netlify here:

👉 [Live Demo](https://animated-api-memory-leak.netlify.app/)

You’ll notice performance degradation within approximately 10 seconds, depending on your machine. Using Chrome DevTools or similar tools, you can observe memory snapshots that reveal the growth of Animated elements in the Array constructor.

**Example Code**

Below is the code used to reproduce the issue. You can implement this yourself to observe the problem:

```javascript
import React, { useState, useEffect, useMemo } from 'react';
import { Animated, View, StyleSheet } from 'react-native';

const COLORS = ['tomato', 'skyblue', 'gold', 'limegreen', 'orchid'];

  const [counter, setCounter] = useState(0);

  const animatedValues = useMemo(
    () => Array.from({ length: 500 }, () => new Animated.Value(0)),
    []
  );

  useEffect(() => {
    const interval = setInterval(() => setCounter((prev) => prev + 1), 15);
    return () => clearInterval(interval);
  }, []);

  useEffect(() => {
    animatedValues.forEach((animatedValue) => {
      Animated.loop(
        Animated.sequence([
          Animated.timing(animatedValue, {
            toValue: 1,
            duration: 1000,
            useNativeDriver: false,
          }),
          Animated.timing(animatedValue, {
            toValue: 0,
            duration: 1000,
            useNativeDriver: false,
          }),
        ])
      ).start();
    });
  }, [animatedValues]);

  const squares = animatedValues.map((animatedValue, index) => (
    <Animated.View
      key={index}
      style={[
        styles.box,
        { backgroundColor: COLORS[index % COLORS.length] },
        {
          transform: [
            {
              translateX: animatedValue.interpolate({
                inputRange: [0, 1],
                outputRange: [0, 100 * (index % 2 === 0 ? 1 : -1)],
              }),
            },
          ],
        },
      ]}
    />
  ));

  return <View style={styles.container}>{squares}</View>;
}

  container: {
    flex: 1,
    flexWrap: 'wrap',
    flexDirection: 'row',
    justifyContent: 'center',
    alignItems: 'center',
  },
  box: {
    width: 50,
    height: 50,
    margin: 10,
  },
});
```

<br>

**Temporary Fix**

A potential short-term fix involves modifying the addChild method in AnimatedWithChildren to avoid redundant additions to the children array:

```javascript
addChild(child) {
  if (!this.children.includes(child)) {
    this.children.push(child);
  }
}
```

This change eliminates memory growth, but it seems like a hack.

c-miles:
Fixes a memory leak in `AnimatedProps` that causes unbounded accumulation of child nodes in applications with frequent re-renders.

**Fixes #2756**. Builds on recent improvements in this area like #2800.

### Background

During investigation of facebook/react-native#49217, I discovered this issue is specific to react-native-web. This area of the vendored Animated implementation has diverged from React Native core, where the `AnimatedProps` constructor does not call `__attach()`.

### Root Cause

The `AnimatedProps` constructor calls `this.__attach()` (line 35), then `useAnimatedPropsLifecycle` calls `node.__attach()` again (line 110 in `useAnimatedProps.js`). This causes the **same `AnimatedProps` instance** to add itself as a child to parent nodes twice.

Each re-render creates a new `AnimatedProps` with this double-attachment, but cleanup only removes one reference. The result is **unbounded accumulation** of duplicate children. In a stress test with 500 animated components re-rendering every 50ms, this produced 359,382 duplicate `__addChild` calls.

**[Live demo of the issue](https://snack.expo.dev/@miles121/smelly-blue-croissant)** (note: Android/iOs works fine, but web version exhibits the leak)

### Approach

This PR removes the constructor `__attach()` call entirely.

**Why this is safe:**
- `useLayoutEffect` in `useAnimatedProps` (line 110) synchronously calls `node.__attach()` before browser paint
- No code path requires immediate attachment during construction
- React Native core's implementation proves this pattern works correctly

Initially considered adding a defensive check in `__addChild()` to prevent duplicates:

```js
if (this._children.includes(child)) return
```

While this prevented the leak, it masked the underlying issue rather than addressing the root cause.

### Impact

- **Breaking changes:** None
- **Performance:** Eliminates memory leak in animated applications
- **Compatibility:** Aligns behavior with React Native core implementation

### DIFF
--- a/packages/react-native-web/src/vendor/react-native/Animated/nodes/AnimatedProps.js
+++ b/packages/react-native-web/src/vendor/react-native/Animated/nodes/AnimatedProps.js
@@ -32,7 +32,6 @@ class AnimatedProps extends AnimatedNode {
     }
     this._props = props;
     this._callback = callback;
-    this.__attach();
   }
 
   __getValue(): Object {


necolas:
Thanks!

Issue #2824, opened 2026-02-09
Bump `styleq` dependency to 0.2.x


joonaojapalo:
### Is there an existing request?

- [x] I have searched for this request

### Describe the feature request

Hi! I’m wondering whether bumping the `styleq` dependency to the latest 0.2.1 would be considered a straightforward change and potentially suitable for an external contribution.

I skimmed through the [changes between styleq 0.1.3 and 0.2.1](https://github.com/necolas/styleq/compare/0.1.3..0.2.1) and it seems that the main requirement would be implementing `localizeStyle` as described in the “Polyfilling features at runtime” section of the [styleq README](
https://github.com/necolas/styleq?tab=readme-ov-file#polyfilling-features-at-runtime).

Background: I’m working in a monorepo that uses react-native-web and stylex in separate sub-projects. The latest versions of these depend on different styleq versions (0.1.3 and 0.2.1 respectively), which currently requires a dependency resolution override. This led me to wonder whether aligning the upstream dependency would be feasible.

I’d appreciate any guidance on whether this upgrade makes sense and if you’d be open to a PR in this area.

PR #2825, opened 2026-02-09
Update styleq, bring localizeStyle transform locally

edkimmel:
Issue #2824

react-strict-dom has bumped styleq up to 0.2.1, causing a conflict with this repository.  The transform-localize-style transformer in styleq was removed.  

This PR bumps the version of styleq to 0.2.1 to match and brings the runtime localizeStyle transform into this repo.


edkimmel:
@necolas it looks like styleq's output changed to [include a debug string](https://github.com/necolas/styleq/commit/3651073043bee810db7c0f1aff8ce5d95f143360) as the third element in the array.  Is that something we want to be stripping out of the array for the StyleSheet API or should I update the snapshots to account for it?

PR #2836, opened 2026-03-23
Allow vertical overflow for Text with `numberOfLines={1}`

mozzius:
Setting `numberOfLines` on Text causes the text to get clipped in both axis, which means if you have a smaller line height it might clip ascenders/descenders. Exaggerated example:

Screenshot 2026-03-23 at 12 28 43

The fix is to use `overflow-inline: clip` instead of `overflow: hidden`, so that it just clips it in the horizontal axis. This matches React Native's behaviour on iOS/Android. Unfortunately this only works for single-line clamps - to my knowledge it's unfixable for multi-line clamps.

Screenshot 2026-03-23 at 12 28 14

[`overflow-inline` is baseline newly available](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Properties/overflow-inline), if that's a problem we could use `overflow-x: clip` instead. However, that wouldn't support vertical writing modes :/. If that's an unreasonable compromise, feel free to put this PR on ice for however many years it takes to become widely available :)

### DIFF
--- a/packages/react-native-web-examples/pages/text/index.js
+++ b/packages/react-native-web-examples/pages/text/index.js
@@ -314,7 +314,7 @@ function LineExample({ description, children }) {
 
       <View
         style={{
-          borderWidth: 2,
+          borderWidth: 1,
           borderColor: 'black',
           width: 200
         }}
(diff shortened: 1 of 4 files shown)


mozzius:
Dammit I did not mean to link these PRs together

Issue #2782, opened 2025-06-21
BUG: "textShadow*" style props are deprecated. Use "textShadow"


psnet:
### Is there an existing issue for this?

- [x] I have searched the existing issues

### Describe the issue

Using shadow `textShadow*` props for `Text` styles causes warning message in console:
```
"textShadow*" style props are deprecated. Use "textShadow".
```

but all `textShadow*` props still relevant and NOT deprecated:

https://reactnative.dev/docs/text-style-props#textshadowcolor
and other `textShadow*` props are working 

Example:
```js
const styles = StyleSheet.create({
    text: {
        textShadowOffset: {
            height: 1,
            width: 0,
        },
        textShadowRadius: 0,
        textShadowColor: '#555',
    },
});
```

Error located here:
https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/StyleSheet/preprocess.js#L169

More errors
---

Also one line

https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/StyleSheet/preprocess.js#L162

we see comma `,` at end of line which seems like an error

```
(options.textShadow === true,
```

the same on line

https://github.com/necolas/react-native-web/blob/master/packages/react-native-web/src/exports/StyleSheet/preprocess.js#L144

```
(options.shadow === true,
```

And these were introduced long time ago

https://github.com/necolas/react-native-web/commit/2a901e5fb11ac314c753867fbc5b87060ed05f53#diff-07ace23d03f77f7d4ac5a51d41ea9ce93ca07a9b45a7f9deeb2f199d366c2217R118

https://github.com/necolas/react-native-web/commit/2a901e5fb11ac314c753867fbc5b87060ed05f53#diff-07ace23d03f77f7d4ac5a51d41ea9ce93ca07a9b45a7f9deeb2f199d366c2217R140

### Expected behavior

Warning message will not be shown

### Steps to reproduce

```js
const styles = StyleSheet.create({
    text: {
        textShadowOffset: {
            height: 1,
            width: 0,
        },
        textShadowRadius: 0,
        textShadowColor: '#555',
    },
});
```

PR #2844, opened 2026-07-16
Fix comma-operator bug that made shadow*/textShadow* deprecation warnings and conversion unconditional

KAMRONBEK:
## Problem

Using `textShadow*` (or `shadow*`) style props triggers the deprecation warning `"textShadow*" style props are deprecated. Use "textShadow".` for every caller of `preprocess`, regardless of the `shadow` / `textShadow` option flags, and the conversion to `boxShadow`/`textShadow` also runs unconditionally.

## Root cause

In `packages/react-native-web/src/exports/StyleSheet/preprocess.js` the two guards were written with the comma operator instead of a logical AND:

```js
if (
  (options.shadow === true,
  style.shadowColor != null || ...)
) {
```

The comma operator evaluates and **discards** `options.shadow === true`, so the condition reduces to "style contains shadow props" and the flag introduced in 2a901e5f ("Once Image is implemented using createElement, we can disable the 'boxShadow' generation just for Image") never had any effect.

## Fix

Use a logical AND so the conversion and its deprecation warning only run when the corresponding option is enabled:

```js
if (
  options.shadow === true &&
  (style.shadowColor != null || ...)
) {
```

Behavior through the public `StyleSheet` API is unchanged (it passes `{ shadow: true, textShadow: true }` by default); only direct `preprocess` callers that did not opt in stop warning/converting, as originally intended.

To be explicit about scope: the deprecation warning for `shadow*`/`textShadow*` through the public API is intentional and retained — those props **are** deprecated. This PR fixes the broken option flag, not the deprecation itself, so the warning in the issue's repro still fires by design. If removing the public-API warning is wanted too, that seems like a separate decision (the `remove-deprecation-warnings` branch deliberately kept these two).

## Tests

- Updated the existing shadow/textShadow conversion tests in `preprocess-test.js` to opt in via the option flags.
- Added a `deprecation warnings for shadow styles` describe block asserting: no `console.warn` and no conversion when the flag is not set (previously false-positive paths), and that the warning still fires when the flag is set. The new no-warning tests fail without the fix.
- `jest` (full dom + node suites), `prettier --check`, `eslint`, and `flow` all pass.

Refs #2782


KAMRONBEK:
@necolas TL;DR: the two shadow guards in `StyleSheet/preprocess.js` use a **comma operator** instead of `&&` — `(options.shadow === true, style.shadowColor != null || …)` — so the `options.shadow`/`options.textShadow` flags are evaluated and discarded, and the deprecation warning + boxShadow/textShadow conversion run for every caller regardless of the flags. This makes the flags work as originally intended (2a901e5f). Note the public `StyleSheet` API passes both flags as `true`, so the deprecation warning there is retained by design — this PR fixes the broken gate, not the deprecation policy. Tests fail without the fix; jest/flow/eslint/prettier all green.

